PR #2, opened 2013-05-21
HSSFWorkbook.getAllEmbeddedObjects wasn't getting all embedded objects

hakanai:
HSSFWorkbook.getAllEmbeddedObjects wasn't getting all embedded objects, only the top-level ones. This fixes it to get them recursively.

### DIFF
--- a/src/java/org/apache/poi/hssf/usermodel/HSSFWorkbook.java
+++ b/src/java/org/apache/poi/hssf/usermodel/HSSFWorkbook.java
@@ -1722,11 +1722,7 @@ public final class HSSFWorkbook extends POIDocument implements org.apache.poi.ss
         if (null == patriarch){
             return;
         }
-        for (HSSFShape shape: patriarch.getChildren()){
-            if (shape instanceof HSSFObjectData){
-                objects.add((HSSFObjectData) shape);
-            }
-        }
+        getAllEmbeddedObjects(patriarch, objects);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Gagravarr:
I may be missing something, but I can't see how your new method ends up getting called?

Also, any chance of a unit test for this (likely with a test file), which shows that some objects were missed before, but are now being fetched?

hakanai:
Now that you ask it, I'm not sure if the diff is complete or not. I'll have to check our offline copy again. The summary of the issue was that the original code only returned the images at the top-level, not those inside groups.

As for samples, the only one I have, I'm not allowed to share.

Gagravarr:
If you do manage to dig out the full patch, that'd be handy. What you could then do is write a very small program that opens each test file we have in turn, and prints out the number of embedded parts. Run that with pre and post fixed jars, if one of them shows a different number then we have an existing test file we can use for the unit test!

hakanai:
https://gist.github.com/trejkaz/5779009 shows results before and after the patch.

I guess there are more sample files than anyone expected - v3.9 only found embedded objects in 4 files and with the fix, 10 files have embedded objects.

Gagravarr:
Looks like files will tend to either have one kind or the other then - I spot lots of files we're now finding things in, but none we found before getting more

I've applied your two patches in r1493001, along with two new unit tests based on your gist, thanks!

PR #4, opened 2013-06-12
introduced XWPFParagraph.getNumFmt function

akhikhl:
New function: XWPFParagraph.getNumFmt, supplied with unit-test.
Another additional function - getNumIlvl - is used by getNumFmt, but could find other uses as well.


Gagravarr:
Thanks, applied with minor changes in r1492312.

PR #7, opened 2014-01-25
Form check box extraction with XWPFWordExtractor

maybeec:
For test issues, we compare automatic filled word forms at least by extracting the documents' text. Therefore it would be great if the XWPFWordExtractor could also extract the values of form check boxes as text to be able to compare their status on text base.

This pull request implements the described feature. Form check boxes in *.docx format will be extracted as |_| for an disabled check box and |X| for an enabled check box.


centic9:
Hi, this is applied with r1636990, should appear in nightly builds and POI 3.11beta3 and later.

maybeec:
Nice, thanks!

maybeec:
Btw: Will the GitHub mirror be updated from time to time? How is the policy?

centic9:
Hi,

Github-update should happen automatically shortly after the checkin to SVN
is done.

Dominik.

maybeec:
Ah ok, I just found it on the trunk branch. Thanks.

PR #8, opened 2014-05-14
Bugfix addpicture

moacyrricardo:
Fix based on patch:
https://issues.apache.org/bugzilla/show_bug.cgi?id=55476


centic9:
We closed the related bug as "Worksforme/Cannotreproduce" a while ago as we could not reproduce the problem. Is this still an issue for you with a recent version of Apache POI? 

Otherwise please close this PR.

moacyrricardo:
We decided to user another strategy to generate our .docx documents.
I will close this PR since your unit tests cover this usage, although I myself haven't tested the current version of Apache POI.

PR #9, opened 2014-07-08
Manipulation of SlideMaster, SlideLayout and Placeholders

valerian-at-pellucid:
Done :
- add / remove a slide layout
- add / remove a slide master
- insert placeholder by type

Also, I had to fix/improve how relationships work.

When inserting a slide master, i create a new Style. This style is similare to the default style of Powerpoint.

Jar versions are available here : 
https://bintray.com/pellucid/maven/poi/view
At the moment the last one is here : https://bintray.com/pellucid/maven/poi/3.11-beta1-pellucid-7/view/files/org/apache/poi


centic9:
Unfortunately this PR is quite large and contains unnecessary/invalid changes like version in build.xml and others. Also in the meantime it is outdated and does not apply any more. 

Any chance you can redo it in a few smaller chunks/PRs without any unrelated changes?

src/ooxml/java/org/apache/poi/POIXMLDocumentPart.java
```diff
@@ -368,6 +368,15 @@ public final POIXMLDocumentPart createRelationship(POIXMLRelation descriptor, PO
      */
     protected final POIXMLDocumentPart createRelationship(POIXMLRelation descriptor, POIXMLFactory factory, int idx, boolean noRelation){
         try {
```
onealj: Update the javadoc and/or add a comment if `idx<0` has a special meaning. It might be worth moving this "find the next highest available reference number" code into its own private function.

asfgit:

centic9:

PR #10, opened 2014-07-18
Improvements to temporary file handling

rwegmann:
A set of commits that allow to override the way temporary files are created (see https://issues.apache.org/bugzilla/show_bug.cgi?id=56735).

There are two small API changes related to what I see as possible race condition vulnerability when creating temporary files:
- org.apache.poi.util.TempFile.createTempFile() now throws an IOException
- I deleted the org.apache.poi.util.PackageHelper.createTempFile() method as it is not used by POI and actively would re-enable the race condition vulnerability.

### DIFF
--- a/src/java/org/apache/poi/util/TempFile.java
+++ b/src/java/org/apache/poi/util/TempFile.java
@@ -18,47 +18,99 @@
 package org.apache.poi.util;
 
 import java.io.File;
-import java.util.Random;
+import java.io.IOException;
 
 /**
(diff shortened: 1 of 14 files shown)


rwegmann:
Thank you for including the changes, Nick.

(I managed to botch up this pull request. I'm new to git and github and it shows. For future reference in case someone cares, the original changes of the request can be seen at https://github.com/rwegmann/poi/tree/bug_56735 . At least I learned something new today.)

PR #16, opened 2015-01-19
Add empty files to compile-lib/lib for git users

virtuald:


centic9:
Comment on https://issues.apache.org/bugzilla/show_bug.cgi?id=57468:
These two directories should always be created by the compilation as they usually contain the downloaded 3rd party libs. In fact some of the POI developers work with Git-svn and do not see this problem, so can you please describe a bit more how you trigger this, e.g. which ant-steps from a clean checkout lead to which errors here for you?

PR #17, opened 2015-01-22
Allow processing of non-OOXML core namespace packages

virtuald:
- For example, visio


virtuald:
Bugzilla link: https://bz.apache.org/bugzilla/show_bug.cgi?id=57484

centic9:
This is merged now, unfortunately I missed to state the relation in the check-in and we cannot close PRs manually currently... :(

virtuald:
Thanks!

PR #18, opened 2015-02-23
Fixed issue where a zero length libre office document with a water mark ...

rojotek:
...image was causing an identifier out of range bug.
#### What's this PR do?

Changes the min value for the identifier manager in XWPFDocument to be 0, as the included test document had an identifier with value 0.

Document Sourced from libre office with a watermark and no text.

The newly added testcase will fail if the code change is removed.


Gagravarr:
Any chance you could update the unit test to also check that we correctly found the image in the header?

rojotek:
how does that look @Gagravarr ?

Gagravarr:
Is it worth also asserting that the text of the header is empty, do you think?

Also, I've just noticed the naming of things like `getHeaderArray(int)` which looks a bit odd to me, so I've started a thread on the dev list about it. Your opinion there would be helpful!

rojotek:
Asserting that the text is empty probably makes sense (so have coded it up :) ).

Gagravarr:
Purely style changes, but were it me:
- Might be good to put all the properties checks together
- Would probably be easier to read if you fetched the XWPFHeader once, then used that everywhere you check, rather than calling `doc.getHeaderArray(0)` each time

rojotek:
was that what you had in mind?

Gagravarr:
Perfect, thanks! Committed in r1661908.

rojotek:
Commited through svn and in trunk so closing this PR.

PR #20, opened 2015-05-30
make line breaks consistent - always use line feeds (LF)

darxriggs:
- all files have been converted via dos2unix
- "git show -w #sha" of this commit shows no changes


onealj:
Since POI is developed on multiple platforms, we don't want to favor any one specific line ending. However, we can set svn:eol-style=native on the source files so that clients can check out the source in their native line endings.

I have also set svn:auto-props="*.java = svn:eol-style=native" to set native line endings on new files for clients that support it.

darxriggs:
This pull request is so old that I'll close it.

onealj:
Anyone object to a massive EOL conversion? Should we put this off to POI
4.0?

doc2unix `find -name "*.java"`; svn propset svn:eol-style `find -name
"*.java"`?

Do we want to do an indentation whitespace conversion while we're touching
all the files and potentially breaking patch files and forks?

PR #23, opened 2015-07-06
creates tmp directory if it doesn't exist

sixinli:
make DefaultTempFileCreationStrategy#createTempFile create directory if it doesn't exist

### DIFF
--- a/src/java/org/apache/poi/util/TempFile.java
+++ b/src/java/org/apache/poi/util/TempFile.java
@@ -99,9 +99,12 @@ public final class TempFile {
                 dir.mkdir();
                 if (System.getProperty("poi.keep.tmp.files") == null)
                     dir.deleteOnExit();
+            } else if (!dir.isDirectory()) {
+                // if the specified directory doesn't exist, create it
+                dir.mkdir();
             }
 
-            // Generate a unique new filename 
+            // Generate a unique new filename
             File newFile = File.createTempFile(prefix, suffix, dir);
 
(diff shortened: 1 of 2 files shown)


sixinli:
I think it's important to cover the case when the directory is removed while the application is running. For example, the tmpdir might be cleaned while the application is running (depends on the clean task of the server) and we will get an IOException when that happens. 

For the case when there's a file named the same thing, we couldn't do anything - something already occupies the namespace so we can only deal with the IOExceptions. But this shouldn't happen if the users aren't doing something very wrong.

The unit test covers the case when you pass in a dir (not null) that doesn't exist, it will create it.

sixinli:
Took a deeper look into the issue: 
it seems like if i have two instances running, and instance1 stops while the instance2 is still alive, it will delete the "poifiles" folder regardless, and instance2 will not be able to write to that folder anymore. 

Adding a check to see if "poifiles" exists and recreates it if not will help when instance2 is trying to write to the folder after its deleted. However, if instance1 stops while instances2 is writing to the folder, everything will be screwed. 

Sorry if I have misunderstood your explanation before. Closing this as I don't this as a proper fix. 

Thanks,
Sixin

centic9:
Thanks for the note, btw there are some bug-entries discussion similar things, e.g. https://bz.apache.org/bugzilla/show_bug.cgi?id=57200, one workaround proposed there is to create a dummy-file in the temp-dir to prevent it from being removed in one of the VMs, thus keeping the files for the other VM in place.

PR #24, opened 2015-10-23
Make client anchors permit row values up to 65535.

pdc:
Previously values over 32767 were treated as negative, and the error message gives the maximum as 65536 rather than 65535. This meant you got exceptions like the following:

```
java.lang.IllegalArgumentException: row1 must be between 0 and 65536, but was: -267
```


davidtnorth:
Looks reasonable to me. Tracking merging this on bugzilla at https://bz.apache.org/bugzilla/show_bug.cgi?id=58549

davidtnorth:
Fixed in SVN r1710552. Closing request.

PR #25, opened 2015-10-27
SXSSFSheet enableLocking and tabColor

danilorangelmg:
I needed to lock and color the tab of a .xlsx spreadsheet generated from a SXSSFSheet and realized that there was no implementation in this type of sheet, the fact of using SXSSFSheet is on the large amount of rows exported in spreadsheet and the large consumption of memory that the XSSFSheet used in Jboss.It is the SXSSFSheet the best option.
## 

Implement SXSSFSheet lock(AutoFilter, cells...)

Thanks.


onealj:
Sorry for the delay. Committed in r1795277. http://svn.apache.org/viewvc?view=revision&revision=1795277
This will be included in POI 3.17 beta 1.

With a little rework of src/ooxml/testcases/org/apache/poi/xssf/TestSheetProtection.java, we could probably get unit test coverage of these new functions. Interested?

PR #26, opened 2015-11-13
formula for non-ascii sheet name

hishidama:


megascus:
+1

hishidama:
I withdrew IsDigit because I treated it incorrectly.

PR #27, opened 2016-01-25
Adding utility method to check any protection is enforced

prasad-babu:
This utility method is useful if it is to check only whether document is
enforced (any type) or not


centic9:
Any chance of a small unit-test together with a sample document (or using one of the available sample-documents) which covers the new function?

prasad-babu:
Added test case with sample document

prasad-babu:
sorry, i accidentally closed the request.

src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFTableCell.java
```diff
@@ -233,8 +233,9 @@ public XWPFVertAlign getVerticalAlignment() {
         XWPFVertAlign vAlign = null;
         CTTcPr tcpr = ctTc.getTcPr();
         if (tcpr != null) {
-            CTVerticalJc va = tcpr.getVAlign();
-            vAlign = stVertAlignTypeMap.get(va.getVal().intValue());
+            CTVerticalJc va = tcpr.getVAlign();
+			if(va != null)
```
centic9, line 237: Is this change related? Looks like a different change/fix

centic9:
FYI, I left out the seemingly unrelated change to XWPFTableCell.java, please create a separate PR with a short description and unit-test which reproduces the problem that you try to fix with it if you think this should be applied as well.

prasad-babu:
Sorry that was different bug fix. Unfortunately it was added to this pull request.

PR #28, opened 2016-01-25
Extends parsing for old-format Excel files

edwgiz-zz:
Apache POI  has 'OldExcelExtractor' class to extract textual data from old Excel formats.

It's helpful, but insufficiently when 
1. required a parsing not from all columns, or data of different columns must be differently processed for example they can have a specific boost in a full-text indexing.
3. numeric data is desirable by reason of performance

When Apache POI is already included in project let's don't add other libraries for parsing old Excel files !


Gagravarr:
Thanks for this. We haven't ignored it! Just waiting for one of our HSSF event model experts to review it, and ensure it follows the current HSSF structure for Excel 97+ formats closely enough to be applied.

If you haven't heard anything more in a couple of weeks, please ping us!

onealj:
(ping)

onealj:
@edwgiz Have you rebased OldExcelExtractor on any of your projects?

onealj:
Here's my attempt at rebaseing these changes to trunk. The OldExcelExtractor unit tests currently fail, though.
[github-28.patch.txt](https://github.com/apache/poi/files/1003644/github-28.patch.txt)

asfgit:

centic9:

PR #30, opened 2016-02-19
Fixing Nullpointer exception

prasad-babu:
By Default, table cell doesn't have alignment details even though it is
considered as TOP alignment in MS-WORD


tballison:
Thank you for submitting this.  It looks like the original intent was to return null if the valign was not specified (i.e. tcpr was null).  Would it be ok with you to add your null checks to prevent an NPE but still return null (as opposed to TOP) if, say, va was null:

```
XWPFVertAlign vAlign = null;
CTTcPr tcpr = ctTc.getTcPr();
if (tcpr != null) {
    CTVerticalJc va = tcpr.getVAlign();
    if(va != null)
        vAlign = stVertAlignTypeMap.get(va.getVal().intValue());
}
return vAlign;
```

prasad-babu:
it is ok. thank you.

PR #34, opened 2016-07-11
Add strategy class for creating FakeZipEntry

pjfanning:
Relate to https://bz.apache.org/bugzilla/show_bug.cgi?id=59841
I would like to be able to override the FakeZipEntry behaviour. This code is based on the POI TempFile creation strategy.
The suggested solution for 59841 is to read the xlsx from file but I need to decrypt it and don't want to store the unencrypted data on disk.
I have access to temp file class that uses keeps small data loads in memory and larger ones are pushed to disk but encrypted. I can create a custom FakeZipEntry that uses this class under the hood.
If this approach is acceptable, I can add extra test coverage.


pjfanning:
The approach in https://github.com/apache/poi/commit/11fe2eda86a6f2e2a8ce950f49ef66dde801004c will work for me, so closing this.

PR #39, opened 2016-10-24
Spelling

icirellik:
Correct the spelling of the word initialization.


onealj:
British spelling versus American spelling. Apache POI is an international project, so I don't think it should give preference to any one spelling.

icirellik:
That's a fair point, I was just attempting to make things consistent.

It does bring up a good question, how would you suggest naming classes in this case:
- Allow mixed internationalization?
- First to arrives wins?

onealj:
Probably FTAW. Fortunately get and set are the same. :)

PR #40, opened 2016-10-24
Exceptional TODO's

icirellik:
It's been nearly a decade since this was marked as `TODO`, it seems safe to say no one's working on it and it would be better tracked as a feature request rather than a `RuntimeException`.


Gagravarr:
If memory serves, half the point of this was to annoy someone with a suitable test file into stepping up and helping with the feature...

Handily a Tika user has provided a v5 test file, so I took a few hours today to add in some of the missing support. Due to a lack of official docks, I'm not certain if I've got the type/format mapping quite right, so do please try it with your v5 test files!

centic9:
I believe this one is now fixed as we now have basic support for V5 pointers, or?

icirellik:
Seems to be working for me.

PR #41, opened 2016-10-28
Adding a new method to the XSSFConditionalFormatting class

lucailmartini:
This pull request addresses [Bug #60314](https://bz.apache.org/bugzilla/show_bug.cgi?id=60314).
The new method can be used to change the formatting range of an existing XSSFConditionalFormatting without the need of removing and then re-adding the rule.

If someone is able to provide an implementation for the HSSF format, the method could be added to the interface ConditionalFormatting.


lucailmartini:
I closed this pull request because I selected the wrong branch, sorry!

PR #42, opened 2016-10-28
Adding a new method to the XSSFConditionalFormatting class

lucailmartini:
This pull request addresses [Bug #60314](https://bz.apache.org/bugzilla/show_bug.cgi?id=60314).
The new method can be used to change the formatting range of an existing XSSFConditionalFormatting without the need of removing and then re-adding the rule.

If someone is able to provide an implementation for the HSSF format, the method could be added to the interface ConditionalFormatting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,47 @@
+# Created by .ignore support plugin (hsz.mobi)
+### JetBrains template
+# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and Webstorm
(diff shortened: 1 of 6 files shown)


lucailmartini:
I implemented `setFormattingRanges` also for HSSF.

Pulled up the `setFormattingRanges` to the `ConditionalFormatting` interface.
Pulled up the unit test to the `BaseTestConditionalFormatting` class.
Added some other unit tests to address corner cases.
New code is fully covered by unit tests.

centic9:
Thanks a lot for the nice patch, this is applied in trunk now, this should be included starting with 3.16-beta1

PR #46, opened 2017-01-27
formula for non-ascii sheet name

hishidama:
Japanese characters (such as '・'(\u30fb) and '【'(\u3010)) are letters in Excel, but Java's Character.isLetter() returns false.
Therefore, if these characters are included in the sheet name, formula that references it throws an exception.

I treated these characters as letters.


src/java/org/apache/poi/ss/formula/FormulaParser.java
```diff
@@ -271,7 +271,17 @@ private RuntimeException expected(String s) {
 
     /** Recognize an Alpha Character */
     private static boolean IsAlpha(int c) {
-        return Character.isLetter(c) || c == '$' || c=='_';
+        return IsLetter(c) || c == '$' || c=='_';
+    }
+
+    /** Recognize a Letter Character */
+    private static boolean IsLetter(int c) {
+        return Character.isLetter(c) || c >= 256;
```
onealj, line 279: Won't this incorrectly catch special unicode characters such as an em space? Should whitespace be considered a letter here?
https://en.wikipedia.org/wiki/Whitespace_character

src/java/org/apache/poi/ss/formula/FormulaParser.java
```diff
@@ -271,7 +271,17 @@ private RuntimeException expected(String s) {
...
+    }
+
+    /** Recognize a Letter Character or a Decimal Digit */
+    private static boolean IsLetterOrDigit(int c) {
+        return Character.isLetterOrDigit(c) || c >= 256;
     }
```
onealj, line 285: See unicode whitespace issue from `isLetter`

Would this work (avoids duplicating the rule `c>=256`)?

    return Character.isDigit(c) || FormulaParser.IsLetter(c);

src/testcases/org/apache/poi/ss/formula/TestWorkbookEvaluator.java
```diff
@@ -606,4 +608,38 @@ private void testIFEqualsFormulaEvaluation_evaluateAllFormulaCells(
...
+            String name = String.valueOf(c);
+            try {
+                WorkbookUtil.validateSheetName(name);
+            } catch (IllegalArgumentException e) {
+                continue;
+            }
+            wb.setSheetName(0, name);
+            cell2.setCellFormula(name + "!A1");
+            CellValue value = eval.evaluate(cell2);
+            assertEquals(123d, value.getNumberValue(), 0);
```
onealj, line 640: Any assertions inside a for-loop should hint at what loop iteration resulted in the failure.

    assertEquals(String.format("With sheet name: '%s' (%i)", name, i), 123d, value.getNumberValue(), 0);

src/testcases/org/apache/poi/ss/formula/TestWorkbookEvaluator.java
```diff
@@ -606,4 +608,38 @@ private void testIFEqualsFormulaEvaluation_evaluateAllFormulaCells(
...
+        Cell cell2 = CellUtil.getCell(CellUtil.getRow(0, s2), 0);
+
+        FormulaEvaluator eval = wb.getCreationHelper().createFormulaEvaluator();
+        {
+            cell2.setCellFormula("Sheet1!A1");
+            CellValue value = eval.evaluate(cell2);
+            assertEquals(123d, value.getNumberValue(), 0);
+        }
+
+        for (int i = 256; i <= 65535; i++) {
```
onealj, line 629: To make this unit test run faster, could you pick just a few exemplary characters that should run test whether POI behaves correctly or not? For example, one character from A-Z, one character from a-z (to test case sensitivity), one character from 0-9, one letter from unicode set (ß), one non-whitespace character from unicode set (halfwidth ¥ and fullwidth ￥), several ascii whitespace (space, \n, \r, \t), a unicode whitespace (en space).

    for (char c : new char[] { ... })

src/testcases/org/apache/poi/ss/formula/TestWorkbookEvaluator.java
```diff
@@ -606,4 +608,38 @@ private void testIFEqualsFormulaEvaluation_evaluateAllFormulaCells(
...
+            char c = (char) i;
+            String name = String.valueOf(c);
+            try {
+                WorkbookUtil.validateSheetName(name);
+            } catch (IllegalArgumentException e) {
+                continue;
```
onealj, line 635: Is this test considered failing if an IllegalArgumentException is ever raised by validateSheetName here?
If so:

    } catch (IllegalArgumentException e) {
        fail("Expected sheet validation to pass, but it failed with a sheet name of " + name);
    }

Or is this more the case of "assuming that name is a valid sheet name, verify that Sheet 0's name can be set to name, and a cell formula can be set and evaluated"?

A unit test should clearly define what should pass and what should fail. If setSheetName, setCellFormula, and evaluate call WorkbookUtil.validateSheetName, then this unit test is a tautology. It will always pass and therefore should be pruned from the unit test suite to eliminated wasted test time.

onealj (changes requested):
1) How does unicode whitespace (em space) get handled, and is it consistent with Microsoft Excel?
2) Unit test run time could be significantly faster if you pick a few examples of what should pass sheet name validation and what should fail sheet name validation.

asfgit:

centic9:

PR #47, opened 2017-01-27
Added fullwidth character to WorkbookUtil.validateSheetName

hishidama:
Some Japanese characters can not be used for Excel sheet names.
see https://support.microsoft.com/ja-jp/help/436482

Invalid characters are as follows:
colon ':', '：'(\uff1a)
yen sign(reverse solidus) '\\', '￥'(\uffe5), '＼'(\uff3c)
question mark '?', '？'(\uff1f)
square bracket '[', ']', '［'(\uff3b), '］'(\uff3d)
solidus '/', '／'(\uff0f)
asterisk '*', '＊'(\uff0a)


onealj:
Related: https://bz.apache.org/bugzilla/show_bug.cgi?id=60260
Partial support for unicode sheet names

onealj (commented):
Are there other unicode characters that are allowed to be used as sheet names in Excel? Can this be generalized?
I verified that I can save an OOXML (*.xlsx) workbook with both a regular width yen sign (\u00a5) and a fullwidth yen sign (\uffe5) in Excel 2013 on Windows. The XML files saved in the xlsx zip archive use UTF-8 encoding and save the literal (unescaped) sheet name in xl/workbook.xml.

Sheet names can also use the Yen character in OLE2 (*.xls) workbooks.

I'd be happy to commit the changes that you have here, but I'd prefer to make POI sheet name validation work for all unicode names that are allowed by Microsoft Excel.

asfgit:

centic9:

PR #49, opened 2017-02-20
Word03encrypt

DusonWang:
Resolving encrypted doc03 file


kiwiwings:
Why do you include MD5 / RC4 code? ... can't you use the jce api?

DusonWang:
@kiwiwings use MD5 / RC4 java code is more flexible and jce api ... and rc4.java is different from  jce's rc4.java

DusonWang:
@onealj Because I use the idea editor，code style is different from you code，I will change it and add the Missing ASL 2.0 header in the code，I would like to join the project and  your organization to make its own contribution to open source

kiwiwings:
@DusonWang Please extract that crypto code - like generating and verifying the hash key - out of the HWPFDocument class. There is a container for that : org.apache.poi.poifs.crypt.CryptoFunctions
I've removed the MD5/RC4 classes which were used for HSSF RC4 crypto ... in what aspect is it different and more flexible than the jce api?

PR #50, opened 2017-03-15
DateUtil.isADateFormat() return false for "yyyy\MM\dd\THH:mm:ss"

pyzhou-talend:


onealj:
Please include a unit test in your pull request so that we can avoid regressions.
You should test both escaped-T and unescaped-T variations and verify POI has the same behavior as Excel.
escaped \T: "yyyy\MM\dd\THH:mm:ss"
unescaped T: "yyyy\MM\ddTHH:mm:ss"

pyzhou-talend:
Sorry for the bothering. I found out It's my mistake here.

PR #51, opened 2017-03-19
Fix typo in exception message

kinow:
s/lenght/length

It said

Could not read complex escher property, lenght was

Now it says

Could not read complex escher property, length was


centic9:
Thanks for the PR, this is fixed now along with two comments that also had the same typo.

PR #52, opened 2017-04-07
add a util function;

zzt93:
add a utility function to ExcelToHtmlConverter.java to support the process of input stream, rather than have to provide file


onealj:
Thanks for the pull request. I committed your changes, removing the duplicated code. I also added a method to operate directly on an HSSFWorkbook in addition to the requested InputStream.

r1795256 https://svn.apache.org/viewvc?view=revision&revision=1795256

PR #53, opened 2017-05-12
Adding Null Pointer check

prafulVaishnav:


prafulVaishnav:
Function **ctPageBreak.getVal()** (https://github.com/apache/poi/blob/trunk/src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFParagraph.java#L824) returns null in some cases. So adding Null Pointer check. 

To reproduce the issue, use the attached docx file ([sample.docx](https://github.com/apache/poi/files/995189/sample.docx)) and run below code:

```
InputStream is = new FileInputStream("<attachedFilePath>");
		XWPFDocument docx = new XWPFDocument(is);
		Iterator<XWPFParagraph> paraIter = docx.getParagraphsIterator();
		while (paraIter.hasNext()) {
			XWPFParagraph para = paraIter.next();
			if (para.isEmpty()) {
				//do Something
			}
		}
```

Function **para.isEmpty()** will throw NPE.

@tballison @poi-benchmark Please review and merge.

onealj:
Possible side-note, but shouldn't we be checking that the value isn't any of the true-like string values that OOXML STOnOff allows: "true", "1", or "on"?

onealj:
Thanks for the pull request and unit test. I didn't include your unit test or sample document in our unit test suite because it's a simple NPE. There weren't any other issues uncovered by the integration testing (org.apache.poi.stress.XWPFFileHandler(document/github53.docx)) with this file.

I also took the opportunity to change isPageBreak to work for all of the boolean-like value: "true", "1", "on", "false", "0", "off".

r1795254 https://svn.apache.org/viewvc?view=revision&revision=1795254

PR #55, opened 2017-05-18
fix: NumberFormatException if XSSFName.setNameName is set with a long name

Mille4Ever:
If you call XSSFName.setNameName with a long value consisting of a letter followed by a big number, you will get a NumberFormatException.

**For example:**
I want to set the name "F04030020010". In Excel using name box, I can set the name without any problems. If I want set the same name using poi, I will get the exception mentioned above.  

The reason for the NumberFormatException: 
The method XSSFName.validateName splits the value "F04030020010" in a column part and in a row part. Columns only have letters, rows only numbers. The outcome looks like:
_Column = F_
_Row = 04030020010_

In the next step, row will be converted into a number using _Integer.parseInt_. But the current row value exceeds the max value of an Integer resulting in a NumberFormatException.

Since the logic is fine, I replaced Integer.parseInt with BigDecimal, so there is no problem with parsing big numbers anymore.

### DIFF
--- a/src/ooxml/testcases/org/apache/poi/xssf/usermodel/TestXSSFName.java
+++ b/src/ooxml/testcases/org/apache/poi/xssf/usermodel/TestXSSFName.java
@@ -26,6 +26,8 @@ import static org.junit.Assert.*;
 import org.apache.poi.ss.usermodel.BaseTestNamedRange;
 import org.apache.poi.ss.util.CellRangeAddress;
 
+import java.io.IOException;
+
 /**
  * @author Yegor Kozlov
  */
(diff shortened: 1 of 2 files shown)


onealj:
Why are we attempting to parse the name as a cell address? Sounds like that is the core of the problem, and changing from Integer to BigInt doesn't address the core problem.

Mille4Ever:
The reason is described here (https://support.office.com/en-us/article/Define-and-use-names-in-formulas-4d0f13ac-53b7-422e-afd2-abd7ff379c64, Section "Learn about syntax rules for names"):

_Cell references disallowed: Names cannot be the same as a cell reference, such as Z$100 or R1C1._

Thus, you have to validate that the given name is not a cell reference.

src/ooxml/testcases/org/apache/poi/xssf/usermodel/TestXSSFName.java
```diff
@@ -130,4 +132,22 @@ public void testSetNameName() throws Exception {
...
+
+    @Test
+    public void testSetNameNameLongName() throws IOException {
+        // Test that renaming named ranges doesn't break our new named range map
+        XSSFWorkbook wb = new XSSFWorkbook();
+        wb.createSheet("First Sheet");
+
+        //Name with very long name. Looks like a cell reference but this is not a reference.
+        XSSFName nameSheet1 = wb.createName();
+        nameSheet1.setNameName("F04030020010");
```
pjfanning, line 144: I think it would make sense to adjust setNameName to catch exceptions from isRowWithinRange.  isRowWithinRange seems like it likely to be used for me than just this use case.

Mille4Ever: You mean to change the code inside XSSFName.setNameName and not altering the test case?

Sorry for my question but your comment is just below the test case, so I'm a little bit confused.

onealj:
How about private XSSFName.validateName swallows the NumberFormatException?

This will means that public CellReference.cellReferenceIsWithinRange will continue to raise a NFE if it gets a silly input, letting the caller decide how to handle a non-integer row string.

pjfanning:
Thanks @onealj - that's basically what I was trying to suggest.
One minor detail is that XSSFName validateName has CellReference.cellReferenceIsWithinRange(col, row, SpreadsheetVersion.EXCEL97) and maybe, this should be EXCEL2007 for XSSF.

onealj:
Danke schön @Mille4Ever für die Pull-Anfrage.

Thanks @pjfanning for pointing out the EXCEL97 error. I fixed that too and grep'd through src/ooxml to make sure there weren't any other oversights in XSSF.

PR #56, opened 2017-06-01
Trailing commas custom formats

lucailmartini:
This is the pull request for bug: https://bz.apache.org/bugzilla/show_bug.cgi?id=61143


src/java/org/apache/poi/ss/usermodel/DataFormatter.java
```diff
@@ -691,6 +677,57 @@ private String cleanFormatForNumber(String formatStr) {
...
+            return s.replaceAll(",+$", "");
+        }
+
+        public InternalDecimalFormatWithScale(String pattern, DecimalFormatSymbols symbols) {
+            df = new DecimalFormat(trimTrailingCommas(pattern), symbols);
+            setExcelStyleRoundingMode(df);
+            Matcher endsWithCommasMatcher = endsWithCommas.matcher(pattern);
+            if (endsWithCommasMatcher.find()) {
+                String commas = (endsWithCommasMatcher.group(1));
+                BigDecimal temp = BigDecimal.ONE;
```
onealj, line 692: Commas are not always used as thousands separators. Some Indian number systems use commas as hundreds separators.

Can you strip the commas and read the number without a for loop?

lucailmartini: Hi @onealj, 
I'm sorry I didn't understand your comment.
I thought that custom formats were only saved in US locale. With this assumption, the code above simply counts the number of commas _at the end_ of the formatting string and then divide by 1000^n.
Could you please be more clear?
Thank you very much for your help.

lucailmartini: Just another quick note. The implementation is based on the following comment in the ECMA-376 5th edition specs:

>**,** Thousands separator. The application shall separate thousands by commas if the format
contains a comma that is enclosed by number signs (#) or by zeros. A comma that follows
a placeholder scales the number by one thousand. [Example: If the format is #.0,, and the
cell value is 12,200,000 then the number 12.2 is displayed. end example]

see section 18.8.31 numFmts (Number Formats)

lucailmartini: hi @onealj, did you have the chance to review my comment?

pjfanning:
I've never come across this format but I tried it out in Excel (en_ie locale) and it does work as described.
There is some work ongoing in this area. It isn't exactly related to this but I'm worried that too many commits in this area right now would be counterproductive.
I'm also concerned about how this works in Indian locale (they group numbers in 2s and 3s - eg 1,00,00,000, 1 crore).
Can we get back to this when we've worked on more of the locale issues?

lucailmartini:
> I'm also concerned about how this works in Indian locale (they group numbers in 2s and 3s - eg 1,00,00,000, 1 crore).
Can we get back to this when we've worked on more of the locale issues?

Yes. However, from what I've learned Excel always translates the format into US locale.

pjfanning:
@lucailmartini could you fix the conflicts and we can look at merging this?

pjfanning:
Thanks @lucailmartini - I've merged this using https://svn.apache.org/viewvc?view=revision&revision=1800975

lucailmartini:
Thank you!

PR #58, opened 2017-06-14
[Bug 61170] add extra logging to SaxHelper

pjfanning:
Maybe the extra logging might help with debugging the class loading issue


onealj:
Applied in https://svn.apache.org/repos/asf/poi/trunk@1798792
Will be included in POI 3.17 beta 1

PR #59, opened 2017-06-22
[bugzilla-60422] handle issue with currency format in Germany Locale

pjfanning:
In the ant build, it appears that the java source is assumed to be ASCII so that's why the Euro symbol is escaped.


pjfanning:
Mistake in the code commit

PR #60, opened 2017-06-22
[bugzilla-60422] handle issue with currency format in Germany Locale

pjfanning:
The ant build seems to assume the source is ASCII encoded so that is why I escaped the Euro symbol.


onealj (approved):
I'd like to be able to use unicode characters in the source files, but the current Ant build fails if a source file uses non-ASCII characters. Hence why everything is escaped (and hard to read).

These changes look good. There are some quirks about Excel that assume the user is US Locale, even when they're not.

pjfanning:
This needs to be reworked

PR #61, opened 2017-06-26
Improvements on XSLF chart user model

Alain-Bearez:
The main goal of this work is to be able to:
  - more easily access definitions for axes;
  - more easily set data series for axes;
  - duplicate a slide that has a chart on it.

What is left to be done:
  - allow for more than one data series on bar charts;
  - implement radar charts and other chart types.


onealj:
Thanks for the PR!

There were a few discussions on dev@poi.apache.org about trying to combine the XSLF Chart API with the relatively small XSSF Chart API.
As such, I expect those who are familiar with XSLF Chart and XSSF Chart may be able to provide some feedback and commit all or part of your PR.

It may be a few more days before one of the devs will review your PR, as we're in the process of releasing 3.17 beta 1. If you don't hear back in a week or two, ping us on the dev list. We're all volunteers, so fingers crossed we'll have time to review.

onealj:
If you have the time, additional unit tests for the new and changed functionality would be helpful.

Alain-Bearez:
I understand I hurried a bit to open this pull request, with only one new test and a rewrite of the existing test for pie chart. On the other hand, I use this for an application I am trying to build, visually checking the result looks like expected. Not really fitted for unit testing...

I will think about it during my holiday, one month with my family, away from my dev machine.

Alain-Bearez:
I prepared an XDDF user model, that can be shared between XSLFChart and XSSFChart.

The new pull request will also have a more complete TestXSLFChart, testing the XDDF classes when used from a PowerPoint presentation.

PR #62, opened 2017-07-01
[Bug-60422] fix issue with number formatting in German locale

pjfanning:
Initial stab at fixing this [issue](https://bz.apache.org/bugzilla/show_bug.cgi?id=60422) - needs plenty of extra testing.

* DataFormatter should set the LocaleUtils user locale to match the DataFormatter locale if set (and if LocaleUtils does not have an explicit locale set already)
* org.apache.poi.ss.format.CellNumberFormatter has a lot of hardcoding about commas for grouping separators and dots for decimal separators - should use DecimalFormatSymbols class
* org.apache.poi.ss.format.CellNumberFormatter also applies the separator from the format string but my understanding is that "#,##0.00" means to apply locale specific separators, that the '.' is assumed to mean apply locale specific decimal separator as opposed to specifically apply '.' as the decimal separator.

### DIFF
--- a/src/java/org/apache/poi/ss/usermodel/DataFormatter.java
+++ b/src/java/org/apache/poi/ss/usermodel/DataFormatter.java
@@ -292,10 +292,6 @@ public class DataFormatter implements Observer {
      * @param cell The cell to retrieve a Format for
      * @return A Format for the format String
      */
-    private Format getFormat(Cell cell) {
-        return getFormat(cell, null);
-    }
-    
     private Format getFormat(Cell cell, ConditionalFormattingEvaluator cfEvaluator) {
         if (cell == null) return null;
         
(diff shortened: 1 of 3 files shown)


src/java/org/apache/poi/ss/usermodel/DataFormatter.java
```diff
@@ -315,62 +311,72 @@ private Format getFormat(Cell cell, ConditionalFormattingEvaluator cfEvaluator)
...
         localeChangedObservable.checkForLocaleChange();
-        
-//      // Might be better to separate out the n p and z formats, falling back to p when n and z are not set.
-//      // That however would require other code to be re factored.
-//      String[] formatBits = formatStrIn.split(";");
-//      int i = cellValue > 0.0 ? 0 : cellValue < 0.0 ? 1 : 2; 
-//      String formatStr = (i < formatBits.length) ? formatBits[i] : formatBits[0];
+        Locale originalUserLocale = LocaleUtil.getUserLocale();
+        if (originalUserLocale != locale) {
+            LocaleUtil.setUserLocale(locale);
```
centic9, line 316: Unfortunately this will create very hard to debug problems in multi-threaded environments, because it changes the current locale for all threads and if one is working on another document with another locale it will break in very strange ways. One option would be locking, however this will be bad for overall performance as it would need to lock at every place that is using LocaleUtils, let me take a look if we can do it differently here somehow...

centic9:
The globally held things in LocaleUtil are a bit dangerous in general, I think we originally intended them so a user can globally change the locale for all code that is using POI. We should avoid using it internally like this as it will break in very subtle ways in multi-threaded environments. Can we avoid using it here by e.g. passing in the locale through the methods instead?

src/java/org/apache/poi/ss/usermodel/DataFormatter.java
pjfanning: setUserLocale sets a ThreadLocal value so only affects the current thread.

pjfanning:
Thanks Dominik for reviewing the code, I'm not a big fan of ThreadLocal usage myself. I dod make a start on trying to passing the Locale as a param but that looks like it will mean pretty big changes in the org.apache.poi.ss.format package.

centic9:
Ah, didn't see that it is a thread-local, then the usage should be fine here as it won't cause any multi-threading issues!

pjfanning:
@centic9 I've created https://github.com/apache/poi/pull/63 which passes the locale through the call stack instead - I prefer this approach but it will require me to add extra test coverage for the new constructors, etc. in the org.apache.poi.ss.format package.

centic9:
Both ways look ok to me.

pjfanning:
thanks @centic9 I think I'll proceed with the PR 63 version

PR #63, opened 2017-07-02
[Bug 60422] fix issue with number formatting in German locale

pjfanning:
Equivalent of https://github.com/apache/poi/pull/62 but without reliance on LocaleUtil.setUserLocale - the locale is passed through the call stack.


pjfanning:
merged using https://svn.apache.org/viewvc?view=revision&revision=1800713

PR #65, opened 2017-07-27
Add xlsx to html converter

sye8:
This class stands alone for functionality tests. Only offers basic functionality. Color formatting is not preserved (due to method changes from HSSF to XSSF). Wrapping content into two additional DIVs not included


pjfanning:
@ysf199711 I think this looks useful but I wonder whether it works better as a standalone project. Or maybe, you could just provide a sample that people could use. The format of the HTML output is something that people would probably want to tweak to suit their own preferences.
If we were to proceed with this in POI, it would need test coverage and the code would need to be consistently formatted (4 spaces for indentation, javadocs, etc.).

sye8:
@pjfanning I was working on a pure  java project to convert excel spreadsheets to PDF and I saw in poi-scractchpad the original xls to html conversion class which preserves some format and I can use Flying Saucer to convert from html to pdf. This was the best free solution I found after some search (aspose is very powerful, but it's proprietary....) The class in scratch pad only supports xls, so I decided to rewrite it to support xlsx. I can provide some test cases and formatting.

asfgit:

onealj:
I agree that this might be more appropriate as a standalone project. This is a lot of code that needs unit tests before it could be merged into POI. It's already a static utility, so it can run standalone as a 3rd party project without issue (no need to modify existing XSSF* classes in order to work).

We could consider adding it to the poi-examples package, which is a package that demonstrates example POI application that converts an office document to HTML. This is starting to get to the fringe of what POI as a library was originally designed for (should POI convert to RTF, ODS, and other formats?).

**Questions for discussion (before further implementation):**
Does this handle rendering comments, pictures, charts, hyperlinks, merged regions, tables, pivot tables, and other advanced features of workbooks? Does it have feature parity with an HSSF to HTML converter? In the interest of reducing duplicated code, what would it take for there to be a single SS to HTML converter?

sye8:
With the tests I have done, I can say that cell formatting will be gone, due to different classes/methods for XSSF. Charts and pictures don't work. I am sorry I haven't tested other features

PR #67, opened 2017-08-03
Introduce XDDF usermodel to share charts implementation between XSLF and XSSF

Alain-Bearez:
The main goal of this work is to be able to:
  - duplicate a slide that has a chart on it (https://bz.apache.org/bugzilla/show_bug.cgi?id=57835);
  - more easily access definitions for axes;
  - more easily set data series for axes;
  - share code between XSLF and XSSF about DrawingML.

What is left to be done:
  - allow for more than one data series on bar, line, scatter and radar charts;
  - complete implementation of first chart types, so that they can be used from XSSFChart;
  - implement other chart types.


Alain-Bearez:
I will work out some more changes in order to have more than one data series per chart, in line with what occurs on XSSF.

PR #70, opened 2017-09-13
fix for bug 61515

Alain-Bearez:
This is basically a one-liner with a test.

Since the creation of the relation leads to reloading the existing relationships for the part equally named that has been deleted, the solution is to clear the relationships before going on.


pjfanning:
Thanks Alain. If you don't mind, I think it is best to wait for a few days to merge this, to avoid causing problems with the ongoing 3.17 release.

PR #73, opened 2017-10-02
Add ReadingOrder support for XSSFCellStyle

irafishbein:
This support already exists for HSSFCellStyle, so just added the same logic for XSSFCellStyle

### DIFF
--- a/src/java/org/apache/poi/ss/usermodel/ReadingOrder.java
+++ b/src/java/org/apache/poi/ss/usermodel/ReadingOrder.java
@@ -1,0 +1,49 @@
+/* ====================================================================
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
+   The ASF licenses this file to You under the Apache License, Version 2.0
+   (the "License"); you may not use this file except in compliance with
+   the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


asfgit:

tomerm:
@ADIR01 / @SEMION1956 FYI.

irafishbein:
@Gagravarr
I am not sure who can take a look at this codechange. 
Can you, please? In case you can't - can you please advise me who can?
Thank you,
Ira

centic9:
I took a quick look, one question in general: 

Is the long-value useful outside of POI at all? Or is it rather something that is only needed internally in the on-disk format? 

Maybe we can just remove the setter/getter which set/get "long" and only provide the enum to the outside to keep the interface simple?

irafishbein:
@centic9 
Thank you very much for looking at it.
I've added mentioned setters/getters to make API similar to an existing one in HSSFCellStyle.
Please let me know if you still believe they should be removed.
Thank you.

centic9:
I have pushed a slightly modified version of the patch which only adds setter/getter for the enum-value. I see no use in providing the long-value to the outside and we probably should switch to the Enum for HSSF as well in the future and deprecate the "short" versions there.

Please close this PR if the change is sufficient for you or comment if you think this should be done differently.

irafishbein:
@centic9 
Thank you, the change looks ok.
Can you please explain me when this change is expected to be part of 'official' build?

tomerm:
@centic9 , answering the "when" question would be great. However I realize that it might be problematic and in any case it would be only tentative. Yet, if you can say in which **version** of POI it will be present for sure it would help a lot. For example, any chances it can make it into 7.10 ?  Many thanks.

centic9:
When is probably hard to say as we currently do not follow a defined release-timeline, release version will be 4.0, which we started to work on 2 months ago, a release typically happens every few months, so early 2018 would be a rough guess.

PR #74, opened 2017-10-12
Remove the useless condition.

Kui-Liu:
The value of the condition "numChars < 0" has been narrowed by a previous IfStatement.
It has no effect on the current control flow.
http://findbugs.sourceforge.net/bugDescriptions.html#UC_USELESS_CONDITION


asfgit:

PR #75, opened 2017-10-12
Replace the inefficient Number constructor with static Number.valueOf() method.

Kui-Liu:
Using new NumberConstructor(num) is guaranteed to always result in a new object whereas Number.valueOf(num) allows caching of values to be done by the compiler, class library, or JVM.
Using of cached values avoids object allocation and the code will be faster.
http://findbugs.sourceforge.net/bugDescriptions.html#DM_NUMBER_CTOR
http://findbugs.sourceforge.net/bugDescriptions.html#DM_FP_NUMBER_CTOR

### DIFF
--- a/src/java/org/apache/poi/hssf/usermodel/DVConstraint.java
+++ b/src/java/org/apache/poi/hssf/usermodel/DVConstraint.java
@@ -228,7 +228,7 @@ public class DVConstraint implements DataValidationConstraint {
 			return null;
 		}
 		try {
-			return new Double(numberStr);
+			return Double.valueOf(numberStr);
 		} catch (NumberFormatException e) {
 			throw new RuntimeException("The supplied text '" + numberStr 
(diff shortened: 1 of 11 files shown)


asfgit:

WoozyG:
This change reverts performance changes made from

https://bz.apache.org/bugzilla/show_bug.cgi?id=57840

see that discussion for empirical results of using Integer.valueOf(int) vs. new Integer(int) in the context of POI rows and cells.

centic9:
Sorry, took me some time to take a look, before simply reverting that again, maybe we can avoid the creation of the objects altogether. In my company we use Trove maps in such performance sensitive places as they use the primitive types directly as keys for maps, this would hopefully provide even better performance than the new Integer(), any taker of a POC for the performance gain?

I created https://bz.apache.org/bugzilla/show_bug.cgi?id=61722 to track this work as this PR is closed already.

PR #76, opened 2017-10-12
Remove the redundant nullcheck of value known to be non-null.

Kui-Liu:
These source code contains redundant checks of known non-null values against the constant null.
http://findbugs.sourceforge.net/bugDescriptions.html#RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE


asfgit:

PR #78, opened 2017-10-14
Add the "final" modifier to public static fields.

Kui-Liu:
This static field public but not final, and could be changed by malicious code or by accident from another package. The field could be made final to avoid this vulnerability.
http://findbugs.sourceforge.net/bugDescriptions.html#MS_SHOULD_BE_FINAL


asfgit:

PR #79, opened 2017-10-14
Remove the method invocation of toString on a string.

Kui-Liu:
Calling String.toString() is just a redundant operation. Just use the String.
http://findbugs.sourceforge.net/bugDescriptions.html#DM_STRING_TOSTRING


asfgit:

PR #80, opened 2017-10-14
Remove the useless object.

Kui-Liu:
The styly is created and modified, but its value never go outside of the method or produce any side-effect. It can be removed.
http://findbugs.sourceforge.net/bugDescriptions.html#UC_USELESS_OBJECT


asfgit:

PR #86, opened 2018-01-04
Expose the language tag associated with a run

glb:
It's useful to have the language tag associated with a run, for example if you want to propagate the language information to an HTML representation of the document for compliance with [WCAG 2.0 3.1.2](https://www.w3.org/WAI/WCAG20/quickref/#qr-meaning-other-lang-id).

### DIFF
--- a/src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
+++ b/src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
@@ -249,6 +249,17 @@ public class XWPFRun implements ISDTContents, IRunElement, CharacterRun {
     }
 
     /**
+     * Get the language tag associated with this run, if any.
+     *
+     * @return the language tag associated with this run, if any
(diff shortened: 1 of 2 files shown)


asfgit:

pjfanning (changes requested):
Thanks @glb - can you add a unit test?

glb:
Thanks @pjfanning -- I've added a unit test based on what I saw for similar methods. I also looked at `TestDocument.docx` which is referenced in the unit test suite and there weren't any language tags in that document, but have tested using documents I have with the language tags set and It Works For Me™.

glb:
Related -- are there any plans for a 3.18 release, and if so could this be cherry-picked back to that release? (I'd be happy to create the PR, but I didn't see an applicable branch.)

pjfanning:
@glb 4.0.0 is the next release but it is not scheduled yet

glb:
@pjfanning thank you!

PR #92, opened 2018-01-24
fixes https://bz.apache.org/bugzilla/show_bug.cgi?id=62038

WolfgangFahl:
With this version I could successfully export 797 slides from 20 powerpoint files that previously had over 117  failing exports.


WolfgangFahl:
see https://bz.apache.org/bugzilla/show_bug.cgi?id=62038

asfgit:

pjfanning (changes requested):
Could we throw IllegalArgumentExceptions if the input parameter is null?

src/ooxml/java/org/apache/poi/xslf/usermodel/XSLFTextRun.java
```diff
@@ -586,10 +597,12 @@ void copy(XSLFTextRun r){
...
         PaintStyle srcFontColor = r.getFontColor();
         if(srcFontColor != null && !srcFontColor.equals(getFontColor())){
-            setFontColor(srcFontColor);
+            // setFontColor will throw IllegalArgumentException if this is not checked
+            if (srcFontColor instanceof SolidPaint)
+                setFontColor(srcFontColor);
         }
 
-        double srcFontSize = r.getFontSize();
+        Double srcFontSize=r.getFontSize();
```
pjfanning: could you retain the original code formatting? ie whitespace around the '=' char

src/ooxml/java/org/apache/poi/xslf/usermodel/XSLFTextRun.java
```diff
@@ -586,10 +597,12 @@ void copy(XSLFTextRun r){
 
         PaintStyle srcFontColor = r.getFontColor();
         if(srcFontColor != null && !srcFontColor.equals(getFontColor())){
-            setFontColor(srcFontColor);
+            // setFontColor will throw IllegalArgumentException if this is not checked
+            if (srcFontColor instanceof SolidPaint)
+                setFontColor(srcFontColor);
```
pjfanning: I'm not sure about this. Is there a good reason for this class not to support all font colours? And if there is a good reason not to accept all font colours, shouldn't we throw an exception if we can't accept the font colour?

Alain-Bearez: The only reason is that my branch for correct support of `CTTextRun` as `XDDFTextRun` is not ready for reviews. 

https://github.com/cuali/poi/branches

src/ooxml/java/org/apache/poi/xslf/usermodel/XSLFTextRun.java
```diff
@@ -222,7 +232,8 @@ public boolean fetch(CTTextCharacterProperties props){
             }
         };
         fetchCharacterProperty(fetcher);
-        return fetcher.getValue() == null ? null : fetcher.getValue()*scale;
+        Double result=fetcher.getValue() == null ? null : fetcher.getValue()*scale;
+        return result;
```
Alain-Bearez: What is the added value of this change?

src/ooxml/java/org/apache/poi/xslf/usermodel/XSLFTextRun.java
```diff
@@ -206,9 +207,18 @@ public void setFontSize(Double fontSize){
     @Override
     public Double getFontSize(){
         double scale = 1;
-        CTTextNormalAutofit afit = getParentParagraph().getParentShape().getTextBodyPr().getNormAutofit();
-        if(afit != null) {
-            scale = (double)afit.getFontScale() / 100000;
+        XSLFTextParagraph pp = getParentParagraph();
+        if (pp!=null) {
```
Alain-Bearez: The project uses spaces around operators.

src/ooxml/java/org/apache/poi/openxml4j/opc/PackagePart.java
```diff
@@ -457,7 +460,16 @@ public boolean hasRelationships() {
 	 * @see org.apache.poi.openxml4j.opc.RelationshipSource#isRelationshipExists(org.apache.poi.openxml4j.opc.PackageRelationship)
 	 */
 	public boolean isRelationshipExists(PackageRelationship rel) {
-		return _relationships.getRelationshipByID(rel.getId()) != null;
```
Alain-Bearez: This is too much of a change to be embedded in this PR about XSLF!

If you really NEED to get these changes in the code base, please open a distinct branch, with the relevant test cases using some of your currently failing files as evidence for the bug fixed. 

The OPC classes are the base for ALL of POI.

Alain-Bearez (changes requested):
I would not accept this pull request changing the behaviour of the base OPC classes.

Alain-Bearez:
For some unclear reason, GitHub does not allow me to open a pull request to your fork.

I implemented all my (few) code review suggestions related to code in XSLF, in my branch: 
https://github.com/cuali/poi/tree/xslf-text-run-font-size

src/ooxml/java/org/apache/poi/xslf/usermodel/XSLFTextRun.java
WolfgangFahl: debugability .. you can see the result in a Debugger.

src/ooxml/java/org/apache/poi/openxml4j/opc/PackagePart.java
WolfgangFahl: for some reason the id might be null - the code should now throw up on this. The patch catches the original problem further up the code hierarchy but it took me to find out because of the null pointer exceptions further down the hierarchy.

src/ooxml/java/org/apache/poi/xslf/usermodel/XSLFTextRun.java
WolfgangFahl: I am sorry about the formatting - but this one of the key changes of the patch. Here you get a null pointer exception if getFontSize returns null

Alain-Bearez:
Feel free to merge my pull request to your branch, in order to solve the whitespace problems.

https://github.com/WolfgangFahl/poi/pull/1

After that, we still need to extract the changes to OPC classes into a distinct pull request, with enough tests to check for any regression in the future. How many of your 20 presentations fail exactly on these relations missing id? Could you share some of them? Or trim them down to some "data-less" presentations but still exposing the behaviour related to relation ids.

WolfgangFahl:
Currently I do not have much time to work on these. 
I suggest to do the double -> Double fix immediately since that problem is obvious.
For the other things I'll need some preparation time which might take a few weeks. I'll put it on my todo list but you'll never know whether or when I get to this.

pjfanning:
I added the double/Double change in https://svn.apache.org/viewvc?view=revision&revision=1822253 - I'll see about adding a unit test overing the coming days.

centic9:

WolfgangFahl:
There is only a "comment" option for me.

centic9:
So it seems that somehow the PRs got broken more severly, can you either report a bug at https://bz.apache.org/bugzilla/ and apply a patch there or create a new PR?

PR #94, opened 2018-01-30
New module RangeCopier.

zmau:
RangeCopier does simple copying of regions on workbook.  It comes in two versions - as XSSFRangeCopier and HSSFRangeCopier. It copies regions in same way as Excel does - tiling destination region with source region as pattern. Source and destination regions may be on different sheets. 
It also takes care of formulas in copied cells.
I have also provided test classes with a few test cases, for both XSSF and HSSF context. Files with test data are also here.

Please take a look at my method RangeCopier.cloneCellContent(). It may be problematic, because I have maybe omitted some Cell attributes which should also be cloned.


asfgit:

pjfanning:
@zmau thanks for starting this Pull Request.
Could you fix the code conflicts?
Also, we need Apache licenses at the top of all source files, could you add them?
Could you also review the code formatting? We like a spaces before '{' characters and to have spaces around expressions.

zmau:
Thanks for letting me jump in.

What do you consider "expressions" in this context ?
Should I - for example - turn 
destRow = destSheet.createRow(rowNo+deltaY)
into 
destRow = destSheet.createRow( rowNo+deltaY ) 
, and
            horizontalFormulaShifter = FormulaShifter.createForColumnCopy(sourceSheet.getWorkbook().getSheetIndex(sourceSheet), 
into
            horizontalFormulaShifter = FormulaShifter.createForColumnCopy( sourceSheet.getWorkbook().getSheetIndex(sourceSheet) , 
?

pjfanning:
rowNo+deltaY should be rowNo + deltaY
other spaces are unnecessary
Have a look at XSSFWorkbook.java, an existing class, to see how we format code.

zmau:
Fixed and pushed.

src/java/org/apache/poi/hssf/usermodel/HSSFRangeCopier.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.apache.poi.hssf.usermodel;
+
+import org.apache.poi.hssf.record.aggregates.FormulaRecordAggregate;
+import org.apache.poi.ss.formula.ptg.Ptg;
+import org.apache.poi.ss.usermodel.Cell;
+import org.apache.poi.ss.usermodel.RangeCopier;
+import org.apache.poi.ss.usermodel.Sheet;
+
+public class HSSFRangeCopier extends RangeCopier {
```
pjfanning, line 28: Could you put `@Beta` annotations on these new classes?

src/java/org/apache/poi/ss/usermodel/RangeCopier.java
```diff
@@ -0,0 +1,162 @@
...
+                     styleMap.put(stHashCode, newCellStyle);   
+                 }   
+                 destCell.setCellStyle(newCellStyle);   
+             }   
+         }   
+         switch(srcCell.getCellType()) {   
+             case STRING:   
+                 destCell.setCellValue(srcCell.getStringCellValue());   
+                 break;   
+           case NUMERIC:
```
pjfanning, line 142: could you fix the indents to align the code?

pjfanning (changes requested):
Could you rename the test files to take out the spaces?

pjfanning (approved):
Merged using https://svn.apache.org/viewvc?view=revision&revision=1824266

mabafu:
Hi there... just found this. Wondering if there is better documentation about RangeCopier and it's correct usage. 

Thanks in advance.

xjlin0:
Here is copy/paste from the test, showing basic usage.  My question: does RangeCopier.copyRange() support copying cell styles?  copyRange() always calls cloneCellContent(sourceCell, newCell, null), which results in not copying styles.
```
    protected void initSheets() {
        sheet1 = workbook.getSheet("sheet1");
        sheet2 = workbook.getSheet("sheet2");
    }

    @Test
    public void copySheetRangeWithoutFormulas() {
        CellRangeAddress rangeToCopy = CellRangeAddress.valueOf("B1:C2");   //2x2
        CellRangeAddress destRange = CellRangeAddress.valueOf("C2:D3");     //2x2
        rangeCopier.copyRange(rangeToCopy, destRange);
        assertEquals("1.1", sheet1.getRow(2).getCell(2).toString());
        assertEquals("2.1", sheet1.getRow(2).getCell(3).toString());
    }
```

PR #95, opened 2018-01-30
Bug 55713: Wrong ErrorStyle for DataValidations

pschucker:
Fix for bug [55713](https://bz.apache.org/bugzilla/show_bug.cgi?id=55713)

Was caused by a missing mapping.

Also added two unit tests.


asfgit:

PR #99, opened 2018-02-08
Bug-62055 - Fix XSSFImportFromXML table resize

kapexx:
Fixes [Bug-62055](https://bz.apache.org/bugzilla/show_bug.cgi?id=62055) and one other issue with `XSSFImportFromXML`.

The main change is that importing `XSSFImportFromXML.importFromXML` now resizes mapped tables when importing XML data. Each table will have the a row count of exactly the amount of imported elements, i.e. it shrinks or grows accordingly. When growing, cells below the table are overwritten.

The other issue this fixes is that table column id's were wrongly used to get the column position. This resulted in data written to wrong columns when the id did not match the positions.

Additional checks have been added to the tests for both issues. Changes to `CustomXMLMappings-complex-type.xlsx` include: 
- Added an additional (unmapped) column to verify that data is written to correct columns when the column id doesn't match the columns position
- Added some text in a cell below a table that will grow, to check if the cell is correctly overwritten when the table grows
- Added some text in a cell inside a table that will shrink, to check if the cell is correctly cleared when the the table shrinks


asfgit:

src/ooxml/java/org/apache/poi/xssf/extractor/XSSFImportFromXML.java
```diff
@@ -87,25 +91,25 @@ public XSSFImportFromXML(XSSFMap map) {
      * @throws IOException  if there are problems reading the input string
      */
     public void importFromXML(String xmlInputString) throws SAXException, XPathExpressionException, IOException {
-
+
```
pjfanning: Could you remove the code formatting changes?

src/ooxml/testcases/org/apache/poi/xssf/extractor/TestXSSFImportFromXML.java
```diff
@@ -98,9 +100,10 @@ public void testMultiTable() throws IOException, XPathExpressionException, SAXEx
 					"<ns1:Schema ID=\"" + cellC7 + "\" SchemaRef=\"b\" />" +
 					"<ns1:Schema ID=\"" + cellC8 + "\" SchemaRef=\"c\" />" +
 					"<ns1:Schema ID=\"" + cellC9 + "\" SchemaRef=\"d\" />");
-
-			for (int i = 10; i < 10010; i++) {
-				testXML.append("<ns1:Schema ID=\"c").append(i).append("\" SchemaRef=\"d\" />");
+			
+			int cellOffset = 10; // cell C10
+			for (int i = 0; i < 10000; i++) {
+				testXML.append("<ns1:Schema ID=\"c").append(i + cellOffset).append("\" SchemaRef=\"d\" />");
```
pjfanning: could you use spaces for indents?

src/ooxml/java/org/apache/poi/xssf/extractor/XSSFImportFromXML.java
```diff
@@ -160,6 +172,57 @@ public void importFromXML(String xmlInputString) throws SAXException, XPathExpre
...
+                XSSFCell cell = row.getCell(cellRef.getCol());
+                if(cell != null) {
+                    cell.setCellType(CellType.BLANK);
+                    cell.setCellStyle(null);
+                }
+            }
+        }
+        
+        // update table area
+        table.getCTTable().setRef(newTableAreaReference.formatAsString());
```
pjfanning: We regard accessing the CT* generated classes as something we want to discourage.
Could you change the the XSSFTable class to expose a setRef method?

src/ooxml/java/org/apache/poi/xssf/extractor/XSSFImportFromXML.java
```diff
@@ -160,6 +172,57 @@ public void importFromXML(String xmlInputString) throws SAXException, XPathExpre
...
+     * If the new row count is greater than the current row count, cells below the table will be overwritten by the table.
+     *
+     * @param table the table to modify
+     * @param newRowCount new row count for the table
+     */
+    private void resizeTable(XSSFTable table, int newRowCount) {
+        
+        table.updateReferences();
+        int rowCount = table.getRowCount();
+        if(rowCount == newRowCount) {
```
Alain-Bearez: The project uses space between keywords (if, for, try, while) and opening parenthesis.

src/ooxml/testcases/org/apache/poi/xssf/extractor/TestXSSFImportFromXML.java
```diff
@@ -85,7 +87,7 @@ public void testImportFromXML() throws IOException, XPathExpressionException, SA
 		}
 	}
 	
-	@Test(timeout=60000)
+	@Test(timeout=60000_00)
```
Alain-Bearez: Why multiply by 100?

Alain-Bearez (changes requested):

src/ooxml/java/org/apache/poi/xssf/extractor/XSSFImportFromXML.java
kapexx: I just noticed `setRef` is already exposed as `XSSFTable.setCellReferences`, so I'll use that.

I also had to use CTTable and CTTableColumn near line 143 though, to get the correct column index. Now I saw that I could have used `XSSFTable.findColumnIndex(String columnHeader)` instead of calculating the column index myself, but it seems I would still have to use `CTTableColumn` to get the column header name. 

I didn't want to make any API changes before, but I guess in this case it would be useful if `XSSFXmlColumnPr` exposes the column header name and maybe also provides its column index. That would avoid using any CT* classes. What do you think?

pjfanning: If you could add any APIs that you need, that would be great. We want to make the generated CT* classes only internal implementation details (allowing us to move away from xmlbeans code generation eventually).

Alain-Bearez: There is an ongoing effort to replace calls to the underlying `CT*` classes, so that these calls be limited to the class directly wrapping the XML object. 

Every attempt in that direction is welcome.

src/ooxml/testcases/org/apache/poi/xssf/extractor/TestXSSFImportFromXML.java
kapexx: Sorry, that was a hack to avoid that the test is terminated to early while debugging. I forgot to change it back. I know that JUnit Timeout rules can be deactivated while in debug mode, but I didn't find any way to do the same for timeouts defined in the annotation.

src/ooxml/java/org/apache/poi/xssf/usermodel/helpers/XSSFXmlColumnPr.java
```diff
@@ -32,56 +31,60 @@ Licensed to the Apache Software Foundation (ASF) under one or more
...
+        return localXPath.toString();
+    }
+
+    public Enum getXmlDataType() {
+
+        return ctXmlColumnPr.getXmlDataType();
+    }
+
+    /**
+     * @return the column name
```
pjfanning: could you add `@since 4.0.0` to the javadoc for new API methods?

kapexx: I reconsidered adding the method in `XSSFXmlColumnPr` because I think it actually belongs to `XSSFTableColumn` (which didn't exist yet, so I've created it - see next commit)

kapexx:
I have created the class `XSSFTableColumn` as wrapper for `CTTableColumn`, which was missing. It exposes the column id, name and `XSSFXmlColumnPr` (if the column has an XML mapping). In this basic version `XSSFTableColumn` only makes it a bit easier to handle non-XML-mapped columns but the class could later be improved to support other column features like column formulas, header/data cell styles, etc.

pjfanning (approved):
LGTM

pjfanning:
@Alain-Bearez are you happy for me to merge the changes?

src/ooxml/java/org/apache/poi/xssf/usermodel/XSSFTableColumn.java
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Create a new table column.
+     *
+     * @param table
+     *            the table which contains the column
+     * @param ctTableColumn
+     *            the table column xmlbean to wrap
+     * @since 4.0.0
+     */
+    public XSSFTableColumn(XSSFTable table, CTTableColumn ctTableColumn) {
```
Alain-Bearez: I would insist in having all methods that use a `CT*` parameter to be annotated as `@Internal` when they are not private.

Alain-Bearez (commented):
My only finding after re-reviewing the changes is related to the (lack of) use of `@Internal` annotation for methods the use internal `CT*` parameters.

Maybe this should be more coherently handled as a distinct global task, all over the `ooxml` code base.

src/ooxml/java/org/apache/poi/xssf/usermodel/XSSFTableColumn.java
kapexx: I'll check the code and add those annotations.

This particular constructor doesn't seem to be useful for the user anyway. I think I could make this constructor's visibility be `protected` instead of `public` and then cleanup `XSSFTable.addColumn` and make it part of the public API. `XSSFTable.addColumn` then is the correct and only way to create new table columns in the API.

Alain-Bearez: Definitively the best option!

kapexx:
I moved the table resizing from the XML import into the XSSFTable. I've gone a step further and implemented some more methods related to the table area, rows and columns. I deprecated and replaced a few methods or made them internal, because those methods easily lead to corrupted tables e.g. if the user forgets set the table's area or to add columns manually after setting the tables area. The replacement methods do some additional verifications and for example automatically add or remove columns to avoid corrupted tables.

PR #101, opened 2018-02-13
Adding a wrapper to evaluate all xssf formula cells "in cell"

sawyna:
There is `XSSFFormulaEvaluator.evaluateAllFormulaCells` already present. But that doesn't remove the formulas from the cells. This wrapper loops over all the cells in all sheets in the workbook, calls `evaluateInCell` which removes the formulas from the cells containing formulas and puts the result.


asfgit:

sawyna:
Hi, can any one please verify this and merge?

pjfanning:
@sawyna could you describe the issue you are trying to fix?
If this seems worth merging, it will need unit test coverage.

sawyna:
There is no wrapper util method to evaluate all formulas in the workbook, replace the formulas with the evaluated values. This pull request tries to address the same

centic9:

sawyna:
@centic9 How do I re-open the PR? I don't seem to get an option anywhere.

centic9:
It seems the GitHub repo is somehow broken as also the diff of the PR is not available any more, so the safest option for now would be to create bug-entry at https://bz.apache.org/bugzilla/ and attach a patch there.

PR #103, opened 2018-03-15
hsmf: support writing properties, fix reading embedded message properties, support getting actual type of property value

dhoelzl:
Hello!

I have found a problem reading properties chunks of embedded/attached MSGs inside a MSG file.
There are 8 bytes less reserved bytes in the header sequence.

Further I have added support for writing message properties chunks and getting the actual property type (not only the usual type) and the raw bytes value from a property value.

This ca be e.g. used when extracting embedded/attached MSGs from a MSG file.

* MessagePropertiesChunk: Fix reading properties from embedded/attached MSG
* Add support for writing MessagePropertiesChunk
* PropertyValue: Add support for getting the actual property type and getting the raw bytes

Regards,
Dominik

### DIFF
--- a/src/scratchpad/src/org/apache/poi/hsmf/datatypes/ChunkBasedPropertyValue.java
+++ b/src/scratchpad/src/org/apache/poi/hsmf/datatypes/ChunkBasedPropertyValue.java
@@ -22,6 +22,9 @@ package org.apache.poi.hsmf.datatypes;
  * TODO Provide a way to link these up with the chunks
  */
 public class ChunkBasedPropertyValue extends PropertyValue {
+    public ChunkBasedPropertyValue(MAPIProperty property, long flags, byte[] offsetData) {
+        super(property, flags, offsetData);
+    }
     public ChunkBasedPropertyValue(MAPIProperty property, long flags, byte[] offsetData, Types.MAPIType actualType) {
         super(property, flags, offsetData, actualType);
     }
--- a/src/scratchpad/src/org/apache/poi/hsmf/datatypes/ChunkBasedPropertyValue.java
+++ b/src/scratchpad/src/org/apache/poi/hsmf/datatypes/ChunkBasedPropertyValue.java
@@ -22,8 +22,8 @@ package org.apache.poi.hsmf.datatypes;
  * TODO Provide a way to link these up with the chunks
  */
 public class ChunkBasedPropertyValue extends PropertyValue {
-    public ChunkBasedPropertyValue(MAPIProperty property, long flags, byte[] offsetData) {
-        super(property, flags, offsetData);
+    public ChunkBasedPropertyValue(MAPIProperty property, long flags, byte[] offsetData, Types.MAPIType actualType) {
+        super(property, flags, offsetData, actualType);
     }
 
     @Override
--- a/src/scratchpad/src/org/apache/poi/hsmf/datatypes/MessagePropertiesChunk.java
+++ b/src/scratchpad/src/org/apache/poi/hsmf/datatypes/MessagePropertiesChunk.java
@@ -20,6 +20,7 @@ package org.apache.poi.hsmf.datatypes;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.List;
 
 import org.apache.poi.util.LittleEndian;
 
@@ -28,6 +29,7 @@ import org.apache.poi.util.LittleEndian;
  * byte header
  */
 public class MessagePropertiesChunk extends PropertiesChunk {
+    private boolean isEmbedded;
     private long nextRecipientId;
     private long nextAttachmentId;
     private long recipientCount;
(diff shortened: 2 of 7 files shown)


asfgit:

src/scratchpad/src/org/apache/poi/hsmf/datatypes/MessagePropertiesChunk.java
```diff
@@ -64,28 +87,34 @@ public void readValue(InputStream stream) throws IOException {
...
         out.write(new byte[8]);
+      }
+      // Now properties.
+      return super.writeHeaderData(out);
+    }
 
-        // Now properties
-        writeProperties(out);
+    @Override
+    public void writeValue(OutputStream out) throws IOException {
```
pjfanning: I'm sure if I understand why this public method is being emptied of its implementation

src/scratchpad/src/org/apache/poi/hsmf/datatypes/ChunkBasedPropertyValue.java
```diff
@@ -22,8 +22,8 @@ Licensed to the Apache Software Foundation (ASF) under one or more
  * TODO Provide a way to link these up with the chunks
  */
 public class ChunkBasedPropertyValue extends PropertyValue {
-    public ChunkBasedPropertyValue(MAPIProperty property, long flags, byte[] offsetData) {
-        super(property, flags, offsetData);
+    public ChunkBasedPropertyValue(MAPIProperty property, long flags, byte[] offsetData, Types.MAPIType actualType) {
```
pjfanning: could you add back the old constructor? we don't remove public methods without deprecating them first

src/scratchpad/src/org/apache/poi/hsmf/datatypes/PropertiesChunk.java
```diff
@@ -105,6 +118,13 @@ public String getEntryName() {
         return props;
     }
 
+    /**
+     * Defines a property. Multi-valued properties are not yet supported.
+     */
+    public void setProperty(PropertyValue value) {
+      properties.put(value.getProperty(), value);
```
pjfanning: the indents are not following the standard - 4 chars per indent

src/scratchpad/src/org/apache/poi/hsmf/datatypes/PropertyValue.java
```diff
@@ -32,13 +33,21 @@ Licensed to the Apache Software Foundation (ASF) under one or more
  */
 public class PropertyValue {
     private MAPIProperty property;
+    private MAPIType actualType;
     private long flags;
     protected byte[] data;
 
     public PropertyValue(MAPIProperty property, long flags, byte[] data) {
+      this.property = property;
```
pjfanning: could you call this(property, flags, data, null) ?

pjfanning (changes requested):
Thanks for the contribution but would it be feasible to add more unit tests?

dhoelzl:
Hello!

Thank you for your quick response!

I have completed the `writeValue` implementation.
But this implementation does not support writing variable length properties (`ChunkBasedPropertyValue`) as this would only work if writing the properties to a `DirectoryEntry` instead of just an `OutputStream`.

The same is true for `readValue`. Optionally, the `readProperties` method could also return all variable length properties in order to implement a variant of `readProperties` from a `DirectoryEntry` (instead of just an `InputStream`).

So far I have not looked at the unit testing system of the POI project. I'll check this.

Regards,
Dominik

dhoelzl:
I've created a unit test which would fail without my changes.
* Reading fixed length properties of an embedded attached message (e.g. getMessageDate)
* Extracting an embedded attached message to a top level message which then can be re-parsed as top level message and opened by Microsoft Outlook if it would be written to disk

pjfanning (commented):
New test fails on my laptop

  <testcase classname="org.apache.poi.hsmf.TestExtractEmbeddedMSG" name="testEmbeddedMSGProperties" time="0.003">
    <failure message="expected:&lt;1276822339000&gt; but was:&lt;1276818739000&gt;" type="junit.framework.AssertionFailedError">junit.framework.AssertionFailedError: expected:&lt;1276822339000&gt; but was:&lt;1276818739000&gt;
	at org.apache.poi.hsmf.TestExtractEmbeddedMSG.testFixedAndVariableLengthPropertiesOfAttachedMSG(TestExtractEmbeddedMSG.java:91)
	at org.apache.poi.hsmf.TestExtractEmbeddedMSG.testEmbeddedMSGProperties(TestExtractEmbeddedMSG.java:68)

dhoelzl:
I fixed it. It was a time zone problem.

pjfanning:
merged using https://svn.apache.org/viewvc?view=revision&revision=1827173

dhoelzl:
I've seen in the type specific PropertyValue derivations that PropertyValue.data is already expected to hold the data in little endian, so there is no need to reverse the data when writing the property.

dhoelzl:
Hello, did you look at my recent changes?

I noticed afterwards that the bytes in the property values have to be stored as little endian and therefore no reversing is allowed when writing the data.

Please merge these changes that no inconsistencies evolve.

Regards,
Dominik

pjfanning:
I added the later changes using https://github.com/apache/poi/commit/3f1972983422cda25eb1420ccc2af8edf7607bdc
Could you verify that what is on trunk has all your changes?

dhoelzl:
Looks good.
Thank you!

PR #112, opened 2018-06-09
 Support for LineStyle in XSSFComment

mayuranchalia:
XSSFComment doesn't have support for Linestyle similar to HSSFComment

In HSSFComment we have functions like setLineStyle and setLineStyleColor, similar feature is required for XSSFComment .

Bugzilla bug 
https://bz.apache.org/bugzilla/show_bug.cgi?id=62247


asfgit:

PR #116, opened 2018-07-23
Feature/xwpf implement endnote

drmacro:
This adds XWPFEndnote, XWPFEndnotes, and abstract base class AbstractXWPFFootnoteEndnote, as well as corresponding methods on XWPFDocument and XWPFParagraph. Adds new FootnoteIdManager class to manage the IDs across footnotes and end notes.

Added unit tests and also tested with real-world application that generates working end notes and footnotes using this API.

### DIFF
--- a/src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFDocument.java
+++ b/src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFDocument.java
@@ -54,11 +54,9 @@ import org.apache.poi.openxml4j.opc.PackagingURIHelper;
 import org.apache.poi.openxml4j.opc.TargetMode;
 import org.apache.poi.poifs.crypt.HashAlgorithm;
 import org.apache.poi.util.IOUtils;
-import org.apache.poi.ooxml.util.IdentifierManager;
 import org.apache.poi.util.Internal;
 import org.apache.poi.util.POILogFactory;
 import org.apache.poi.util.POILogger;
-import org.apache.poi.ooxml.util.PackageHelper;
 import org.apache.poi.wp.usermodel.HeaderFooterType;
 import org.apache.poi.xwpf.model.XWPFHeaderFooterPolicy;
 import org.apache.xmlbeans.XmlCursor;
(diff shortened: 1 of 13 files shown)


asfgit:

src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFDocument.java
```diff
@@ -898,19 +899,32 @@ public XWPFFootnotes createFootnotes() {
...
 
         return footnotes;
     }
 
+    /**
+     * Add a CTFtnEdn footnote to the document.
+     *
+     * @param note CTFtnEnd to be added.
+     * @return New {@link XWPFFootnote}
+     */
```
pjfanning, line 914: Could we add methods that take generated CT classes as `@Internal`?

src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFDocument.java
```diff
@@ -1332,7 +1346,7 @@ public void setZoomPercent(long zoomPercent) {
     public void insertTable(int pos, XWPFTable table) {
         bodyElements.add(pos, table);
         int i = 0;
-        for (CTTbl tbl : ctDocument.getBody().getTblArray()) {
```
pjfanning, line 1335: despite being deprecated, the array methods are better than the list methods - we have a job that cherry picks the necessary CT generated classes and it works better with methods that return concrete results like CTTable[] but misses List<CTTable> which is backed by an inner class (the one our job misses)

src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFEndnotes.java
```diff
@@ -0,0 +1,193 @@
...
+    public XWPFEndnotes() {
+        super();
+    }
+
+    /**
+     * Construct XWPFEndnotes from a package part
+     *
+     * @param part the package part holding the data of the footnotes,
+     * 
+     * @since POI 3.14-Beta1
```
pjfanning, line 43: this doesn't seem correct - this is a new method

pjfanning:
I've merged these changes with some of the adjustments I suggested. I would suggest that a few more checks need to be made about `@Internal` methods etc.

PR #118, opened 2018-07-30
Added get/set for mirror margins setting

drmacro:
Added get and set for even-and-odd headings setting.

I couldn't find any other simple settings that seemed useful for exposing on Document.

Added unit test and tested with real-world project.


asfgit:

PR #119, opened 2018-08-01
Implemented a number of new set and get methods for useful formatting…

drmacro:
This adds getters and setters for additional Run formatting properties. My goal was to implement the properties likely to be used in generating DOCX from non-trivial HTML documents. The only property I couldn't figure out how to set was underline color, which I can come back to.

I also fixed a bug in isHighlight() that returned a false positive if highlight was not set at all (the test case was flawed and didn't catch this case).

I used the built-in OOXML enums so that I could set values using the same strings as specified in the OO XML spec rather than having to translate them to e.g., uppercase keywords as used with the custom enumerations (this is why I implemented setVerticalAlignment() as a replacement for setSubscript() [which was a bad method name anyway.]

I did not implement versions of the methods that involve enumerations to also allow direct specification of the enumeration values themselves--I can if there's a need for it, but for most uses it seems more likely that string values will be the input, e.g., from parsing of XML or HTML or whatever.

### DIFF
--- a/src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
+++ b/src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
@@ -32,6 +32,7 @@ import org.apache.poi.ooxml.POIXMLException;
 import org.apache.poi.openxml4j.exceptions.InvalidFormatException;
 import org.apache.poi.ooxml.util.DocumentHelper;
 import org.apache.poi.util.Internal;
+import org.apache.poi.util.Removal;
 import org.apache.poi.wp.usermodel.CharacterRun;
 import org.apache.xmlbeans.XmlCursor;
 import org.apache.xmlbeans.XmlException;
@@ -394,11 +395,10 @@ public class XWPFRun implements ISDTContents, IRunElement, CharacterRun {
     }
 
     /**
-     * Specifies that the contents of this run should be displayed along with an
-     * underline appearing directly below the character heigh
+     * Get the underline setting for the run.
      *
-     * @return the Underline pattern applyed to this run
-     * @see UnderlinePatterns
+     * @return the Underline pattern applied to this run
+     * @see (@link UnderlinePatterns}
      */
     public UnderlinePatterns getUnderline() {
         CTRPr pr = run.getRPr();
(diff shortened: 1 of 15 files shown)


asfgit:

src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
```diff
@@ -578,7 +577,8 @@ public void setEmbossed(boolean value) {
      * altering the font size of the run properties.
      *
      * @return VerticalAlign
-     * @see VerticalAlign all possible value that could be applyed to this run
+     * @see {@link VerticalAlign} all possible value that could be applyed to this run
+     * @deprecated use {@link XWPFRun.getVerticalAlignment}
```
pjfanning, line 672: could you add `@Removal(version = "4.2")`?

src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
```diff
@@ -1278,4 +1279,162 @@ private void _getText(XmlObject o, StringBuilder text) {
         eastAsia /* east asia */,
         hAnsi /* high ansi */
     }
+
+    /**
+     * Set the text expand/collapse scale value.
+     *
+     * @param percentage The percentage to expand or compress the text
+     */
```
pjfanning: could you add `@since 4.0.0`?

src/ooxml/testcases/org/apache/poi/xwpf/usermodel/TestXWPFRun.java
```diff
@@ -225,7 +214,7 @@ public void testSetGetColor() {
         XWPFRun run = new XWPFRun(ctRun, irb);
         run.setColor("0F0F0F");
         String clr = run.getColor();
-        assertEquals("0F0F0F", clr);
+        assertEquals("0F0F0F", clr);
```
pjfanning: nit: could you remove the extra whitespace at end of line?

drmacro:
I will attempt to implement Mark's idea about run property access.

pjfanning:
@drmacro could you rebase due to PR 120 being merged?

PR #120, opened 2018-08-02
Per Mark Murphy's request, make "XWPF" start of names of abstract cla…

drmacro:


asfgit:

PR #121, opened 2018-08-15
Corrected bugs in setting vertical alignment and emphasis mark. 

drmacro:
Refined test cases to catch the bug.

Tested with real-world code (where the failure was originally detected).


asfgit:

PR #122, opened 2018-08-16
Set table and table cell widths to percentage, twips, or "auto".

drmacro:
Tested in context of real-world code

### DIFF
--- a/src/ooxml/java/org/apache/poi/xwpf/usermodel/TableWidthType.java
+++ b/src/ooxml/java/org/apache/poi/xwpf/usermodel/TableWidthType.java
@@ -1,5 +1,22 @@
+/* ====================================================================
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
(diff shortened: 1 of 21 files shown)


asfgit:

src/ooxml/java/org/apache/poi/xwpf/usermodel/TableWidthType.java
```diff
@@ -0,0 +1,36 @@
+package org.apache.poi.xwpf.usermodel;
```
pjfanning, line 17: needs an apache license header

src/ooxml/java/org/apache/poi/xwpf/usermodel/TableWidthType.java
```diff
@@ -0,0 +1,36 @@
...
+    protected STTblWidth.Enum getSTWidthType() {
+        return this.type;
+    }
+    
+    /**
+     * Get the underlying STTblWidth enum value.
+     *
+     * @return STTblWidth.Enum value
+     */
+    public STTblWidth.Enum getStWidthType() {
```
pjfanning: should be `@Internal` or not public

pjfanning (changes requested):

PR #133, opened 2018-11-15
Hemf

hekebin:
hello, I have a mem  problem with  this code . pls check out for us  .
  
        FileInputStream is =null;
    	is= new FileInputStream(new File(path));
    	SlideShow slideShow = new  HSLFSlideShow(is);
    	SlideShowExtractor extractor = new SlideShowExtractor(slideShow);

the problem is that  this stream can’t reslese  the mem  although i have close it . just ppt & doc  file


asfgit:

kiwiwings:
I will merge the "hemf" branch via SVN, when it is ready.
Please use the mailing list for general questions.
The stream itself doesn't hold the memory, but the HSLFSlideShow keeps it in memory as a byte array and parsed objects.
There's no streaming parser for .ppt files - so if your .ppt is too large for your memory, you could only have a look at HSLFSlideShowImpl.buildRecords() and implement your own parser.

centic9:
CI-run is at https://builds.apache.org/view/P/view/POI/job/POI-Github-PullRequests/64/console

PR #141, opened 2019-01-23
Super-streaming SXSSF

mobreza:
https://bz.apache.org/bugzilla/show_bug.cgi?id=63100

SXSSF works as designed and manages a small memory footprint when generating large files from a database.  But it only writes data to an output stream once everything has been written to SXSSF. This is problematic when used in web applications:

In our use case (our website allows users to generate Excel from the database), generating the SXSSF on the server takes about 5 minutes. Most clients give up within a minute (or the browser does it automatically), or the proxy times out due to no data being sent. Some users also retry the download request. A new request for download is initiated (while the server is busy generating the SXSSF for a client that already gave up). This can potentially lead to DOS.

To work around this issue, I've implemented a super-streaming version of SXSSF, a `SuperSXSSF`, that relies on `rowWriter` callback to generate row data.

With this approach our service is able to stream the generated Excel directly to the client and, best of all, is terminated in case the user cancels the download request. 

The `SuperSXSSF` prevents both download timeouts and potential DOS, while allowing developers all other XSSF actions (i.e. define styles) that don't take much processing time.


asfgit:

mobreza:
The changes mostly make private vars protected for access in `SuperSXSSF`-* classes. Workbook is a copy of SXSSFWorkbook, but calls the `rowGenerator` before closing the stream.

src/ooxml/java/org/apache/poi/xssf/streaming/IRowGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public interface IRowGenerator {
+	
+	/**
+	 * Generate rows
+	 *
+	 * @param sheet the sheet
+	 * @throws Exception the exception
+	 */
+	void generateRows(SXSSFSheet sheet) throws Exception;
```
pjfanning: please use spaces for indents, not tabs

src/ooxml/java/org/apache/poi/xssf/streaming/SXSSFSheet.java
```diff
@@ -65,15 +65,23 @@ Licensed to the Apache Software Foundation (ASF) under one or more
 public class SXSSFSheet implements Sheet
 {
     /*package*/ final XSSFSheet _sh;
-    private final SXSSFWorkbook _workbook;
-    private final TreeMap<Integer,SXSSFRow> _rows = new TreeMap<>();
-    private final SheetDataWriter _writer;
+    protected final SXSSFWorkbook _workbook;
+    private final TreeMap<Integer,SXSSFRow> _rows=new TreeMap<Integer,SXSSFRow>();
```
pjfanning: this line change is unnecessary

src/ooxml/java/org/apache/poi/xssf/streaming/StreamingSheetWriter.java
```diff
@@ -0,0 +1,87 @@
...
+    public void close() throws IOException {
+        _out.flush();
+    }
+    
+    @Override
+    public InputStream getWorksheetXMLInputStream() throws IOException {
+    	throw new RuntimeException("Not supported with StreamingSheetWriter");
+    }
+    
+    @Override
```
pjfanning: no need for this finalize

src/ooxml/java/org/apache/poi/xssf/streaming/StreamingSheetWriter.java
```diff
@@ -0,0 +1,87 @@
...
+    protected void finalize() throws Throwable {
+    	// super.finalize();
+    }
+    
+    
+    @Override
+    boolean dispose() throws IOException {
+    	try {
```
pjfanning: I don't think exception should be swallowed

src/ooxml/java/org/apache/poi/xssf/streaming/StreamingSheetWriter.java
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+    public StreamingSheetWriter(OutputStream out) throws IOException {
+    	super(createWriter(out));
+        logger.log(POILogger.DEBUG, "Preparing SSXSSF sheet writer");
+    }
+
+    @Override
+    public File createTempFile() throws IOException {
+    	throw new RuntimeException("Not supported with StreamingSheetWriter");
```
pjfanning: could you ensure consistent indents of 4 spaces per indent?

src/ooxml/java/org/apache/poi/xssf/streaming/SheetDataWriter.java
```diff
@@ -71,6 +71,12 @@ public SheetDataWriter() throws IOException {
         _fd = createTempFile();
         _out = createWriter(_fd);
     }
+    
+    // Constructor for StreamingSheetWriter
```
pjfanning: this comment has little value

src/ooxml/java/org/apache/poi/xssf/streaming/StreamingSheetWriter.java
```diff
@@ -0,0 +1,87 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+
+import org.apache.poi.util.POILogFactory;
+import org.apache.poi.util.POILogger;
+
+public class StreamingSheetWriter extends SheetDataWriter {
```
pjfanning: could you add a little javadoc with `@since 4.1.0` annotation?

src/ooxml/java/org/apache/poi/xssf/streaming/SuperSXSSFSheet.java
```diff
@@ -0,0 +1,60 @@
...
+
+package org.apache.poi.xssf.streaming;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+import org.apache.poi.xssf.usermodel.XSSFSheet;
+
+public class SuperSXSSFSheet extends SXSSFSheet {
```
pjfanning: Can you add a little javadoc with `@since POI 4.1.0`? I'd prefer a more neutral descriptive class name.

src/ooxml/java/org/apache/poi/xssf/streaming/SuperSXSSFWorkbook.java
```diff
@@ -0,0 +1,346 @@
...
+import org.apache.poi.openxml4j.util.ZipEntrySource;
+import org.apache.poi.ss.usermodel.Sheet;
+import org.apache.poi.util.IOUtils;
+import org.apache.poi.util.POILogFactory;
+import org.apache.poi.util.POILogger;
+import org.apache.poi.xssf.usermodel.XSSFChartSheet;
+import org.apache.poi.xssf.usermodel.XSSFSheet;
+import org.apache.poi.xssf.usermodel.XSSFWorkbook;
+
+public class SuperSXSSFWorkbook extends SXSSFWorkbook {
```
pjfanning: Can you add a little javadoc with `@since POI 4.1.0`? I'd prefer a more neutral descriptive class name.

pjfanning (changes requested):
Thanks. The changes look useful. My biggest concern is that there are no tests included. Could you add some?

mobreza:
@pjfanning I've cleaned up indentation and removed my notes (comments). I'm preparing a few unit tests.  Feel free to suggest a better name than `SuperSXSSF*`.

pjfanning:
Thanks. I'm struggling to think of ideal names for the new classes. Maybe you could put an `@Beta` annotation on the new classes and maybe just use a name like SXSSF2Workbook. I suspect that we will be able to merge SXSSF2Workbook and SXSSFWorkbook but we can do that later.

mobreza:
@pjfanning I've recycled existing SXSSFWorkbook tests. It was helpful in identifying a couple of bugs.

I wanted to also reduce the amount of duplicated code from SXSSFWorkbook and because `copyStreamAndInjectWorksheet` is static I opted for `ISheetInjector` approach.

src/ooxml/testcases/org/apache/poi/xssf/streaming/TestSuperSXSSFWorkbook.java
```diff
@@ -0,0 +1,500 @@
...
+
+    @Test
+    public void addToExistingWorkbook() throws IOException {
+    	XSSFWorkbook xssfWb1 = new XSSFWorkbook();
+    	xssfWb1.createSheet("S1");
+    	Sheet sheet = xssfWb1.createSheet("S2");
+    	Row row = sheet.createRow(1);
+    	Cell cell = row.createCell(1);
+    	cell.setCellValue("value 2_1_1");
+    	SuperSXSSFWorkbook wb1 = new SuperSXSSFWorkbook(xssfWb1);
```
pjfanning: inconsistent indents in whole class

src/ooxml/testcases/org/apache/poi/xssf/streaming/TestSuperSXSSFWorkbook.java
```diff
@@ -0,0 +1,500 @@
...
+    public void existingWorkbook() throws IOException {
+    	XSSFWorkbook xssfWb1 = new XSSFWorkbook();
+    	xssfWb1.createSheet("S1");
+        SuperSXSSFWorkbook wb1 = new SuperSXSSFWorkbook(xssfWb1);
+    	XSSFWorkbook xssfWb2 = SXSSFITestDataProvider.instance.writeOutAndReadBack(wb1);
+    	assertTrue(wb1.dispose());
+
+    	SuperSXSSFWorkbook wb2 = new SuperSXSSFWorkbook(xssfWb2);
+    	assertEquals(1, wb2.getNumberOfSheets());
+    	Sheet sheet  = wb2.getStreamingSheetAt(0);
```
pjfanning: 2 spaces before =

src/ooxml/java/org/apache/poi/xssf/streaming/IRowGenerator.java
```diff
@@ -0,0 +1,35 @@
...
+@Beta
+public interface IRowGenerator {
+    
+    /**
+     * Generate and add rows to the sheet
+     *
+     * @param sheet the sheet
+     * @throws Exception the exception
+     */
+    void generateRows(SXSSFSheet sheet) throws Exception;
```
pjfanning: I don't quite understand this - why not just add rows using Sheet#createRow?

https://github.com/apache/poi/pull/141/files#diff-ea5bbd436cd5c48efd0716b8d6f0e6ebR155 (highlight line in yellow) - seems a very verbose way to create a row

pjfanning (changes requested):

src/ooxml/java/org/apache/poi/xssf/streaming/IRowGenerator.java
mobreza: @pjfanning Row and cell generation happens in `SXSSFWorkbook#injectData()` and `#copyStreamAndInjectWorksheet()` and that allows for more direct streaming to the client. A call to `SuperSXSSFSheet#createRow` outside `generateRows()` is invalid as `SuperSXSSFSheet` does not have a `_writer` until `workbook.write(out);` is running.

https://github.com/apache/poi/pull/141/files#diff-ea5bbd436cd5c48efd0716b8d6f0e6ebR274 in `populateData()` demonstrates it better.

mobreza: Not 100% true: you can call `#createRow` as long as you stay within `randomAccessWindowSize` of SXSSFSheet.

pjfanning:
xlsx files are zip files that contain XML files. There is an XML file for each sheet. There are also XML files for 'shared strings', styles and comments. How does this code base support these requirements?

mobreza:
@pjfanning SXSSF utilizes lots of XSSF, but avoids OutOfMemory exception by writing row data to a temporary file, the contents of which is injected at the right place when the final xlsx XML+ZIP is written to the output stream:

1. Prepare XSSF (styles, comments, sheet info....) in memory
1. Prepare entire SXSSF sheet XML content -- write row XML to a temporary file (this may take minutes)
1. Write XSSF to a new temporary xlsx file (contains everything but the row data from SXSSF)
1. Copy temporary xslx contents to output stream
1. ... inject row XML data from the temporary file (`SXSSFWorkbook#injectData()`)

Step 1 is quick. Most time is spent in step 2 to write XML row data to a temporary file. Steps 3 is quick. Steps 4 and 5 are quick, only copying existing XML from temporary files.

Writing the final xlsx to output stream begins in step 4, but is delayed (for minutes) until we've prepared the temporary file with all SXSSF row data.

SuperSXSSF postpones the generation of the sheet XML contents:

1. Prepare XSSF (styles, comments, sheet info....) in memory
1. Declare how SXSSF sheet XML content is generated using `setRowGenerator()`, but do not spend time generating row XML
1. Write XSSF to a new temporary xlsx file (contains everything but the row data from SXSSF)
1. Copy temporary ZIP+XML file contents to output stream
1. ... and generate and inject SXSSF XML row data on the fly (this may take minutes).

Steps 1 is quick. Step 2 is just a declaration and is executed quickly. Step 3 is quick. Step 4 can start copying XML from temporary files to the final output stream. Most time is spent in step 5 for generation of XML row data.

Writing the final xslx to output stream begins in step 4, but we have not spent any time generating SXSSF row data.

pjfanning:
Could you add a sample to src/examples to demo the usage?

mobreza:
@pjfanning Yes, I'll add some.

I've been thinking about `SXSSF2` that could be a decent hybrid of `SXSSF` and `SuperSXSSF`: keep the declarative nature of `SuperSXSSF` (postponing row generation until needed) and allow for all operations on `XSSFWorkbook` (styles, shared strings, etc.). A two-phase `XSSFWorkbook#write(out)` is then used:

1. write the `XSSFWorkbook` to temporary file
1. copy from temporary file to the final output stream the basic elements of the ZIP that are required for xlsx compatibility, but skip  styles, shared strings, etc.
1. write sheet data to the final output stream (including `SuperXSSF` delayed sheets (and here we could declare new styles, shared strings, etc.).
1. the `XSSFWorkbook` now contains updated styles, strings, etc. 
1. write the workbook with updates to another temporary file
1. copy the updated shared resources from this temporary file to the final output stream

This would allow us to start streaming xlsx immediately, make changes to common XMLs and still write out the full Excel file.

centic9:

jandhollander:
@mobreza @pjfanning Is this something that could make it in, in the future?
We have a web application that uses SXSSF and would like the download of a workbook to start immediately while generating the workbook.

pjfanning:
This PR has been corrupted but if someone wants to create a new one, we might consider it.

mobreza:
Is there a way to reopen this PR?

pjfanning:
@mobreza could you create a new MR from your branch?

PR #143, opened 2019-04-05
MAPIType.isFixedLength: not true in case of length > 8

dhoelzl:
`MAPIType.isFixedLength`: not true in case of length > 8 as fixed length properties are always stored as 8 bytes (zero-padded if length < 8) and cannot exceed 8 bytes.

Writing e.g. `MAPIType` `CLS_ID` which has length 16 would fail otherwise.
(Defined here: https://github.com/apache/poi/blob/f509d1deae86866ed531f10f2eba7db17e098473/src/scratchpad/src/org/apache/poi/hsmf/datatypes/Types.java#L69)


asfgit:

kiwiwings:
Patch applied via r1857708, thank you for bringing this up.
First I thought about having an additional attribute or specific handling of CLS_ID, but for custom types your solution seems to be straight forward.

PR #145, opened 2019-05-10
fix: PackagingURIHelper.encode(性) get wrong value

greedbell:
`PackagingURIHelper.encode(性) ` return `%E6%A7`, but it should be `%E6%80%A7`


asfgit:

pjfanning:
Merged using https://github.com/apache/poi/commit/a805c30efb157646ce2a833a43b3913fd8bb69c0

PR #149, opened 2019-05-23
improve MAPIMessage.guess7BitEncoding, improve MAPIMessage.getHtmlBody

dhoelzl:
improve MAPIMessage.guess7BitEncoding

 - general properties code page sources: PR_MESSAGE_CODEPAGE, default ANSI code page of PR_MESSAGE_LOCALE_ID, charset from content-type header
 - code page source for PR_BODY and PR_BODY_HTML: PR_INTERNET_CPID (as documented)
   no need to get from HTML; Outlook skips that too
   special case: outlook uses code page CP1252 instead of utf-8 for PR_BODY

improve MAPIMessage.getHtmlBody:

 - in case if binary property use code page defined in PR_INTERNET_CPID, default to CP1252 (no need to get from HTML; Outlook skips that too)
 - in case of string property rely on code page set by MAPIMessage.guess7BitEncoding


asfgit:

pjfanning (commented):
Do you have any sample files to try the new code on? If so, can you use them in unit tests (for regression purposes)?

dhoelzl:
I have added a unit test to the PR.

pjfanning:
Closed using https://github.com/apache/poi/commit/721180d35f2777c38a7d8206b5d1555a82f92acb

PR #150, opened 2019-05-26
XWPFPicture: easy access to width and depth

dourouc05:
I suppose adding a test case wouldn't hurt (something like calling XWPFRun.addPicture then calling these two new methods).

(BTW, would a pull request to add an overload to XWPFRun.addPicture be accepted? It would be to compute the width and depth from the input file, using javax.ImageIO -- an acceptable dependency or not?)

### DIFF
--- a/src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFPicture.java
+++ b/src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFPicture.java
@@ -77,6 +77,20 @@ public class XWPFPicture {
         }
         return null;
     }
+   
+    /**
+     * Returns the width of the picture (in EMU). 
+     */
+    public long getWidth() {
+        return ctPic.getSpPr().getXfrm().getExt().getCx();
+    }
+   
+    /**
+     * Returns the width of the picture (in EMU). 
+     */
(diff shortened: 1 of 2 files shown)


asfgit:

src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFPicture.java
```diff
@@ -77,6 +77,20 @@ public XWPFPictureData getPictureData() {
...
+   
+    /**
+     * Returns the width of the picture (in EMU). 
+     */
+    public long getWidth() {
+        return ctPic.getSpPr().getXfrm().getExt().getCx();
+    }
+   
+    /**
+     * Returns the width of the picture (in EMU).
```
Alain-Bearez: `width` was copied from above.

Most of the other API use `Units.toPoints()` to convert out of EMU to more human understandable points.

dourouc05:
I've just added a test for the new methods. Is this enough for merging?

dourouc05:
@pjfanning @Alain-Bearez is this OK to merge?

pjfanning:
merged using https://svn.apache.org/repos/asf/poi/trunk@1863434

PR #152, opened 2019-06-01
XWPFParagraph: direct control over keepNext.

dourouc05:


asfgit:

src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFParagraph.java
```diff
@@ -402,6 +402,24 @@ public BigInteger getNumStartOverride() {
         }
         return null;
     }
+    
+    /**
+     * Indicates whether this paragraph should be kept on the same page as the next one.
+     */
+    public boolean isKeepNext() {
+        if (getCTP() != null && getCTP().getPPr() != null) {
```
pjfanning: does this compile? seems to me, this has no return if the if-condition fails

dourouc05: I've corrected this and added a test. Better?

pjfanning:
this is merged - https://github.com/apache/poi/pull/156 had the same code and I merged it there

PR #153, opened 2019-06-04
XWPFParagraph: easier way to create a link.

dourouc05:
Inspiration from https://stackoverflow.com/questions/55275241/how-to-add-a-hyperlink-to-the-footer-of-a-xwpfdocument-using-apache-poi

### DIFF
--- a/src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFDocument.java
+++ b/src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFDocument.java
@@ -273,6 +273,7 @@ public class XWPFDocument extends POIXMLDocument implements Document, IBody {
         // Get the hyperlinks
         // TODO: make me optional/separated in private function
         try {
+            hyperlinks = new ArrayList<>();
             for (PackageRelationship rel : getPackagePart().getRelationshipsByType(XWPFRelation.HYPERLINK.getRelation())) {
                 hyperlinks.add(new XWPFHyperlink(rel.getId(), rel.getTargetURI().toString()));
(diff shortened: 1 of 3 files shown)


asfgit:

dourouc05:
I've just added a test; is this OK to merge?

pjfanning (changes requested):
Test fails for me.

```
Testcase: testAddHyperlink took 0.007 sec
	Caused an ERROR
null
java.lang.NullPointerException
	at org.apache.poi.xwpf.usermodel.TestXWPFDocument.testAddHyperlink(TestXWPFDocument.java:194)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
```

dourouc05:
The DOCX always got generated as it should have, but retrieving a newly added link could not work. What I did is call initHyperlinks in XWPFDocument each time a link is requested by ID, in case a new link was added (first emptying the `hyperlinks` field). 

Is this approach correct? From my understanding, I think so… Performancewise, it seems problematic, as the list is rebuilt each time a request for a link is made, this is why I made a second commit to rebuild this list only when the link is not found (which is already better).

pjfanning:
Merged using https://svn.apache.org/repos/asf/poi/trunk@1863433

PR #154, opened 2019-06-12
[bug-57342] Excel compatible Zip64 implementation

rzymek:
https://bz.apache.org/bugzilla/show_bug.cgi?id=57342 

I did an in depth analysis of this issue. Turns out the problem is not with the OOXML data generated by POI. The problem has to do with the ZIP format. Specifically with ZIP64 extension. That's why it's all OK up until sheet1.xml reaches over 4GB (uncompressed). 
I have all the details written up in a blog post: https://rzymek.github.io/post/excel-zip64/
Short story: Excel will want to repair the file if uncompressed size of a zip entry exceeds 4GB and ZIP's Local File Header (LFH) does not specify zip spec version 4.5

This pull request uses custom (Excel compatible) Zip64 implementation when Zip64Mode is set to Always.

### DIFF
--- a/src/ooxml/java/org/apache/poi/xssf/streaming/OpcOutputStream.java
+++ b/src/ooxml/java/org/apache/poi/xssf/streaming/OpcOutputStream.java
@@ -1,0 +1,124 @@
+package org.apache.poi.xssf.streaming;
+
+
+import org.apache.poi.xssf.streaming.Zip64Impl.Entry;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.zip.*;
+
+/**
+ * ZIP64 OutputStream implementation compatible with MS Excel.
+ * Drop in replacement for `java.util.ZipOutputStream`.
+ *
+ * For more information see https://github.com/rzymek/opczip
+ *
+ * @author Krzysztof Rzymkowski
+ */
+class OpcOutputStream extends DeflaterOutputStream {
+
+    private final Zip64Impl spec;
+    private final List<Entry> entries = new ArrayList<>();
+    private final CRC32 crc = new CRC32();
+    private Entry current;
+    private int written = 0;
+    private boolean finished = false;
+
+    /**
+     * Creates ZIP64 output stream
+     *
+     * @param out target stream to write compressed data to
+     */
+    public OpcOutputStream(OutputStream out) {
+        super(out, new Deflater(Deflater.DEFAULT_COMPRESSION, true));
+        this.spec = new Zip64Impl(out);
+    }
+
+    /**
+     * @see Deflater#setLevel(int)
+     */
+    public void setLevel(int level) {
+        super.def.setLevel(level);
+    }
+
+    /**
+     * @see ZipOutputStream#putNextEntry(ZipEntry)
+     */
+    public void putNextEntry(String name) throws IOException {
+        if (current != null) {
+            closeEntry();
+        }
+        current = new Entry(name);
+        current.offset = written;
+        written += spec.writeLFH(current);
+        entries.add(current);
+    }
+
+    /**
+     * @see ZipOutputStream#closeEntry()
+     */
+    public void closeEntry() throws IOException {
+        if (current == null) {
+            throw new IllegalStateException("not current zip current");
+        }
+        def.finish();
+        while (!def.finished()) {
+            deflate();
+        }
+
+        current.size = def.getBytesRead();
+        current.compressedSize = (int) def.getBytesWritten();
+        current.crc = crc.getValue();
+
+        written += current.compressedSize;
+        written += spec.writeDAT(current);
+        current = null;
+        def.reset();
+        crc.reset();
+    }
+
+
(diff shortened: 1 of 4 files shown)


asfgit:

src/ooxml/java/org/apache/poi/xssf/streaming/OpcOutputStream.java
```diff
@@ -0,0 +1,124 @@
+package org.apache.poi.xssf.streaming;
```
pjfanning: can you add the apache license to all files?

rzymek: sure. done.

pjfanning:
thanks - merged with https://svn.apache.org/repos/asf/poi/trunk@1861196

bosofelipe:
@rzymek Could you give me an example because i try generate a large excel with 37000 rows and 2500 coluns and file still corrupted using apache 4.1.2

bosofelipe:
@rzymek my current code failed

`import java.io.File;
import java.io.FileOutputStream;

import org.apache.commons.compress.archivers.zip.Zip64Mode;
import org.apache.poi.ss.usermodel.Row;
import org.apache.poi.xssf.streaming.SXSSFWorkbook;

public class TestExcel {

	public static void main(String[] args) throws Exception {
		SXSSFWorkbook workbook = null;
		workbook = new SXSSFWorkbook(null, 10000, true, false);
		workbook.setZip64Mode(Zip64Mode.Always);
		int ROWS_COUNT = 37000;
		int COLS_COUNT = 2500;

		org.apache.poi.ss.usermodel.Sheet sheet = workbook.createSheet("1");
		for (int i = 1; i <= ROWS_COUNT; i++) {
			Row row = sheet.createRow(i);
			for (int j = 1; j <= COLS_COUNT; j++) {
				row.createCell(j).setCellValue("1");
				System.out.println(i + " " + j);
			}
		}

		FileOutputStream out = new FileOutputStream("~/Downloads/Excel.xlsx");
		workbook.write(out);
		out.close();
		workbook.close();
		workbook.dispose();

		File file = new File("~/Downloads/Excel.xlsx");
		System.out.println(file.length());

	}
}`

pjfanning:
Zip64Mode.AsNeeded is more correct - always may mean you use zip64 mode when you don't need it

rzymek:
Are you getting "corrupted file" error from Excel or OpenOffice or something else? OpenOffice Calc  has a [limit of 1024 columns](https://simple.wikipedia.org/wiki/OpenOffice_Calc#Specifications) (Excel's limit is 16k columns). Other than that, the code looks ok.

Zip64Mode needs to be Always in this case to enable ZIP64 handling compatible with Excel.

pjfanning:
@rzymek thanks for clarifying - do you know what effect setting Zip64Mode.Always has if you create a small spreadsheet - will this file cause problems for Excel?

rzymek:
As far as I checked, Zip64Mode.Always does not cause problem with Excel even in small files. When it comes to Excel and big files (XML over 4Gb), then ZIP64 must be declared in the zip entry header before the actual zip entry contents.

pjfanning:
Thanks @rzymek - we might want to make Zip64Mode.Always the default - needs some experimentation before we'd make that change though

rzymek:
Exactly. I think that custom zip64 implementation should sit as an option for a few versions (it's only enabled when Zip64Mode.Always).

bosofelipe:
@rzymek I tested with Libre Office... now i tested with MS Excel and the problem was solved, Is a limitation of Libre office, with you told us? 

Tks a lot!

pjfanning:
@rzymek We recently had a user start a thread about the Zip64 support in POI
* https://lists.apache.org/thread/fr1ddq8j4047jc9c73y531g6o8oqlj4x
I also raised:
* https://bz.apache.org/bugzilla/show_bug.cgi?id=69400

Commons-Compress added support for Zip64Mode.AlwaysWithCompatibility since this was added to POI.
At least for the reporter on the lists.apache.org thread, using Zip64Mode.AlwaysWithCompatibility with the standard ZipArchiveOutputStream sorted their problem. This bypasses the OpcZipArchiveOutputStream class that you wrote.

You may no longer have an interest in this topic but I'm wondering if you still have interest, would you be able to help in evaluating ZipArchiveOutputStream with Zip64Mode.AlwaysWithCompatibility ?

manticore-projects:
> Commons-Compress added support for Zip64Mode.AlwaysWithCompatibility since this was added to POI. At least for the reporter on the lists.apache.org thread, using Zip64Mode.AlwaysWithCompatibility with the standard ZipArchiveOutputStream sorted their problem. 

We found out that it does not solve it since at least EXCEL-2016 can't open those files (although LibeOffice will open those, with all the new checks passing.)


> This bypasses the OpcZipArchiveOutputStream class that you wrote.

This particular class seem to write "holes/unallocated blocks" as shown here https://bugs.documentfoundation.org/show_bug.cgi?id=163384#c5
If this problem could be solved we would have the perfect solution.

PR #155, opened 2019-06-21
Update XWPFHeaderFooterPolicy.java

souhaib100:
Add case when there is already header in the docx.


asfgit:

pjfanning (commented):
Thanks. Would it be possible to add a test case?

souhaib100:
Where?

pjfanning:
maybe in poi/src/ooxml/testcases/org/apache/poi/xwpf/usermodel/TestXWPFHeader.java

Tests can be with `ant clean test`

centic9:

PR #158, opened 2019-08-02
Read-only Row-like view of columns

luca-vercelli:
I often have to loop over cells of a given column.
I think that it would be useful to have an object representing a column.
Here, I created the class Column, and I added a method getColumn() in the Sheet interface.


asfgit:

src/java/org/apache/poi/ss/usermodel/Column.java
```diff
@@ -0,0 +1,333 @@
...
+/**
+ * A read-only Row view of a sheet column.
+ * 
+ * Intended to iterate over cells of a sheet.
+ *
+ * Get this object with Sheet.getColumn(colNum).
+ *
+ * @author luca vercelli 2018
+ */
+public class Column implements Row {
```
donalmurtagh, line 20: `Row` shouldn't be implemented because a column is obviously not a row. On account of implementing this interface there are various methods in this class which don't make any sense

```java
    @Override
    public int getRowNum() {
        return colNum;
    }

    @Override
    @NotImplemented
    public short getHeight() {
        throw new IllegalStateException("Not implemented");
    }

    @Override
    @NotImplemented
    public float getHeightInPoints() {
        throw new IllegalStateException("Not implemented");
    }
```

donalmurtagh (changes requested):

luca-vercelli:
"obviously" not so much, IMHO. A column /is/ a row, intended as a
1-dimensional array of cells. There are situations where you want to
iterate on the cells of a single column.


Il mer 1 apr 2020, 19:54 Dónal Murtagh <notifications@github.com> ha
scritto:

PR #159, opened 2019-09-27
multiple paras in Word document comments are conflated

cmacdonald:
I propose adding a newline after each paragraph.


centic9:

cmacdonald:
I dont seem to have a re-open button. Should I make a new pull request?

centic9:
Then something really bad happened to this repo as also the diff does not work any more for me, can you try what happens if you open a new PR?

PR #160, opened 2019-09-29
Add support for new java.time API, deprecate HSSFDateUtil, move test cases to org.apache.poi.ss.usermodel

xzel23:
While preparing for adding support for the new Java date/time API I noticed that DateUtil is extended in HSSFDateUtil which itself doesn't add any functionality. Sole purpose seems to be making the protected method `absoluteDay()` accessible for unit testing in package org.apache.poi.hssf.usermodel. This should not be necessary as the test code should live in the same package as where the methods are defined and also makes maintenance harder.

I left the TestHSSFDateUtil with `onARealFile()`, the only method that really belongs there because it has a dependency on HSSF.

Instead of completely removing HSSFDateUtil, I marked it as deprecated and changed static access to DateUtil throughout the codebase. Since HSSFDateUtil is public, it should not be removed before the next major version.

Please review and comment. If possible, I'd like to have this applied to trunk soon so that I can focus on the real task, aka adding the new date/time API.


xzel23:
For more detail, see [Bug 63779 - RFE: Add support for the new Java date/time API added in Java 8](https://bz.apache.org/bugzilla/show_bug.cgi?id=63779).

centic9:
CI-build is running at https://builds.apache.org/view/P/view/POI/job/POI-DSL-Github-PullRequests/30/

xzel23:
PullRequests/src/scratchpad/testcases/org/apache/poi/hwpf/usermodel/TestBugs.java:49: error: package org.apache.poi.ooxml.util does not exist
    [javac] import org.apache.poi.ooxml.util.DocumentHelper;
    [javac]                                 ^
    [javac] 1 error

I had that error on my clean checkout of POI, so I commented out the missing import and affected test case. Thought it might be to a dependency that had not been publicly updated. How to go on?

xzel23:
Ok, in the meantime, I completed support for LocalDateTime and LocalDate in the spreadsheet component. I have never worked with the other parts of the library, so if support is needed in other places too, just point me to it.

centic9:
The import-failure is caused because the module "scratchpad" does not have a dependency to the "ooxml" module, but you try to use code from that module in tests.

Your IDE seems to not take that into account, but the ant-based build does.

centic9:
CI: https://builds.apache.org/view/P/view/POI/job/POI-DSL-Github-PullRequests/31

xzel23:
No, it’s not me who is trying - POI is. I didn’t change that part of the code. I have investigated the cause and found out that it’s because the GitHub code isn’t synched. In the meantime 23 commits are missing from the GitHub code base. Please have a look at bug 63794. Github stopped syncing on august 31. Even a clean checkout fails whereas a checkout from SVN builds without problem.

I could also add the dependency for the build, but that would just be dressing the symptoms, not the cause and would later probably cause problems when merging into SVN.

centic9:
Oh, ok, sorry, I thought this would be solved in the meantime, but obviously it is not. For reference this is handled at https://bz.apache.org/bugzilla/show_bug.cgi?id=63794 and https://issues.apache.org/jira/plugins/servlet/mobile#issue/INFRA-19154

xzel23:
Since there’s still no information when the mirroring issue will be resolved, I could create a patch against SVN trunk instead. I’d then look for other things to do on POI (I have some time between projects). What do you think?

src/java/org/apache/poi/hssf/usermodel/HSSFCell.java
```diff
@@ -42,14 +43,7 @@ Licensed to the Apache Software Foundation (ASF) under one or more
...
 import org.apache.poi.ss.formula.ptg.Ptg;
-import org.apache.poi.ss.usermodel.CellBase;
-import org.apache.poi.ss.usermodel.CellStyle;
-import org.apache.poi.ss.usermodel.CellType;
-import org.apache.poi.ss.usermodel.CellValue;
-import org.apache.poi.ss.usermodel.Comment;
-import org.apache.poi.ss.usermodel.FormulaError;
-import org.apache.poi.ss.usermodel.Hyperlink;
-import org.apache.poi.ss.usermodel.RichTextString;
+import org.apache.poi.ss.usermodel.*;
```
centic9: We will adjust this when applying the diff, but in general please try to not change imports when not necessary, otherwise we would play import-ping-pong with changes by others.

centic9:
I did a quick review and changes look very good, thank you working on those. 

You don't need to do an SVN-based patch, we can download a patch-file from Github and apply it to SVN easily and it should still be valid even while the mirror is out of sync.

On other possible work I'll send you some information via Email to keep it out of this more focused discussion here.

centic9:
this is applied now via SVN r1868198

PR #161, opened 2019-10-09
keep different paragraphs of comments separate

cmacdonald:
Replacement for PR #159


centic9:
We applied this in a slightly different way to not cause a trailing newline always, please check if this works for you. 

See https://svn.apache.org/viewvc?view=rev&rev=1868978

cmacdonald:
Looks good to me, thanks!

PR #162, opened 2019-11-12
fix documentation

m2nhawk:
t = true - it returns the periodic payment, when the payments are due at the beginning of the period, but the documentation means it should return the periodic payment, when the payments are due at the end of the period.


pjfanning:
updated using https://github.com/apache/poi/commit/2d6b1800dc8c7c84429a83a36329de8489c0b307 if you'd like to review

m2nhawk:
Ok. You should edit a documentation [here](https://poi.apache.org/apidocs/dev/org/apache/poi/ss/formula/functions/FinanceLib.html) too.

PR #163, opened 2019-12-03
Add, set the level to a numbering paragraph on XWPFParagraph

miguoliang:
To set the level of a paragraph in a multilevel numbering list by `setNumILvl`.


miguoliang:
People can Set the level of a paragraph in a multilevel numbering list in Microsoft Word Docx File manually by this way https://www.miguoliang.com/apache-poi-multilevel-numbering-paragraph-microsoft-word-docx.html before my PR is merged.

However, I found that `numPr` tag does not appear in *document.xml* directly in the most of docx generated by MS word with a large number of styles, it is wrapped in a style definition in *styles.xml*.

Current Numbering APIs are good to generate a word document from zero, but maybe not very great for general *docx* files generated by any other software, such as Microsoft Word.

pjfanning:
Merged using https://svn.apache.org/repos/asf/poi/trunk@1871999

PR #164, opened 2019-12-05
Fix Bug in XSSFTable.setCellReferences when table is single cell

travisrussell:
Calling `XSSFTable.setCellReferences` throws an `ArrayIndexOutOfBoundsException` in the case when the table is only one cell.

`XSSFTable.setCellReferences` sets the table's `startCellReference` and `endCellReference`.
It fails because it expects a colon in the table reference, but in the one cell case there is no colon. In this case, it is simply a single cell, e.g., `A2`. This causes the exception to be thrown when setting the `endCellReference`, as the code expects a second part of the table reference.

This fix is to check the parts after we split the reference. If there is only one part, set the end reference to the same value as the start reference.


pjfanning:
merged using https://github.com/apache/poi/commit/f12a414662702c56ccb6b3b4339465f71dcfe903 and https://github.com/apache/poi/commit/4c034f0e202ab6c3a56a2c30716d2c56f9be3311

PR #166, opened 2020-01-28
Add support for invertIfNegative in bar charts

nikeee:
The XML element "invertIfNegative" of the bar chart series is not exposed to the user.


Alain-Bearez:
merged by r1873282

nikeee:
Thanks!

PR #167, opened 2020-02-20
HSMF enhancements

dhoelzl:
- introduce NameIdChunks.GetPropertyTag which enables evaluating property ids from properties identified by name/id in property sets
  (simple version of IMAPIProp::GetIDsFromNames)
- AttachmentChunks.getAttachData: use new ByteStreamChunk instead of ByteChunk which enables delayed reading of attachments
  to avoid all attachments are completely read into memory when parsing which may cause OutOfMemoryErrors on e-mails with big attachments.
- POIFSChunkParser: support reading multi valued chunks (e.g. required when reading the Keywords ("categories") property)
- add MAPIProperty.RECEIVED_BY_SMTP_ADDRESS
- add unit tests


dhoelzl:
Thank you for applying this!

PR #170, opened 2020-03-23
XWPFTableCell does not process bodyElements when handle paragraph

Sayi:
When I used XWPTFTableCell in depth, I found this bug.


Alain-Bearez:
applied by commit r1875746

This PR can be closed.

PR #171, opened 2020-03-23
XWPFNumbering.addAbstractNum will definitely throw an exception

Sayi:
XWPFNumbering.addAbstractNum code is an if branch. There are two points to note:
1. When `abstractNum.getAbstractNum ()` is not null, you must ensure that AbstractNumId has a value, so in order for users to find the correct AbstractNumId value, two getter methods are exposed, of course, the list cannot be modified.

2. When `abstractNum.getAbstractNum ()` is null, an exception will be thrown when cascading the method of `abstractNum.getAbstractNum ()`, so you need to set the attribute CtAbstractNum of the abstractNum.

PS: For the first point, we can also assert in the `XWPFNumbering.addAbstractNum` that AbstractNumId is not null, fail fast.


Alain-Bearez:
applied by commit r1875747

this PR can be closed

PR #172, opened 2020-03-24
#63624: setText in XWPFTableCell updates the xml and also updates runs

Sayi:
Use an existing XWPFParagraph or create a new XWPFParagraph


Alain-Bearez:
applied by commit r1875744

this PR can be closed

PR #173, opened 2020-03-25
allow add and remove a HyperlinkRun or a FieldRun in XWPFParagraph

Sayi:
### What is the pr?
- add `createFieldRun`method
- add `insertNewHyperlinkRun` method
- add `insertNewFieldRun` method
- refactor `insertNewRun` method
- refactor `removeRun` method, Add support for removing HyperlinkRun and FieldRun
- add some testcases
- update examples/SimpleDocument, add hyperlink case

### How to work
`XWPFHyperlinkRun` and `XWPFFieldRun` both inherit `XWPFRun`, In the future, this code design should be re-discussed, but now I maintain the original data structure, and the code is also compatible with the previous use.

1. Since `w: p` can contain many child elements, I think the simple way should be to insert run through cursor instead of` CTP.insertNewR (int) `. If the currently inserted position is an `XWPFHyperlinkRun` or `XWPFFieldRun`, it will look up for the correct cursor position

2. When removing a run at a certain location, if it is an `XWPFHyperlinkRun` or `XWPFFieldRun` and they contain multiple Runs, the run will be removed directly. When there is only one run, the hyperlink or fieldrun will be removed


Alain-Bearez:
applied by commit r1875862

this PR can be closed

PR #177, opened 2020-04-23
add null checker to sourceRow so it can be skipped if it's null

xjlin0:
When copying the sheet in the memory, if the row is null, currently it will cause NPE. So just a null checker to protect it, similar to how it skips null columns as of now.   @zmau 
```
2020-04-23 09:49:07.390 [,myservice,,] ERROR https-jsse-nio-9420-exec-2 [dispatcherServlet] - Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is java.lang.NullPointerException] with root cause
java.lang.NullPointerException: null
        at org.apache.poi.ss.usermodel.RangeCopier.copyRange(RangeCopier.java:89)
        at org.apache.poi.ss.usermodel.RangeCopier.copyRange(RangeCopier.java:68)
```

### DIFF
--- a/src/java/org/apache/poi/ss/usermodel/RangeCopier.java
+++ b/src/java/org/apache/poi/ss/usermodel/RangeCopier.java
@@ -85,7 +85,9 @@ public abstract class RangeCopier {
         
         for(int rowNo = sourceRange.getFirstRow(); rowNo <= sourceRange.getLastRow(); rowNo++) {   
             Row sourceRow = sourceClone.getRow(rowNo); // copy from source copy, original source might be overridden in process!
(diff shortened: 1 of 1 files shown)


pjfanning:
Would you be able to provide a test xlsx file and a test case that reproduces the issue? We like to have regression tests.

centic9:
Sorry, I was too fast, I could quickly reproduce it and fix it as part of some other improvement to the tests.

xjlin0:
it's ok, as long as the null row can now be skipped and will be release in the next version, I will be happy.

PR #178, opened 2020-04-26
Make isLatentStyle() public

drmacro:
This very small enhancement makes it possible to determine if a given style name is a latent style.

I need this in Wordinator in order to report attempts to use latent styles as distinct from styles that are simply not defined at all.


pjfanning:
thanks - merged using https://github.com/apache/poi/commit/18c98662aa0f8c6b1ecd1fddb88a86513153aa54

PR #179, opened 2020-04-30
add an option for RangeCopier.copyRange() also clone styles

xjlin0:
Found out that currently RangeCopier.copyRange() only copy values and formulas, however it always pass hardcoded null styleMap to RangeCopier.cloneCellContent(), therefore no styles were copied.  Not so sure why, so I added a boolean option for RangeCopier.copyRange() to copy styles.

The MR provides an additional argument on the current RangeCopier.copyRange(), while passing the boolean argument is optional. So it will be compatible with all existing codes using RangeCopier.copyRange().

```
myRangeCopier.copyRange(originalDataRange, destinationDataRange, true);
// the additional true enables style copying

myRangeCopier.copyRange(originalDataRange, destinationDataRange);
// works as before, copy range without copying styles
```

Tested in my company and the cell styles can be copied by copyRange(). Somehow we also need to include `compile("org.apache.commons:commons-math3:3.0")` for it to copy range with styles, probably due to local jar files including.

### DIFF
--- a/src/java/org/apache/poi/ss/usermodel/RangeCopier.java
+++ b/src/java/org/apache/poi/ss/usermodel/RangeCopier.java
@@ -19,6 +19,7 @@
 
 package org.apache.poi.ss.usermodel;
 
+import java.util.HashMap;
 import java.util.Map;
 import org.apache.poi.ss.formula.FormulaShifter;
 import org.apache.poi.ss.formula.ptg.Ptg;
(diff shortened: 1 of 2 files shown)


centic9:
Can you add a test-case which verifies the new option?

xjlin0:
Sure, I updated a test case. Here is the passing results: Tests passed: 15 of 15 tests - 2s 304ms.
```
BUILD SUCCESSFUL in 6s
22 actionable tasks: 13 executed, 9 up-to-date
3:38:39 PM: Task execution finished ' :main:test --tests "org.apache.poi.hssf.usermodel.TestHSSFRangeCopier" :ooxml:test --tests "org.apache.poi.ss.usermodel.TestXSSFRangeCopier" --continue'.
```
One follow up question for you @centic9  , I found out that the RangeCopier.copyRange() not only cannot copy styles but also cannot copy merged cells.  So I am adding options to that function. Do you think the following signature looks good? Where could I also update [the document](https://poi.apache.org/apidocs/dev/org/apache/poi/ss/usermodel/RangeCopier.html#copyRange-org.apache.poi.ss.util.CellRangeAddress-org.apache.poi.ss.util.CellRangeAddress-)? Thanks!

```
copyRange(
   CellRangeAddress tilePatternRange,
   CellRangeAddress tileDestRange,
   boolean copyStyles,
   boolean copyMergedRanges
)
```
Or should we just forget about the boolean option, make it always copy styles & merged ranges?

pjfanning:
Thanks - merged

PR #182, opened 2020-05-19
Fix root property size calculation.

netmackan:
Before this patch the POIFSMiniStore.syncWithDataStore() method sets the root property size to be based on the number of sectors in each block that is used. 

However it can happen is some files that a block is not only filled in the beginning. Instead there could be gaps with unused sectors within it and currently POI does not count those. However the MS tool (in this case signtool) calculates the occupied size (i.e. the size from beginning of block up to and including the last used sector). Specifically for Windows Installer (MSI) files, if the root size is not calculated this way the digital signature will not show up as correct.

This pull request changes the calculation of the root property so that Windows will not complain also for files with blocks that has unused sectors in the middle.

### DIFF
--- a/src/java/org/apache/poi/poifs/filesystem/POIFSMiniStore.java
+++ b/src/java/org/apache/poi/poifs/filesystem/POIFSMiniStore.java
@@ -252,7 +252,7 @@ public class POIFSMiniStore extends BlockStore
           if (!sbat.hasFreeSectors()) {
               blocksUsed += _filesystem.getBigBlockSizeDetails().getBATEntriesPerBlock();
           } else {
-              blocksUsed += sbat.getUsedSectors(false);
+              blocksUsed += sbat.getOccupiedSize();
           }
        }
        // Set the size on the root in terms of the number of SBAT blocks
--- a/src/java/org/apache/poi/poifs/storage/BATBlock.java
+++ b/src/java/org/apache/poi/poifs/storage/BATBlock.java
@@ -192,6 +192,25 @@ public final class BATBlock implements BlockWritable {
         return usedSectors;
     }
 
+    /**
+     * How much of this block is occupied?.
+     * This counts the number of sectors up and including the last used sector.
+     * Note that this is different from {@link #getUsedSectors(boolean)} which
+     * could be smaller as it does not count unused sectors where there are
+     * used ones after it (i.e. fragmentation).
+     */
+    public int getOccupiedSize() {
+        int usedSectors = _values.length;
+        for (int k = _values.length - 1; k >= 0; k--) {
+            if(_values[k] == POIFSConstants.UNUSED_BLOCK) {
+                usedSectors--;
+            } else {
+                break;
+            }
+        }
+        return usedSectors;
+    }
+
     public int getValueAt(int relativeOffset) {
        if(relativeOffset >= _values.length) {
           throw new ArrayIndexOutOfBoundsException(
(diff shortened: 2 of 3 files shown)


netmackan:
I also have patches for 4.1.2, 4.0.1 and 3.15 if wanted.

pjfanning:
@netmackan do you have a file that demos this problem?

pjfanning:
@netmackan is there any extra context you can give us on this issue? The code has probably never supported the scenario you seem to want to fix.

netmackan:
Hi @pjfanning, 
Unfortunately, I do not own the test file so I can not share it. But I will try to create a new one and get back as soon as I get time. 

The file (MSI/Windows Installer) was created with MS Visual Studio 2017 15.9 and it happened that only some of the created files would have a fragmented block and trigger the issue.

Regarding the scenario, my thought on why this has not been seen before is probably because in most scenarios it is not important that the root property has the right size. For the MSI use case probably Windows is more strict in verifying things as part of the digital signature verification. On top of that this only happens when the file added to the POIFS is larger then 4096 bytes (i.e. not a small block) and only when the original file happens to have a block that is having gaps in it so that the size calculation will be different than the current implementation.

ebourg:
I'm late to the party, I stumbled on this PR as I was debugging a MSI file with an invalid signature generated by Jsign (using POI 5.3.0). This file has 6 mini FAT sectors, the first 4 sectors are full with all mini sectors allocated, and the last two sectors have the following layout:

```
  Sector #88115 : Mini FAT #5
    Mini FAT entries:
      [...]
      120 : 633
      121 : 634
      122 : 635
      123 : 636
      124 : -2 (End of chain)
      125 : -2 (End of chain)
      126 : -1 (Free sector)
      127 : -1 (Free sector)
  
  Sector #173743 : Mini FAT #6
    Mini FAT entries:
      0 : -1 (Free sector)
      1 : -1 (Free sector)
      2 : -1 (Free sector)
      3 : -2 (End of chain)
      4-127 : -1 (Free sector)
```

For this file POI computes 4 x 128 + 126 + 4 = 642 mini sectors, for a size of 41088 bytes. However signtool expects 644 mini sectors (41216 bytes) in order to validate the signature.

It looks like the occupied size has only to be computed for the last mini FAT sector, and the other sectors are counted as fully allocated. In this case it gives 5 x 128 + 4 = 644.

I don't know how the size is computed if the last mini FAT sector consists only of unallocated mini sectors, for example:

```
  Sector #88115 : Mini FAT #5
    Mini FAT entries:
      [...]
      120 : 633
      121 : 634
      122 : 635
      123 : 636
      124 : -2 (End of chain)
      125 : -2 (End of chain)
      126 : -1 (Free sector)
      127 : -1 (Free sector)
  
  Sector #173743 : Mini FAT #6
    Mini FAT entries:
      0-127 : -1 (Free sector)
```

In this case signtool could either:
* trim the last sector, and there would be  4 x 128 + 126 = 638 mini sectors
* keep the last sector but ignoring it, so 638 mini sectors
* keep the last sector and count as guessed above, so 5 x 128 + 0 = 640 mini sectors

ebourg:
I created a test file with 8 mini FAT sectors:
* Sector 1 is unallocated
* Sector 2 has 8 unallocated mini sectors at the beginning
* Sector 3 has 4 unallocated mini sectors in the middle
* Sector 4 is unallocated
* Sector 5 has 32 unallocated mini sectors at the end
* Sector 6 has 64 unallocated sectors at the beginning and 16 unallocated mini sectors at the end
* Sector 7 is unallocated
* Sector 8 is unallocated

With this layout POI 5.3 computes a root size of 29696 bytes, that is (0 + 128 + 128 + 0 + (128-32) + (128-16) + 0 + 0) x 64.

After signing the file with signtool:
* The first 2 mini sectors of the first mini FAT sector have been allocated (for the MsiDigitalSignatureEx entry)
* The empty mini FAT sectors 4, 7 and 8 were not removed
* The size field of the root entry is 48128 bytes, that's (5 x 128 + (128 - 16)) x 64

So the right method to compute the size is to find the absolute index of the last allocated mini sector, add one and multiply by 64.

PR #183, opened 2020-05-27
fix gradle build after #64411 changes

kortov:
After rearranging some lib folders of ooxml in https://github.com/apache/poi/commit/114a4dbd41c56e4c544836e58fae02aae6793cdf , gradle build (and therefore IDE imports) was broken. This fixes the issue.


pjfanning:
LGTM. Merged.

PR #185, opened 2020-07-02
Fix if function in array formulas with 2nd argument evaluating to error

RemboL:
I reported this before as https://bz.apache.org/bugzilla/show_bug.cgi?id=64459 , basically issue is that formula "IF(FALSE(), #VALUE!, 1)" when being used as array formula evaluates to error value passed as 2nd argument despite the fact that condition is false.


pjfanning:
thanks - but can you add a test case that uses IfFunctionTestCaseData.xls ?

RemboL:
https://github.com/apache/poi/blob/trunk/src/testcases/org/apache/poi/ss/formula/functions/TestIFFunctionFromSpreadsheet.java already does that, I believe

pjfanning:
So you are replacing a pre-existing xls file and believe that the pre-existing test is enough and that no extra assertions should be added to this test?

RemboL:
Yes.

This test is built so that it scans the excel file and adds new data row to `@Parameterized` test defined in the base class per properly defined row in excel. I've added 4 new rows to pre-existing xls (while leaving all the previous rows intact) that asserts behavior I'm interested in - so in fact I've added 4 new assertions to pre-existing `@Parameterized` test. 

These assertions failed (specifically, the one that was testing this particular situation) until I've modified the if function itself - and after modifying the function, all the other tests were still green - so personally I'm satisfied with how this modification is tested.

pjfanning:
merged with https://github.com/apache/poi/commit/be096396384f93c4d9027484a997509edf240708 - thanks

PR #187, opened 2020-08-02
Add length validation for Excel DataValidations that are list literals.

leowebb:
Hello,

I am not entirely sure if this is a bug, but as I was writing some code to produce a list literal for data validations, I was running into an issue where when I was producing an xlsx, I was unable to open the file in Microsoft Excel due to what I eventually understood to be a hard limit on the size of the data validation literal, which was 255 characters (including separators, excluding quotes). I spoke with a maintainer on IRC about whether this was an issue or not, and he advised that I double check the specifications for OOXML documents. I wasn't able to find anything about limiting the size of `formula1` literals, but sure enough if one manually edits the file XML and then opens it in Excel, the user will be presented with an error saying that the file cannot be read by Excel due to some error.

The reason I'm not sure for whether or not is is a bug is due to the specification seemingly being right (these files open just fine in OpenOffice/LibreOffice), but Excel seems to ignore this for whatever reason.

For wholeness, I also checked in the HSSF code, and this limit is enforced there (specifically `StringPtg`s have a hard limit of 255 ASCII characters, and that is what is used when creating the Data Validations).

Otherwise, assuming this is a bug because Excel cannot open the file but the software still generates the file, then here is a PR to hopefully fix the issue.

Thanks.


centic9:
Thanks for the contribution.

Any chance you can add a sample file to the test-data/spreadsheet directory and a unit-test which verifies the added length-validation?

leowebb:
Hello! Thank you for the response. 

I've added a sample file that won't open in Excel, and a corresponding test that throws an exception from the DataValidation when reading from it. Hopefully this is what you were looking for - still new to the entire code base. If the test should look different, please help me find which files I should be testing against.

Thanks!

pjfanning:
thanks - merged with https://github.com/apache/poi/commit/a8d90aba91ba8788c784e644a6cff2acbcaba4f1

PR #188, opened 2020-08-07
CTChartSpace may have set CTExternalData element

Sayi:
Use the existing CTExternalData element, so that we can keep the settings such as autoUpdate in CTExternalData.


pjfanning:
Looks good to me. I can merge (using svn) if you haven't got svn write access yet.

Sayi:
r1880707

PR #189, opened 2020-08-11
Move date parsing logic to DateParser

RemboL:
Recently I implemented DateValue function that would parse dates in different formats.

Since then, I noticed that there are other functions (YEARFRAC, in particular) that use date parsing logic and that they are using a class DateParser. This brought my attention, because DateParser is unable to parse dates in other formats, like 'yyyy-MM-dd'.

I think it would make sense to use the same logic for date parsing in all places, so I moved my code from DateValue function to DateParser. Since both DateParser and DateValue tests are still green without any changes, I think this refactor is valid.

I also moved DateParser from atp package to util, since it is used outside atp now - although I see that it might raise some flags, since it is a public class and theoretically it might break compatibility for other projects that might be using DateParser - please advice, whether it is acceptable to move this class or should it be left in current package.


pjfanning:
Thanks - merged with https://github.com/apache/poi/commit/e86ba86f2dbc6c48b34a24bed3cbb2ba2e087648

PR #193, opened 2020-10-20
Change TRUNC implementation to use MathX

jacobharris919:
The TRUNC implementation is still broken.  (BUG:  https://bz.apache.org/bugzilla/show_bug.cgi?id=62506 )
1. Bug 62506 reports that TRUNC(0.29, 2) = 0.28
2. The solution proposed in that bugfix still fails for some calculations (see test case)

The main argument against the proposed fix in 62506 was the performance hit of using BigDecimal for every round() operation.  Since the round() methods now use MathX methods which construct a BigDecimal from the Excel string, the TRUNC function can be fixed by using the MathX.roundDown() method.


pjfanning:
Merged with https://github.com/apache/poi/commit/dd9f132c232c24d939adba5065280fbba0e4ed42 - thanks

PR #196, opened 2020-11-13
Delete unused certificate exceptions

MariusVolkhart:
These exceptions are unused.

This is an API breaking change. It was not clear to me if that is desired or not for v5.0.


pjfanning:
Thanks - merged with https://github.com/apache/poi/commit/bebbf025b6ba809f9adeb9104062804998d3ed99

PR #197, opened 2020-11-13
Modify test utility class to compile on OpenJ9

MariusVolkhart:
The HeapDump class was referencing a function that exists only on VMs that use HotSpot. Other VMs, such as OpenJ9 don't have this function, so the tests did not compile.

### DIFF
--- a/src/integrationtest/org/apache/poi/stress/HeapDump.java
+++ b/src/integrationtest/org/apache/poi/stress/HeapDump.java
@@ -18,6 +18,8 @@ package org.apache.poi.stress;
 
 import java.io.IOException;
 import java.lang.management.ManagementFactory;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 
 import com.sun.management.HotSpotDiagnosticMXBean;
 import org.apache.poi.ss.formula.eval.NotImplementedException;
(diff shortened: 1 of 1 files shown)


src/integrationtest/org/apache/poi/stress/HeapDump.java
```diff
@@ -42,7 +43,17 @@ Licensed to the Apache Software Foundation (ASF) under one or more
     public static void dumpHeap(String fileName, boolean live) throws IOException {
         // initialize hotspot diagnostic MBean
         initHotspotMBean();
-        hotspotMBean.dumpHeap(fileName, live);
+        throw new NotImplementedException("Developer: You need to uncomment the corresponding line here.");
```
pjfanning: I'd prefer if this was modified to use Java reflection

MariusVolkhart: Done! @pjfanning mind taking another look? I tested w/ OpenJ9 11 and OpenJDK 16

pjfanning:
Thanks. Merged with https://github.com/apache/poi/commit/00b9048d59e8800a59560425d11bbe13d44a60a6 - will check later to see if there any problems with the Jenkins builds.

PR #198, opened 2020-11-21
Remove jdk.charset module dependency for spreadsheets generation

robmv:
Move double byte charset utility code to scratchpad...hwpf that is the only code that currently requires them. This allows using spreadsheet related code without pulling jdk.charset to custom built JREs using jlink.


robmv:
The build failure looks unrelated to my changes

```
/home/travis/build/apache/poi/build.xml:938: javac doesn't support the "release" attribute
```

kiwiwings:
> The build failure looks unrelated to my changes
> 
> ```
> /home/travis/build/apache/poi/build.xml:938: javac doesn't support the "release" attribute
> ```

If you get that warning, your javac is not running in Java 9+ mode. I thought I've changed the build in a way, that it recognizes the JVM. So it uses classpath mode for Java 8 and modulepath mode for Java 9+. if this doesn't work for your environment, I need to investigate what's the difference there ... with your help.

anyway with this warning and your proposed change, I think the tests are running in classpath mode and therefore aren't significant ...

robmv:
> If you get that warning, your javac is not running in Java 9+ mode. 

My build using Java 8 says: `Support for javac --release has been added in Java9 ignoring it`.

The build machines say the Ant version is:  `Using Ant: Apache Ant(TM) version 1.9.6 compiled on July 20 2018`

 This is a [message from current Ant](https://github.com/apache/ant/blob/a8645a151bc706259fb1789ef587d05482d98612/src/main/org/apache/tools/ant/taskdefs/compilers/DefaultCompilerAdapter.java#L402) that is [not present in that old Ant](https://github.com/apache/ant/blob/rel/1.9.6/src/main/org/apache/tools/ant/taskdefs/compilers/DefaultCompilerAdapter.java)

An Ant update on the build machines should fix it

Edit: My Ant is `Apache Ant(TM) version 1.10.5 compiled on July 10 2018`

robmv:
> If you get that warning

Just for clarifying, I don't get that error, in my machine the build is ok. The Travis CI display that error, see my previous message about the Travis CI using and older version of Ant than me

pjfanning:
@robmv could you merge the poi trunk to your MR? I think a recent change could help with the travis build.

robmv:
@pjfanning Done. Please note that I had to remove `src/multimodule/scratchpad/test9/module-info.class` that is a committed file in order to allow the build script to take the changes, and for the same reason it had to be added to the commit so the CI could take it.

I wonder if the reason compiled `module-info.class`are stored on git is to support building with Java 8 and still support modules, or a mistake

PR #202, opened 2020-12-06
Remove PackagePart#flush() API

MariusVolkhart:
This method is never called by the POI code, and none of the built-in implementations have meaningful implementation of the method.

PackagePropertiesPart#flush() was called in one location, but as the implementation is empty, the call is unnecessary.

Fixes Bugzilla 64959


pjfanning:
We have a policy of deprecating code before removing it. I don't think this change is a good idea, even if the code is not really used.

PR #203, opened 2020-12-06
Prototype nullability annotations

MariusVolkhart:


src/java/org/apache/poi/common/usermodel/fonts/FontGroup.java
```diff
@@ -42,10 +44,14 @@ Licensed to the Apache Software Foundation (ASF) under one or more
     COMPLEX_SCRIPT
     ;
 
-
     public static class FontGroupRange {
         private int len;
-        private FontGroup fontGroup;
+        private final FontGroup fontGroup;
+
+        public FontGroupRange(FontGroup fontGroup) {
```
MariusVolkhart, line 51: Adding a constructor here avoids needing to add `@Nullable` to the `fontGroup` field. I imagine this pattern would arise several times.

src/java/org/apache/poi/common/usermodel/fonts/FontGroup.java
```diff
@@ -110,7 +116,7 @@ public FontGroup getFontGroup( ) {
      * @param runText the text which font groups are to be analyzed
      * @return the FontGroup
      */
-    public static List<FontGroupRange> getFontGroupRanges(final String runText) {
+    public static List<FontGroupRange> getFontGroupRanges(@Nullable final String runText) {
```
MariusVolkhart, line 119: I think the intent is for people to provide a non-null value. However, the function can handle `null` since it has an explicit null check. Leaving the `@Nullable` off and having the null check triggers an Intellij inspection. I figure adding `@Nullable` is probably an OK choice to 1. resolve the inspection 2. allow tools using the nullness annotations to not complain about passing something that _might_ be null.

I think a lot of parameters will end up being marked as `@Nullable`

src/java/org/apache/poi/common/usermodel/fonts/package-info.java
```diff
@@ -0,0 +1,4 @@
+@NonNullApi
```
MariusVolkhart, line 1: Marks every return value and parameter as being or needing to be non-null unless annotated otherwise. I don't know if this is desired or feasible, but it's what most projects choose as the base. "non-null by default, explicitly nullable via `@Nullable`.

src/java/org/apache/poi/common/usermodel/fonts/FontGroup.java
pjfanning: I think it is safe to change this class now - because the 2 members are private so setting them is not really feasible from 3rd party code.

pjfanning: https://github.com/apache/poi/commit/10652e33feb2ee164437ae265f98d8443a94aa19 -- I've removed the stray import in a 2nd commit

PR #205, opened 2020-12-08
Change XPWFPictureData#getChecksum() to return long instead of Long

MariusVolkhart:


src/ooxml/java/org/apache/poi/xwpf/usermodel/XWPFPictureData.java
```diff
@@ -146,7 +146,7 @@ public int getPictureType() {
         return 0;
     }
```
pjfanning, line 148: with public APIs, we need to go through a deprecation cycle

MariusVolkhart: Is this a change that is desired to go through a deprecation cycle?

PR #206, opened 2020-12-10
Improve perfomance of cell evaluation

ledstellar:
The `SXSSFRow.getRowNum()` method is heavily used from various parts of the code (in particular, from the cell evaluation method). Current implementation of the method is using iterating over sheet rows to find the row number. This approach is ineffective for large sheets. The proposed patch stores the row number directly in the row object to improve method performance.

### DIFF
--- a/src/ooxml/java/org/apache/poi/xssf/streaming/SXSSFRow.java
+++ b/src/ooxml/java/org/apache/poi/xssf/streaming/SXSSFRow.java
@@ -39,7 +39,6 @@ import org.apache.poi.util.NotImplemented;
 public class SXSSFRow implements Row, Comparable<SXSSFRow>
 {
     private static final Boolean UNDEFINED = null;
-
     private final SXSSFSheet _sheet; // parent sheet
     private final SortedMap<Integer, SXSSFCell> _cells = new TreeMap<>();
     private short _style = -1; // index of cell style in style table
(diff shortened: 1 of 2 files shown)


src/ooxml/java/org/apache/poi/xssf/streaming/SXSSFRow.java
```diff
@@ -49,6 +48,7 @@ Licensed to the Apache Software Foundation (ASF) under one or more
     // use Boolean to have a tri-state for on/off/undefined
     private Boolean _hidden = UNDEFINED;
     private Boolean _collapsed = UNDEFINED;
+    private int number;
```
pjfanning: could you call this rowNumber?

src/ooxml/java/org/apache/poi/xssf/streaming/SXSSFRow.java
```diff
@@ -437,6 +437,10 @@ public SXSSFSheet getSheet()
     {
         return _sheet;
     }
+
+	public void setNumber(int rownum) {
```
pjfanning: does this need to be public - can't it be package-private? -- can it be called setRowNum? To match the name of getRowNum.

ledstellar:
Sure. Updated the patch to meet existing code style.

pjfanning:
Merged. Thanks.

PR #209, opened 2020-12-19
Improve performance of Bitmap#setData

MariusVolkhart:


centic9:
Just wondering as we did various performance benchmarks over time, do you actually see these things causing performance issues? If yes, can you share some details about it?

MariusVolkhart:
@centic9 No, we have not seen these as hotspots. Just things I come across as I'm becoming more familiar with the internals. Closing this as it was part of https://github.com/apache/poi/commit/d1c9a07860e365db4e335d24ad67e87544bdcceb

centic9:
It's fine,  "new ByteArrayOutputStream(int size)" would allow to populate the stream without additional allocation, but toByteArray() performs a full array-copy again, so doing it ourselves should avoid this additional copying. 

In one of my projects I use the following helper class which avoids all the manual array-positioning and copying yourself and keeps using the ByteArrayOutputStream.

```
class NonCopyingByteArrayOutputStream extends ByteArrayOutputStream {
	public NonCopyingByteArrayOutputStream(int size) {
		super(size);
	}

	public synchronized byte[] toByteArrayDirect() {
		return buf;
	}
}
```

In general I find it useful to verify such changes with results "before" and "after" via some tests as sometimes the JDK applies complicated optimizations which can cause changes to actually cause a slowdown.

PR #211, opened 2020-12-19
#65009 Fix HSLFPictureData indices during presentation load

MariusVolkhart:
Index HSLFPictureData starting at 1.


centic9:
Applied via r1884948 in a slightly different way to use the same approach across two similar code-places.

PR #214, opened 2021-01-04
poi-ooxml: add new API to add text transparency on XLSX

artragis:
This API uses XSSFColor built with alpha channel to set up the right value.
We must take care of the fact Excel uses 1000th of percent and positive numbers

### DIFF
--- a/src/ooxml/java/org/apache/poi/xssf/usermodel/XSSFTextParagraph.java
+++ b/src/ooxml/java/org/apache/poi/xssf/usermodel/XSSFTextParagraph.java
@@ -28,6 +28,7 @@ import org.apache.poi.util.Units;
 import org.apache.poi.xssf.model.ParagraphPropertyFetcher;
 import org.apache.xmlbeans.XmlObject;
 import org.openxmlformats.schemas.drawingml.x2006.main.*;
+import org.openxmlformats.schemas.drawingml.x2006.main.impl.STPositiveFixedPercentageImpl;
 import org.openxmlformats.schemas.drawingml.x2006.spreadsheetDrawing.CTShape;
 
 /**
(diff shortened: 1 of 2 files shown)


src/ooxml/java/org/apache/poi/xssf/usermodel/XSSFTextParagraph.java
```diff
@@ -863,6 +864,41 @@ private boolean fetchParagraphProperty(ParagraphPropertyFetcher visitor){
         return ok;
     }
 
+    public void setTextFillColor(XSSFColor color) {
```
pjfanning, line 867: could you add a getTextFillColor?

artragis: Must I return null if no fill colour is found?

pjfanning: Probably best to return null if no value is set

artragis: Ok, I will try to do so. But translating CTColors into XSSFColor may be hard, Is there any helper to do that in the project?

pjfanning: XSSFColor has `public static XSSFColor from(CTColor color, IndexedColorMap map)` - you can use `new DefaultIndexedColorMap()`

artragis: Ok I will try to build `org.openxmlformats.schemas.spreadsheetml.x2006.main.CTColor` from CTSolidColorFillProperties.

I'm wondering if we cannot create a sub interface for all those colors instead of CTColor

PR #220, opened 2021-01-19
Bug 65016 - Creating a chart throws exception

Alain-Bearez:


centic9 (approved):
LGTM, seems to fix the issue for me

Alain-Bearez:
applied with r1885700

PR #221, opened 2021-01-19
Revert addRow to behaviour before 4.1.2

Alain-Bearez:


kiwiwings:
Thank you for providing a fix so fast - applied via release 1885699 with an additional test

PR #222, opened 2021-01-24
Replace deprecated methods used  in tests

Alain-Bearez:
Also small changes in examples code to avoid bug reports.


Alain-Bearez:
applied with r1885886

PR #223, opened 2021-02-01
Bug 63901: Clone content of imported chart

Alain-Bearez:


Alain-Bearez:
Applied with r1886087

PR #224, opened 2021-02-01
Use Log4j 2 for logging

MariusVolkhart:
Hi devs! Here is iteration 1 of switching the logging to be entirely based on Log4j 2 and removing `POILogger` & co. It would be helpful if you could try this out in your workflows (running/analyzing integration tests, jenkins/sonar intergrations, OSGi, etc) and see if there are things you're now missing. The code "works" but I might have configured something incorrectly.

I want to give a special thanks to @pjfanning whose previous work for adding a SLF4J integration for `POILogger` made it a lot easier to figure out where in all the ant/gradle/maven build files dependencies needed to be changed.

As discussed on the mailing list, this change it targeted for 5.1.0. If/When there are no problems, I'll take this over to SVN.

_Commit message below_
This removes the POILogger and POILogFactory mechanism for logging. This mechanism was created at a time when the Java landscape looked very different than it does today.

Log4j 2 is an Apache Foundation project that is well maintained and is widely regarded as having good performance and capabilities. We use only the Log4j API artifact. This lets application developers choose how they want to capture logging events if at all. Integrations with Log4j 2 Core and Logback are available from the Log4j project.


build.xml
```diff
@@ -289,7 +288,7 @@ under the License.
     <dependency prefix="main.byte-buddy" artifact="net.bytebuddy:byte-buddy:1.10.19" usage="main-tests"/>
     <dependency prefix="main.byte-buddy-agent" artifact="net.bytebuddy:byte-buddy-agent:1.10.19" usage="main-tests"/>
     <dependency prefix="main.objenesis" artifact="org.objenesis:objenesis:3.1" usage="main-tests"/>
-    <dependency prefix="main.slf4j-simple" artifact="org.slf4j:slf4j-simple:1.7.30" usage="main-tests"/>
+    <dependency prefix="main.log4j-core" artifact="org.apache.logging.log4j:log4j-core:2.14.0" usage="main-tests"/>
```
MariusVolkhart, line 289: We could keep using `slf4j-simple`, but would need to add an additional "bridge" dependency

legal/LICENSE
```diff
@@ -285,29 +285,9 @@ CurvesAPI / Curve API
...
-    included in all copies or substantial portions of the Software.
-
-    THE  SOFTWARE IS  PROVIDED  "AS  IS", WITHOUT  WARRANTY  OF ANY  KIND,
-    EXPRESS OR  IMPLIED, INCLUDING  BUT NOT LIMITED  TO THE  WARRANTIES OF
-    MERCHANTABILITY,    FITNESS    FOR    A   PARTICULAR    PURPOSE    AND
-    NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
-    LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
-    OF CONTRACT, TORT OR OTHERWISE,  ARISING FROM, OUT OF OR IN CONNECTION
-    WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+    Apache License Version 2.0
```
MariusVolkhart, line 290: Not sure if this is necessary, or the right way to document this?

log4j2-console.xml
```diff
@@ -0,0 +1,42 @@
...
+    <Logger name="org.apache.poi.hpsf.CodePageString" level="ERROR" />
+    <Logger name="org.apache.poi.hdgf.chunks.ChunkFactory" level="ERROR" />
+    <Logger name="org.apache.poi.hslf.model.textproperties.BitMaskTextProp" level="ERROR" />
+    <Logger name="org.apache.poi.hslf.usermodel.HSLFTextParagraph" level="ERROR" />
+    <Logger name="org.apache.poi.openxml4j.opc.ZipPackage" level="ERROR" />
+    <Logger name="org.apache.poi.POIDocument" level="WARN" />
+    <Logger name="org.apache.poi.openxml4j.opc.OPCPackage" level="ERROR" />
+    <Logger name="org.apache.poi.xssf.usermodel.XSSFWorkbook" level="ERROR" />
+    <Logger name="org.apache.poi.hslf.usermodel.HSLFGroupShape" level="WARN" />
+    <Logger name="org.apache.poi.hslf.record.Record" level="ERROR" />
```
MariusVolkhart: These logger/level pairs are an aggregation of various logging files I saw

osgi/pom.xml
```diff
@@ -104,7 +103,8 @@
                             commons-math3,
                             commons-compress,
                             commons-collections4,
-                            commons-codec
+                            commons-codec,
+                            log4j-api
```
MariusVolkhart, line 107: I don't know how this works or how to verify it. I'd appreciate any help or guidance.

build.xml
pjfanning: best to drop slf4j if log4j2 is what we really want to support

osgi/pom.xml
pjfanning: remove any unnecessary slf4j or commons-logging entries

build.gradle
```diff
@@ -437,6 +439,7 @@ project('scratchpad') {
         compile project(':main')
         compile "commons-codec:commons-codec:${commonsCodecVersion}"
         compile "org.apache.commons:commons-math3:${commonsMathVersion}"
+        implementation "org.apache.logging.log4j:log4j-api:${log4jVersion}"
```
pjfanning: can you stick with 'compile' and we can change all compile to implementation later?

log4j2-console.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
centic9: Is there a way to move the two xml-configuration files away from the root-directory?

For other logging frameworks these are loaded via the classpath, so would go to the "src/resources". 

Not sure how this would work in the actual deployment of Apache POI as part of a bigger application, would they also have to store those files in some additional directory and need to set the system-property?

MariusVolkhart: > Is there a way to move the two xml-configuration files away from the root-directory?
>
> For other logging frameworks these are loaded via the classpath, so would go to the "src/resources".

Indeed. Our tests use Log4j 2 Core as the logging backend, which also loads configuration from the classpath. So in the case of the POI tests, which use the `log4j2-no-logging.xml` configuration by default, I could place a copy of that config into each test sourceset's `src/resources` directory, and wouldn't need to specify the `log4j.configurationFile` system property.

To then get the console behavior, one option is to continue using the `log4j.configurationFile` system property as this code currently does. Another option is to combine the configurations and the developer running the tests comments/uncomments lines of code in the configuration as they need.

I don't have a strong opinion on one method or another.

> Not sure how this would work in the actual deployment of Apache POI as part of a bigger application, would they also have to store those files in some additional directory and need to set the system-property?

I'll answer assuming the application developer is also using Log4j 2 Core as the backend (same as the POI tests in this PR), but I assume similar rules apply if they are using Logback or another logging backend.

The existing logging configuration that the developer has will decide how POI log messages are captured. That configuration likely lives at `main/src/resources` and is loaded from the classpath, but it is possible the developer has previously chosen to use the system property, is doing programmatic configuration, etc.

Here are some examples of what a developer might indicate in their configuration:
1. Capture all POI logs at the same level as their application logging. Basically, they just say "I want all messages `INFO` and higher, regardless where they come from", and POI messages are included in that.
2. Capture only messages they opt in to. They might say "Only collect messages from package `com.myapplication.*` that are INFO and higher". POI messages are excluded.
3. Capture all messages except POI. They might say "Collect everything _except_ `org.apache.poi.*`."

I hope that makes sense. If not, let me know! I haven't started on the changes to the logging documentation for the site, but if seeing those changes as part of the PR would be helpful, I can work on those in the next few days while this sits :)

MariusVolkhart: @centic9 I took a stab at moving the logging configs to be more how a project typically would use them. See https://github.com/MariusVolkhart/poi/pull/1 for the changes. If you'd like those changes pulled in, let me know.

centic9: Thanks for the effort, to me it looks clearer this way as it makes it easy to see that these files are only used for tests and not deployed as part of the released library itself.

kiwiwings:
When is this applied to SVN?
I'd like to change to Log4j 2 also on XmlBeans too, but I think this makes only sense, when POI is already migrated.

MariusVolkhart:
I was waiting for 5.0.1 to be released, but I can commit this to SVN this
weekend.

kiwiwings:
SLF4J is missing in XmlBeans module-info, but it's used. So we might want to release a new version there anyway, i.e. before POI 5.0.1.
And if the logging matches that would make things smoother I guess ...

pjfanning:
@kiwiwings the logging in xmlbeans is minimal and the slf4j was only added recently by me - since the last release so it could be replaced by log4j2x without any major impact - if we make these logging changes, it does seem like the next releases should be XmlBeans 5.0.0 and POI 6.0.0

PR #225, opened 2021-02-06
Add the ability to edit HSLFPictureData contents

MariusVolkhart:
Hey Devs, I'd appreciate a review of the following code for any obvious flaws you see with this approach of enabling picture editing in `HSLFPictureData`. In my initial testing I haven't seen any issues, but more eyes are always good.

Pictures can now be edited by calling HSLFPictureData#setData(byte[]). The byte[] should contain the image data as an image viewer might read it.

To enable this functionality, a tighter coupling between the EscherBSERecords of the slideshow and the HSLFPictureData was required. This ensures that changes in image data size are accurately recorded in the records.

In the course of coupling the records and the HSLFPictureData, various scenarios arose where a mapping of records to pictures was non-trivial. Accordingly, the HSLFSlideShowImpl#matchPicturesAndRecords(...) function was added to perform a more sophisticated matching pass. This function is heavily exercised by org.apache.poi.hslf.usermodel.TestBugs.testFile[5] and PPTX2PNG.render[2], as well as the new TestPictures#testSlideshowWithIncorrectOffsets().


kiwiwings:
Hi Marius,
I've just quickly browsed the changes.
Maybe we should add a small testcase with encryption enabled, i.e. create encrypted hslf with an image > save/reopen > add / edit picture > reopen > compare checksum. I think when reading the pictures, the stream is cached, but a test wouldn't hurt.

MariusVolkhart:
@kiwiwings I added a test. It uses a premade encrypted slideshow, but otherwise does what you described.

PR #226, opened 2021-02-08
[Bug-57835] Import slide notes when importing slide content

Alain-Bearez:
This was a low hanging fruit which I neglected for too long.


Alain-Bearez:
Applied with r1886338

PR #227, opened 2021-02-09
[bug-63902] reference cloned sheet in cloned chart data series

Alain-Bearez:


Alain-Bearez:
Applied with r1886605

PR #228, opened 2021-02-17
Manipulate individual data point properties

Alain-Bearez:
This should implement the feature requested some months ago on
https://pt.stackoverflow.com/questions/472162/mudar-a-cor-de-apenas-uma-das-barras-colunas-do-gr%c3%a1fico-gerado-no-pptx-apache


MariusVolkhart (approved):

Alain-Bearez:
Applied with r1886642

PR #229, opened 2021-03-09
Allow negative numbers in conditional part of format string

aschott-looker:
* Negative numbers do not work as expected if they are used in the conditional part of a format string. For example, using `[<-10]#" Wow"` will blow up (poi thinks it is an elapsed time format).
* The fix was pretty straightforward but Excel itself does not seem to support this case very well either. If you do `=TEXT(-20, [<-10]#" Wow")`, you'd expect an output of `-20 Wow` but you actually get `20 Wow`. Oddly enough, `>` works just fine. For example, `=TEXT(-5, [>-10]#" Wow")` correctly outputs `-5 Wow`.


centic9:
The included test-file seems to be an empty file, is this intended?

aschott-looker:
It's an XLSX file. GitHub can't render the diff.

aschott-looker:
@centic9 -- Any thoughts on this change/estimation on when this would be available in a new POI release? We are trying to transition from https://github.com/SheetJS/ssf to this library and would be great to not have to fork this repo.

centic9:
Applied to trunk via https://svn.apache.org/viewvc?view=revision&revision=1887802 but I cannot state when exactly this will be released, as the developer community did not decide fully on the next release yet.

PR #234, opened 2021-03-31
Ensure the hours position is correct after string replacements and perform y to yy conversion in the part handler

aschott-looker:
* This PR addresses two bugs described below. Consider the following format: `"d \d\a\y\s h"`.
* For one, The logic in `org.apache.poi.ss.format.CellDateFormatter.DatePartHandler#finish` uses the position of the "h" to upcase unless AM/PM formatting was requested in the string. The example format will blow up with an index out of bounds exception because the `\d\a\y\s` section is replaced with `'days'` which changes the length of the string. My fix was to update the position whenever we update the length of the string so that we don't go out of bounds of the string.
* Secondly, there was a previous fix that regex'd the "y" out of the string and replaced it with "yy". This does not work since it does not account for "y"'s in an escaped text block like in my example so you would get an output of `dayys`. It is better to do this type of thing in the part handler while we are parsing/tokenizing the string.
* For testing, I added the following test cases:
```
d \d\a\y\s h
d "days" h
d \d\a\y\s h a/p
d "days" h am/pm
```


aschott-looker:
@centic9 -- Any thoughts on this?

pjfanning:
could you add a unit test that uses DateFormatTests.xlsx ?

aschott-looker:
@pjfanning -- https://github.com/apache/poi/blob/382714eccd92667fc83f70115b736c64ebff9700/poi-ooxml/src/test/java/org/apache/poi/ss/tests/format/TestCellFormatPart.java#L132

poi/src/main/java/org/apache/poi/ss/format/CellDateFormatter.java
```diff
@@ -96,6 +96,10 @@ public String handlePart(Matcher m, String part, CellFormatType type,
                 mStart = -1;
                 if (part.length() == 3)
                     part = "yyyy";
+                // tweak the format pattern to pass tests on JDK 1.7,
```
pjfanning, line 99: we don't support jdk1.7

pjfanning:
Thanks - merged with https://github.com/apache/poi/commit/e7a2df7dda2458b78a4d5495703289d3ec31c69f

PR #235, opened 2021-04-23
Fix date formatting for number cell values

aschott-looker:
* Numbers are not supported correctly in POI nor are they tested. Excel interprets numbers as fractional days since `1/1/1900` or `1/4/1900` depending on your settings. The fix here was to convert fractional days to milliseconds and round up in a way that is consistent with Excel.
* I added a new spreadsheet with tests that verify rounding and no rounding cases.


pjfanning:
thanks - merged using https://github.com/apache/poi/commit/50cfe3c10ed4dbee2d9b066d78e5fedf01921e4e

PR #238, opened 2021-07-09
Implement specificWorkbookSheet method

HasaanA16:
The specificWorkbookSheet() takes a workbook and sheetNumber as an argument, and creates an html of a specific sheet using the predefined method processSheet().

Initially it was not possible as processSheet() is protected and cannot be used.

### DIFF
--- a/poi-scratchpad/src/main/java/org/apache/poi/hssf/converter/ExcelToHtmlConverter.java
+++ b/poi-scratchpad/src/main/java/org/apache/poi/hssf/converter/ExcelToHtmlConverter.java
@@ -136,7 +136,7 @@ public class ExcelToHtmlConverter extends AbstractExcelConverter {
     public static Document process(HSSFWorkbook workbook) throws IOException, ParserConfigurationException {
         ExcelToHtmlConverter excelToHtmlConverter = new ExcelToHtmlConverter(
             XMLHelper.newDocumentBuilder().newDocument());
-        excelToHtmlConverter.processWorkbook(workbook);
+        excelToHtmlConverter.processWorkbookSheet(workbook);
         return excelToHtmlConverter.getDocument();
     }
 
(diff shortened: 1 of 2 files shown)


poi-scratchpad/src/main/java/org/apache/poi/hssf/converter/ExcelToHtmlConverter.java
```diff
@@ -668,7 +668,7 @@ protected void processSheetHeader(Element htmlBody, HSSFSheet sheet) {
         htmlBody.appendChild(h2);
     }
 
-    public void processWorkbook(HSSFWorkbook workbook) {
+    public void processWorkbookSheet(HSSFWorkbook workbook) {
```
pjfanning, line 671: This doesn't need to be renamed

poi-scratchpad/src/main/java/org/apache/poi/hssf/converter/ExcelToHtmlConverter.java
```diff
@@ -692,6 +692,24 @@ public void processWorkbook(HSSFWorkbook workbook) {
         htmlDocumentFacade.updateStylesheet();
     }
 
+    public void processWorkbookSheet(HSSFWorkbook workbook, int sheetNumber) {
```
pjfanning, line 695: Can you call this processWorksheet?

pjfanning: would it be better to just pass a HSSFWorksheet in the input instead of a workbook and sheetNumber?

poi-scratchpad/src/main/java/org/apache/poi/hssf/converter/ExcelToHtmlConverter.java
HasaanA16: Yes, naming is not an issue. It was just my idea. Keeping the name same is also fine

PR #239, opened 2021-07-12
#59388 - Set comment with option isVisible in .xlsx

ryoii:
Fix the bug reported from https://bz.apache.org/bugzilla/show_bug.cgi?id=59388


pjfanning:
Thanks - I merged this - made a small change to fix a broken unit test

PR #240, opened 2021-07-24
Try to make cell copy code more generic

pjfanning:
XSSFRow has beta code to copyRowFrom that uses XSSFCell code. This XSSFCell code is moved to CellUtil in this PR.
* The aim is to facilitate later changes to have copyRowFrom work for HSSF and SXSSF
* The copy style code only currently works for case where source and dest rows are in same XSSFWorkbook. Aim is to allow the styles to be copied to dest workbook. A CellCopyContext has been added to try to keep track of styles that are added to dest workbook so that we don't unnecessarily add too many. No attempt has yet been made to support HSSF styles.


pjfanning:
merged using https://github.com/apache/poi/commit/c12f90378968b3ed7491cdbb36922a72c6dd5dca

PR #243, opened 2021-08-12
Xlookup function

pjfanning:


pjfanning:
A partial implementation was added using https://github.com/apache/poi/commit/c8a3870064b49ce280e87da67740696e001488fd - still plenty to do

PR #252, opened 2021-09-27
Fix spelling errors

aspojo:


pjfanning:
thanks @aspojo - that looks like it should have already been causing issues - but since it doesn't, it could actually be right. Do you have a file that fails to load properly as a result or a test case?

aspojo:
> thanks @aspojo - that looks like it should have already been causing issues - but since it doesn't, it could actually be right. Do you have a file that fails to load properly as a result or a test case?

that code has never been used before. i want to implement XSSFWorkbook's linkExternalWorkbook method, this will be the first it has been used, so i want it to be '.xml' like excel does.

pjfanning:
the change is made on trunk and no tests failed - as you say, this code is not currently used - https://github.com/apache/poi/commit/f2528bdfdc58faae3247fbfe344bd6f944681cc5

pjfanning:
@aspojo could you close this, now that it is merged to svn?

PR #256, opened 2021-09-30
Wrong computation of functionality.

Bhanu17Sharma02:


Bhanu17Sharma02:
I got to know that  Npv is giving a wrong value .

pjfanning:
@Bhanu17Sharma02 can you provide an example that we can use to verify this? For instance, some numbers and what the function should return so that we can verify this change?

Bhanu17Sharma02:
@pjfanning 
Example :
Rate=8%
cfs[]=[-1000, 500, 300, 800]
where 1000 is costs of project and other values are cash flow in cfs[].
Pervious code returns 328.9170387681793 but the correct value is 355.2304018696335 according to  Npv formula.
so,I go through the code and then i  saw a small computational mistake.

pjfanning:
my excel returns 328.92 for npv(0.08, -1000, 500, 300, 800) - so I won't be merging this PR as is

Bhanu17Sharma02:
@pjfanning ok , but take a look in your excel . I think you are  applying this term "-1000/(1+0.08)^1"  also in computation but-1000  should not be divide by (1+0.08)^1.So it is a computational mistake.

pjfanning:
the existing code works with the examples in https://support.microsoft.com/en-us/office/npv-function-8672cb67-2576-4d07-b67b-ac28acf2a568 

I think you have a bug in your logic

We will not be changing the POI code unless you can provide an excel file where the result of the npv calc does not match what POI calculates.

Bhanu17Sharma02:
Thanks for sharing a Microsoft reference, I went through it, 
According to the doc shared by you, the initial investment is not supposed to be provided as an input to the npv function, but is supposed to be added to the final output of the npv function to get the final outcome.
We executed the same function  with the data that you provided and my own dataset, in both the cases the output of your npv function did not match with the output from the excel npv function.Attaching the screenshots and the excel files for both the cases.

[Correct NPV Calculation.xlsx](https://github.com/apache/poi/files/7258927/Correct.NPV.Calculation.xlsx)

NPV Excel
NPV with inbuilt function

navneet79:
Just to add to this @pjfanning , I've been using a python library called numpy-financial that offers the same functions and has the same function signature, the results we get from POI and numpy-financial do not match.
Although the result from numpy-financial matches the Excel results, Attaching a screenshot for reference 
image

pjfanning:
I'm not sure if you are not misusing the POI npv call - in your excel, you have `=NPV(F15,B14:B16)+B13` - in POI, this is FinanceLib.npv(F15, B14:B16) + B13 - so only the last 3 numbers in the cell area `B13:B16` are passed to the NPV function  (`B14:B16`) and then you add B13 separately 

If you use `=NPV(F15,B13:B16)`, you get 32891.7038768179

POI is a Java port of Excel and we have to match Excel, we do not want to match numpy - our users expect us to match Excel

PR #261, opened 2021-10-11
Correct Maximum Width Calculation of unmerged Cells

thoniTUB:
I strumbled upon this code line, it did't cause a problem for me. However it could  cause a problem if at some point merged cells are used, but the unmerge width is requested.


pjfanning:
Thanks - this is merged - will see about test coverage later

thoniTUB:
Alright, your welcome!

pjfanning:
I'll probably need to revert this as it breaks a unit test

pjfanning:
I modified the assertion on the test. Will still need to look into this more to see if this change is actually right.

PR #254, opened 2021-09-28
implement the linkExternalWorkbook method of XSSFWorkbook

aspojo:

PR #264, opened 2021-10-14
keep 'externalLink file names'  the same as excel

aspojo:


pjfanning:
@aspojo would it be possible to supply a test case?

aspojo:
> @aspojo would it be possible to supply a test case?

image
I unziped link-external-workbook-b.xlsx, and these places used to be 0, it will become 1 after I changed the code。(This is also my mistake，you can see [implement the linkExternalWorkbook method of XSSFWorkbook #254](https://github.com/apache/poi/pull/254))

PR #265, opened 2021-10-18
Use BigDecimal to maintain precision of doubles

aschott-looker:
* Multiplying certain numbers by `100.0` causes a loss in precision for certain doubles/floats. This can be demonstrated as follows:
rounding bug
* This causes a value of `0.00035` formatted by `#,##0.00%` to output as `0.03%` when Excel actually outputs `0.04%`. Using `BigDecimal` should remedy this issue.


pjfanning:
Can you provide tests or explain how the modification to test-data/spreadsheet/NumberFormatTests.xlsx tests this change?

Could you also provide some reproducible examples - a calculation that works one way in Excel but not in POI?

aschott-looker:
@pjfanning - Here is the test case I added!

Screen Shot 2021-10-18 at 2 34 38 PM

aschott-looker:
Before my change, POI would output `0.03%`

aschott-looker:
https://stackoverflow.com/questions/177506/why-do-i-see-a-double-variable-initialized-to-some-value-like-21-4-as-21-3999996 has some useful links.

pjfanning:
that stackoverflow link is generic - it is not specifically about excel and how POI can second guess the way that excel does calculations - in excel files, the data is stored as doubles so that does hint towards a requirement to use doubles as opposed to big decimals for calculations

POI is a mature project and not many people are complaining about this so I am reluctant to add this without better and more examples - and unit tests

pjfanning:
@aschott-looker I don't have Excel installed and Google Spreadsheets changes a lot of calculations in NumberFormatTests.xlsx (some weird edge cases). Could you edit your NumberFormatTests and move your new line to the percentages section (approx row 100) and update column D too to have 'Percentages'.

aschott-looker:
@pjfanning - Merged in latest trunk and updated the XLSX file!

aschott-looker:
I see you merged per https://github.com/apache/poi/commit/d14dc1c659d82baae76fa92ba1e43cb169141580 and https://github.com/apache/poi/commit/69eb168051cb4d4f789ffa94878cf907ffc1d5fe

thanks!

PR #269, opened 2021-10-26
Issue with formula .

aspojo:
1. Issue with sheet range in formula with spaces and Quotes .
2. Issue with quote in sheet name.
3. Try to format formula when call setFormula .


pjfanning:
Thanks - I ran some tests and this seems to run ok. We are in the middle of a release so it will be probably be a week before this gets merged.

pjfanning:
thanks - merged

pjfanning:
@aspojo this is causing too many problems - I'm going to basically revert this - see https://github.com/apache/poi/pull/478

The FormulaParser does not retain Table information and the rewriting of the formulas that have references to Tables are all getting broken. I don't have time to enhance the FormulaParser.

PR #270, opened 2021-10-27
Draw correct XDGF circular arc

komarovd95:
Hi, I faced an issue with drawing circular arcs provided from VSDX files (`ArcTo` geometry rows). The current implementation has some bugs: it tries to draw an elliptical arc (see arguments `width` and `height` of `Arc2D.Double` constructor) because chord length equals `2 * radius` only if this chord is a diameter. And it always draws a half of that ellipse (180 degrees angle extent). 

The proposed fix uses 5 simple steps to draw an accurate circular arc: 
1. Find a normal vector to the arc's chord. 
2. Normalise the length of the normal vector (just divide by the length).
3. Set the origin of the normal vector to the midpoint of the chord. 
4. Multiply the normal vector by the height of the arc (aka `a` parameter from the arc). Note that this parameter can be negative (like in example below).
5. The endpoint of this vector is the third point on the circle. Using these 3 points we can draw an elliptical arc by calling an existing method `EllipticalArcTo#createEllipticalArc`. Note that we need to set the parameter `d` of this method to 1 because we draw a circular arc.

This is how my shape is drawn with the current version of Apache POI:
Screenshot 2021-10-27 at 15 55 30

This is how my shape is drawn after the patch (the shape looks like it looks in MS Visio): 
Screenshot 2021-10-27 at 16 19 49

Also, I attached a VSDX file with that diagram (just rename it from `.zip` to `.vsdx`). 
[arcs.zip](https://github.com/apache/poi/files/7426997/arcs.zip)

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xdgf/usermodel/section/geometry/ArcTo.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xdgf/usermodel/section/geometry/ArcTo.java
@@ -17,8 +17,6 @@
 
 package org.apache.poi.xdgf.usermodel.section.geometry;
 
-import java.awt.geom.AffineTransform;
-import java.awt.geom.Arc2D;
 import java.awt.geom.Path2D;
 import java.awt.geom.Point2D;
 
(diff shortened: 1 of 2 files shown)


pjfanning:
@komarovd95 would it be possible to add some unit test coverage?

komarovd95:
@pjfanning sure, will simple unit tests that assert path be enough? Or we need something more complex?

pjfanning:
something basic is enough - ideally something that sets some values, saves the file (to an ByteArrayOutputStream) and read back into a XDGFDocument and assertions made against this

komarovd95:
@pjfanning  sorry for delay. I've added tests for ArcTo but I didn't find an easy way to create VSDX files on the flight

pjfanning:
@komarovd95 thanks - we are finalising a release over the next few days - so I will wait to merge this until after that is complete

PR #273, opened 2021-10-31
trimming sheetnames > 31 chars

manticore-projects:
Warn about trimmed sheet names
Throw exception when trimmed sheetname collides with existing sheet
Implement sheetname trimming for HSSFWorkbook
Add use case tests


poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFWorkbook.java
```diff
@@ -875,8 +876,23 @@ public XSSFSheet createSheet(String sheetname) {
         validateSheetName(sheetname);
 
         // YK: Mimic Excel and silently truncate sheet names longer than 31 characters
+        // Issue a WARNING though in order to prevent a situation, where the provided long sheet name is
+        // not accessible due to the trimming while we are not even aware of the reason and continue to use
+        // the long name in generated formulas
         if(sheetname.length() > 31) {
```
pjfanning: any chance that the 31 could be made into a shared constant?

poi/src/test/java/org/apache/poi/hssf/usermodel/TestHSSFWorkbook.java
```diff
@@ -59,20 +59,13 @@ Licensed to the Apache Software Foundation (ASF) under one or more
...
-import org.apache.poi.ss.usermodel.BaseTestWorkbook;
-import org.apache.poi.ss.usermodel.Cell;
-import org.apache.poi.ss.usermodel.ConditionalFormatting;
-import org.apache.poi.ss.usermodel.ConditionalFormattingRule;
-import org.apache.poi.ss.usermodel.Name;
-import org.apache.poi.ss.usermodel.Row;
-import org.apache.poi.ss.usermodel.Sheet;
-import org.apache.poi.ss.usermodel.SheetConditionalFormatting;
-import org.apache.poi.ss.usermodel.Workbook;
+import org.apache.poi.ss.usermodel.*;
```
centic9: We try to avoid star-imports, please adjust your IDE settings to not use them.

centic9:
Applied most of the changes, however throwing an Exception was not necessary as both HSSF and XSSF already validate for existing sheetnames with taking the max-length into account properly.

manticore-projects:
Awesome! Thank you for looking into this and giving feedback.
Cheers!

PR #275, opened 2021-11-02
Better parallel build performance

manticore-projects:
Following tweaks to Gradle reduces JAR build time by 27%:

1) `gradle clean jar` (after a first full warmup build), on a Ryzen7 5800U (8 physical cores, 16GB Ram)
2) the measurement is not scientific, I just used the Gradle Build time and Variable CPU frequencies and worked in parallel on the computer
3) the measurement has had around 15 seconds variance 
4) original build time was around 08:40 plus/minus 15 sec. new build time reduce to 06:20 plus/minus 15 seconds
5) build still worked on a memory starved laptop (2 Cores, 3 GByte)
6) build was stable, I ran more than 30 tests with various combinations of settings. The commit shows the best/optimal combination.
7) during the build, all CPUs have been used continuously and the machine was under serious load
8) build has been tested with JDK11

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -293,12 +293,14 @@ subprojects {
 				'-Djunit.jupiter.execution.parallel.config.strategy=fixed',
 				'-Djunit.jupiter.execution.parallel.config.fixed.parallelism=2'
 			]
-			maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 2 
+			maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
 		} else {
 			jvmArgs += [
 			    '-Djunit.jupiter.execution.parallel.enabled=true',
 			    '-Djunit.jupiter.execution.parallel.config.strategy=dynamic',
-				'-Djunit.jupiter.execution.parallel.mode.default=concurrent'
+
+                // this setting breaks the test builds, do not use it!
+				//'-Djunit.jupiter.execution.parallel.mode.default=concurrent'
 			]
 			
 			 // Explicitly defining the maxParallelForks was always slower than not setting it
--- a/build.gradle
+++ b/build.gradle
@@ -37,6 +37,7 @@ buildscript {
 
 plugins {
     id 'base'
+    id "com.dorongold.task-tree" version "2.1.0"
     id('org.nosphere.apache.rat') version '0.7.0'
     id 'distribution'
 }
@@ -145,6 +146,7 @@ subprojects {
         options.encoding = 'UTF-8'
         options.compilerArgs << '-Xlint:unchecked'
         options.deprecation = true
+        //options.incremental = true
 
         onlyIf {
             (name != "compileJava9" && name != "compileTest9") || JavaVersion.current() != JavaVersion.VERSION_1_8
@@ -258,7 +260,12 @@ subprojects {
 
         // set heap size for the test JVM(s)
         minHeapSize = "128m"
-        maxHeapSize = "1512m"
+        maxHeapSize = "1G"
+        
+        // Setting the maxParallelForks was always slower than not setting item
+        //maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1 
+        //maxParallelForks = Math.max( Runtime.runtime.availableProcessors() - 1, 1 )
+
 
         // Specifying the local via system properties did not work, so we set them this way
         jvmArgs << [
@@ -271,8 +278,10 @@ subprojects {
             '-Djavax.xml.stream.XMLInputFactory=com.sun.xml.internal.stream.XMLInputFactoryImpl',
             "-Dversion.id=${project.version}",
             '-ea',
-            '-Djunit.jupiter.execution.parallel.config.strategy=fixed',
-            '-Djunit.jupiter.execution.parallel.config.fixed.parallelism=2'
+            '-Djunit.jupiter.execution.parallel.enabled=true',
+            '-Djunit.jupiter.execution.parallel..mode.default=concurrent'
+            // '-Djunit.jupiter.execution.parallel.config.strategy=fixed',
+            // '-Djunit.jupiter.execution.parallel.config.fixed.parallelism=2'
             // -Xjit:verbose={compileStart|compileEnd},vlog=build/jit.log${no.jit.sherlock}   ... if ${isIBMVM}
         ]
 
--- a/build.gradle
+++ b/build.gradle
@@ -22,6 +22,7 @@ import javax.xml.xpath.XPath
 import javax.xml.xpath.XPathConstants
 import javax.xml.xpath.XPathFactory
 
+
 buildscript {
     repositories {
         maven { url 'https://plugins.gradle.org/m2/' }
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -271,8 +278,10 @@ subprojects {
             '-Djavax.xml.stream.XMLInputFactory=com.sun.xml.internal.stream.XMLInputFactoryImpl',
             "-Dversion.id=${project.version}",
             '-ea',
-            '-Djunit.jupiter.execution.parallel.config.strategy=fixed',
-            '-Djunit.jupiter.execution.parallel.config.fixed.parallelism=2'
+            '-Djunit.jupiter.execution.parallel.enabled=true',
+            '-Djunit.jupiter.execution.parallel..mode.default=concurrent'
```
centic9: double dot seems incorrect

gradle.properties
```diff
@@ -1,3 +1,15 @@
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx4096m
\ No newline at end of file
+# Less than 2G definitely slows things down.
+org.gradle.jvmargs=-Xmx2G -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
centic9: Previously we needed more than 2g for running all the tests, but maybe memory requirements did change in the meantime?

Enabling heap-dump always can lead to leftover files and disk filling up in CI, can we at least send it somewhere else, e.g. /tmp?

centic9 (commented):
Thanks for the test-runs, would be very nice to achieve such an improvement of local builds!

There have already been efforts to use parallel building before, e.g. as part of https://svn.apache.org/viewvc?view=revision&revision=1890089 

Later some of it was reverted/changed in https://svn.apache.org/viewvc?view=revision&revision=1892478 

So we should avoid running into the same problems again. Can we have at least a way to configure the run in CI to not run with full parallel build so we can keep CI stable without adding such hard to track-down breakages?

build.gradle
manticore-projects: > So we should avoid running into the same problems again. Can we have at least a way to configure the run in CI to not run with full parallel build so we can keep CI stable without adding such hard to track-down breakages?

Thank you for the feedback. It is definitely possible, although I have one major question: How to detect the "CI" and to know that we are building on the "CI"? 
I have built everything just fine on a 2 Core 3GB machine, is the "CI" anyhow worse equipped? Does the "CI" have a certain distinct hostname or configuration we could use for identifying?

gradle.properties
manticore-projects: > Previously we needed more than 2g for running all the tests, but maybe memory requirements did change in the meantime?

I was able to run everything with Xmx=1GB, but it got slower. I found 2GB the optimal setting.
Also bear in mind, a lots of old Laptop Computers still are equipped with 4GB only (less 1GB for shared graphics making it 3 GB effectively). I think POI still should build on such a Laptop and the original build froze the system on my junk laptop computer.

> Enabling heap-dump always can lead to leftover files and disk filling up in CI, can we at least send it somewhere else, e.g. /tmp?

I will just remove it. This has no direct advantage. Apologies for accidentally "re-introducing" such stuff, I was not aware that there was a history on this already.

build.gradle
manticore-projects: Would it be possible to rely on an Environment Variable, e. g.:
```
boolean isJenkinsBuild = Boolean.valueOf(System.getenv("CI_BUILD"));
boolean isSoftwarePresent = new ProcessBuilder("check software presence").start().waitFor() == 0;
Assume.assumeTrue("Software not present", isSoftwarePresent || isJenkinsBuild);
```
for restricting the build to serial mode?

Would you be able and willing to set the Environment Variable "CI_BUILD" when kicking off Jenkins:
```
pipeline {
            stage('test'){ 

                    sh '''CI_BUILD="TRUE" // setting the env variable in the same shell where you are running mvn
                        mvn test'''
             }
        }
```

Any better idea?

pjfanning: The CI build uses the 'jenkins' task in build.gradle while human users would probably use 'test' task. So, we can have the build.gradle behave differently depending on the task that is called.

manticore-projects:
I gave it my best try:

On CI please call gradle as shown below in order to achieve serial build and Testing on 2 threads. (Calling the `JENKINS` task would set CI_BUILD automatically.)
```bash
CI_BUILD=TRUE gradle --no-parallel --max-workers=1 clean build
```
On a normal build environment, just call
```
gradle clean build
```
in order to achieve parallel build.

build.gradle
```diff
@@ -499,7 +533,9 @@ task allJavaDoc(type: Javadoc) {
 }
 
 
-task jenkins(dependsOn: ['replaceVersion', subprojects.build, 'binDistZip','binDistTar','srcDistZip','srcDistTar']) {}
+task jenkins(dependsOn: ['replaceVersion', subprojects.build, 'binDistZip','binDistTar','srcDistZip','srcDistTar']) {
+	isCIBuild = true;
```
centic9, line 537: I actually use the "jenkins" task also locally as a "Pre-Commit-Check", as otherwise things like "rat-check" and "javadoc" are not run and so you potentially commit changes which break the build.

We can safely set the environment-variables as part of the CI builds via our DSL-based approach, see https://github.com/apache/poi/blob/trunk/jenkins/create_jobs.groovy#L264 for the related code-based CI configuration.

gradle.properties
centic9: Thanks, if the full build via the "jenkins" task still works it should be fine.

build.gradle
centic9: I'd rather keep the "jenkins" task useable and quick locally for developers as well so we can use it as a full local "pre-commit-check".

manticore-projects: I do not think, I damaged this `jenkins` task, I only let it set the global `CI_BUILD` variable (in case it has not been set on the command line).

build.gradle
manticore-projects: Sorry for being slow: does `isCIBuild = true` do any harm?
I could remove that easily of course, but I found it handy when `jenkins` always simulated the CI build.

But yes, adding the Environment Variables and also setting the Gradle Switches at https://github.com/apache/poi/blob/584c8c059b6b317178df635b37244c6e393d6c96/jenkins/create_jobs.groovy#L372 should do.

Can someone please try? :-D

build.gradle
centic9: Sorry, wrong word "useable", I rather meant to keep the env-variable as single place which decides how to run the build. 

If I run "jenkins" locally it would be good to still be able to decide if I want to get the speedup of the parallel building.

build.gradle
centic9: No, sorry for the confusion, it won't do harm except forcing "jenkins" to be slower even when run locally :)

centic9:
I have now applied a slightly adjusted version of these changes. Please take a look if it works for you now.

manticore-projects:
Thank you so much, especially for the fine tuning.
It works well and I am glad that my stuff was partially helpful!
```
BUILD SUCCESSFUL in 6m 33s
```

Cheers!

PR #281, opened 2021-11-23
Change regex to account for decimals with no leading digit

theo-browne:
This PR changes the condition regex to account for the case where the comparison value is a decimal with no preceding digits (i.e [<=.01])


pjfanning:
Thanks - merged - I messed up the commit message but I'll get it into the release notes

PR #285, opened 2021-12-08
Fixes XDGFShape#getTxtPinY method

komarovd95:
There is an obvious bug in the calculation of TxtPinY.


pjfanning:
thanks but any chance you can add some basic test coverage?

PR #290, opened 2021-12-22
Customize Spliterator implementations for better parallelism

daniel-shuy:
# Why

There are many classes in Apache POI that implements [Iterable](https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html), but only implements [Iterable#iterator()](https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html#iterator--) without overriding [Iterable#spliterator()](https://docs.oracle.com/javase/8/docs/api/java/lang/Iterable.html#spliterator--).

The default implementation of [Iterable#spliterator()](https://github.com/openjdk/jdk/blob/9a9add8825a040565051a09010b29b099c2e7d49/jdk/src/share/classes/java/lang/Iterable.java#L100-L102) is:
```java
default Spliterator<T> spliterator() {
    return Spliterators.spliteratorUnknownSize(iterator(), 0);
}
```
[Spliterator#spliteratorUnknownSize(Iterator, int)](https://docs.oracle.com/javase/8/docs/api/java/util/Spliterators.html#spliteratorUnknownSize-java.util.Iterator-int-) returns a `Spliterator` with no initial size estimate, which has poor splitting capabilities. The default implementation has to do this because not all `Iterable`s have a fixed size.

This is important when trying to convert the `Iterable` to a parallel `Stream`. For example, when trying to process [Row](https://poi.apache.org/apidocs/5.0/org/apache/poi/ss/usermodel/Row.html)s in a [XSSFSheet](https://poi.apache.org/apidocs/5.0/org/apache/poi/xssf/usermodel/XSSFSheet.html) (`XSSFSheet` implements `Iterable<Row>`) in parallel, the performance is terrible:
```java
StreamSupport.stream(sheet.spliterator(), true)
    .forEach(row -> {
        // ...
    });
```

The `XSSFSheet`'s `Iterator` is backed by a [TreeMap#values()](https://docs.oracle.com/javase/8/docs/api/java/util/TreeMap.html#values--):
https://github.com/apache/poi/blob/cf1354dc6c6128d8fd024ee1f57359761cd5fd1f/poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFSheet.java#L95
https://github.com/apache/poi/blob/cf1354dc6c6128d8fd024ee1f57359761cd5fd1f/poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFSheet.java#L2049-L2067

`TreeMap#values()` returns a [Collection](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html), and thus has a fixed size. Its `Spliterator` is hence sized and has good splitting capabilities, allowing it to be properly parallelized. Therefore an easy way to customize `XSSFSheet#spliterator()` is to simply delegate to the underlying `TreeMap#values()`, eg.
```java
@Override
@SuppressWarnings("unchecked")
public Spliterator<Row> spliterator() {
    return (Spliterator<Row>)(Spliterator<? extends Row>) _rows.values().spliterator();
}
```

# How

- For classes that expose an `Iterator` factory method with a backing `Collection`/`Iterable`, I simply delegated the `Spliterator` factory method to the underlying `Collection`/`Iterable`.
- For classes that expose an `Iterator` factory method but do not have a backing `Collection`/`Iterable`, but have a fixed size, I customized the `Spliterator` factory method to return a sized `Spliterator` using [Spliterators#spliterator(Iterator, long, int)](https://docs.oracle.com/javase/8/docs/api/java/util/Spliterators.html#spliterator-java.util.Iterator-long-int-).
- Classes that expose an `Iterator` factory method but do not have a fixed size remain unchanged.

I also took the liberty to perform the following refactors:
- Move `iterator()` implementation in child classes to `Workbook`/`Sheet`/`Row` interface using `default` methods (the `iterator()` method is an alias of `sheetIterator()`/`rowIterator()`/`cellIterator()`, and is currently being duplicated in all child classes)
- Add `Iterable` interface to `IntMapper` and `XDDFTextParagraph` (so that they can be iterated using an enhanced `for` loop)

None of the changes break source/binary compatibility.

---

For anyone interested, the current workaround is create a sized `Spliterator` using [Sheet#getPhysicalNumberOfRows()](https://poi.apache.org/apidocs/5.0/org/apache/poi/ss/usermodel/Sheet.html#getPhysicalNumberOfRows--), eg.
```java
Spliterator<Row> spliterator = Spliterators.spliterator(sheet.iterator(), sheet.getPhysicalNumberOfRows(), 0);
StreamSupport.stream(spliterator, true)
    .forEach(row -> {
        // ...
    });
```
This however, is still less optimal than using the backing `Collection`'s `Spliterator`.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/PackageRelationshipCollection.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/PackageRelationshipCollection.java
@@ -374,6 +374,16 @@ public final class PackageRelationshipCollection implements Iterable<PackageRela
     }
 
     /**
+     * Get this collection's spliterator.
+     *
+     * @since POI 5.2.0
+     */
+    @Override
+    public Spliterator<PackageRelationship> spliterator() {
+        return relationshipsByID.values().spliterator();
+    }
+
+    /**
      * Get an iterator of a collection with all relationship with the specified
      * type.
      *
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFRow.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFRow.java
@@ -101,15 +101,6 @@ public class SXSSFRow implements Row, Comparable<SXSSFRow>
     }
 //begin of interface implementation
     /**
-     * {@inheritDoc}
-     */
-    @Override
-    public Iterator<Cell> iterator()
-    {
-        return new FilledCellIterator();
-    }
-
-    /**
      * Use this to create new cells within the row and return it.
      * <p>
      * The cell that is returned is a {@link CellType#BLANK}. The type can be changed
@@ -427,7 +418,7 @@ public class SXSSFRow implements Row, Comparable<SXSSFRow>
     @Override
     public Iterator<Cell> cellIterator()
     {
-        return iterator();
+        return new FilledCellIterator();
     }
 
     /**
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFRow.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFRow.java
@@ -21,6 +21,8 @@ import java.util.Iterator;
 import java.util.Map.Entry;
 import java.util.NoSuchElementException;
 import java.util.SortedMap;
+import java.util.Spliterator;
+import java.util.Spliterators;
 import java.util.TreeMap;
 
 import org.apache.poi.ss.SpreadsheetVersion;
@@ -59,6 +61,11 @@ public class SXSSFRow implements Row, Comparable<SXSSFRow>
     public Iterator<Cell> allCellsIterator()
     {
         return new CellIterator();
+    }
+
+    public Spliterator<Cell> allCellsSpliterator()
+    {
+        return Spliterators.spliterator(allCellsIterator(), getLastCellNum(), 0);
     }
 
     public boolean hasCustomHeight()
@@ -419,6 +426,22 @@ public class SXSSFRow implements Row, Comparable<SXSSFRow>
     public Iterator<Cell> cellIterator()
     {
         return new FilledCellIterator();
+    }
+
+    /**
+     * Create a spliterator over the cells from [0, getLastCellNum()).
+     * Includes blank cells, excludes empty cells
+     *
+     * Returns a spliterator over all filled cells (created via Row.createCell())
+     * Throws ConcurrentModificationException if cells are added, moved, or
+     * removed after the spliterator is created.
+     *
+     * @since POI 5.2.0
+     */
+    @Override
+    @SuppressWarnings("unchecked")
+    public Spliterator<Cell> spliterator() {
+        return (Spliterator<Cell>)(Spliterator<? extends Cell>) _cells.values().spliterator();
     }
 
     /**
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFSheet.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFSheet.java
@@ -24,6 +24,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.Spliterator;
 import java.util.TreeMap;
 
 import org.apache.poi.ss.SpreadsheetVersion;
@@ -501,6 +502,20 @@ public class SXSSFSheet implements Sheet, OoxmlSheetExtensions {
         @SuppressWarnings("unchecked")
         Iterator<Row> result = (Iterator<Row>)(Iterator<? extends Row>)_rows.values().iterator();
         return result;
+    }
+
+    /**
+     *  Returns a spliterator of the physical rows
+     *
+     * @return a spliterator of the PHYSICAL rows.  Meaning the 3rd element may not
+     * be the third row if say for instance the second row is undefined.
+     *
+     * @since POI 5.2.0
+     */
+    @Override
+    @SuppressWarnings("unchecked")
+    public Spliterator<Row> spliterator() {
+        return (Spliterator<Row>)(Spliterator<? extends Row>) _rows.values().spliterator();
     }
 
     /**
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFSheet.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFSheet.java
@@ -100,11 +100,6 @@ public class SXSSFSheet implements Sheet, OoxmlSheetExtensions {
     }
 
     //start of interface implementation
-    @Override
-    public Iterator<Row> iterator() {
-        return rowIterator();
-    }
-
     /**
      * Create a new row within the sheet and return the high level representation
      *
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFWorkbook.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFWorkbook.java
@@ -31,6 +31,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.Spliterator;
 
 import org.apache.commons.compress.archivers.ArchiveOutputStream;
 import org.apache.commons.compress.archivers.zip.Zip64Mode;
@@ -756,6 +757,19 @@ public class SXSSFWorkbook implements Workbook {
     @Override
     public Iterator<Sheet> sheetIterator() {
         return new SheetIterator<>();
+    }
+
+    /**
+     *  Returns a spliterator of the sheets in the workbook
+     *  in sheet order. Includes hidden and very hidden sheets.
+     *
+     * @return a spliterator of the sheets.
+     *
+     * @since POI 5.2.0
+     */
+    @Override
+    public Spliterator<Sheet> spliterator() {
+        return _wb.spliterator();
     }
 
     protected final class SheetIterator<T extends Sheet> implements Iterator<T> {
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFWorkbook.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFWorkbook.java
@@ -787,15 +787,6 @@ public class SXSSFWorkbook implements Workbook {
     }
 
     /**
-     * Alias for {@link #sheetIterator()} to allow
-     * foreach loops
-     */
-    @Override
-    public Iterator<Sheet> iterator() {
-        return sheetIterator();
-    }
-
-    /**
      * Get the Sheet object at the given index.
      *
      * @param index of the sheet number (0-based physical and logical)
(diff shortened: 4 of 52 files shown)


poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/PackageRelationshipCollection.java
```diff
@@ -373,6 +373,14 @@ public PackageRelationshipCollection getRelationships(String typeFilter) {
         return relationshipsByID.values().iterator();
     }
 
+    /**
+     * Get this collection's spliterator.
+     */
```
pjfanning, line 380: could you add `@since POI 5.2.0` on any new public methods?

poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFRow.java
```diff
@@ -100,15 +103,6 @@ public void setCollapsed(Boolean collapsed) {
         this._collapsed = collapsed;
     }
 //begin of interface implementation
-    /**
-     * {@inheritDoc}
-     */
-    @Override
```
pjfanning, line 106: I'm not sure we can remove a public method without first deprecating it

poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFSheet.java
```diff
@@ -100,11 +101,6 @@ public InputStream getWorksheetXMLInputStream() throws IOException {
     }
 
     //start of interface implementation
-    @Override
-    public Iterator<Row> iterator() {
-        return rowIterator();
```
pjfanning, line 105: ditto - please add this back

poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFWorkbook.java
```diff
@@ -786,15 +798,6 @@ public void remove() throws IllegalStateException {
         }
     }
 
-    /**
-     * Alias for {@link #sheetIterator()} to allow
-     * foreach loops
-     */
-    @Override
-    public Iterator<Sheet> iterator() {
```
pjfanning, line 794: add this back

poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFRow.java
```diff
@@ -120,18 +121,14 @@ public XSSFSheet getSheet() {
...
-     *     ...
-     * }
-     * </pre></blockquote>
+     * Cell spliterator over the physically defined cells
      *
-     * @return an iterator over cells in this row.
+     * @return a spliterator over cells in this row.
      */
     @Override
-    public Iterator<Cell> iterator() {
```
pjfanning, line 133: add back the original iterator

poi/src/main/java/org/apache/poi/hssf/usermodel/HSSFWorkbook.java
```diff
@@ -31,17 +31,8 @@ Licensed to the Apache Software Foundation (ASF) under one or more
...
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.NoSuchElementException;
-import java.util.Set;
+import java.util.*;
```
pjfanning: can you keep the explicit imports and remove the * import?

poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFRow.java
daniel-shuy: I've moved it to the `Row` parent interface, so its essentially the same (alias for `cellIterator()`), see https://github.com/apache/poi/pull/290/files#diff-41fe318702760f407b9cfab2d3e3ceb3693f848c2255e3e5c993a0d7b4832624

poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFSheet.java
daniel-shuy: I've moved it to the `Sheet` parent interface, so its essentially the same (alias for `rowIterator()`), see https://github.com/apache/poi/pull/290/files#diff-c77e303b24d9da0a72d46cf8ae7319ba6876b1b5c7d89e6ddc51acea6a6571ce

poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFWorkbook.java
daniel-shuy: I've moved this to the `Workbook` parent interface, so its essentially the same (alias for `sheetIterator()`), see https://github.com/apache/poi/pull/290/files#diff-a4d997b0b0814dae7bbc74ef62d8b413d102a7eae2b40aa5bf1ce4a7840a9233

poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFRow.java
daniel-shuy: I've moved it to the `Row` parent interface, so its essentially the same (alias for `cellIterator()`), see https://github.com/apache/poi/pull/290/files#diff-41fe318702760f407b9cfab2d3e3ceb3693f848c2255e3e5c993a0d7b4832624

poi/src/main/java/org/apache/poi/ss/usermodel/Workbook.java
```diff
@@ -220,6 +222,32 @@ Licensed to the Apache Software Foundation (ASF) under one or more
...
+        return sheetIterator();
+    }
+
+    /**
+     *  Returns a spliterator of the sheets in the workbook
+     *  in sheet order. Includes hidden and very hidden sheets.
+     *
+     * @return a spliterator of the sheets.
+     */
+    default Spliterator<Sheet> sheetSpliterator() {
```
pjfanning: maybe, we only need spliterator() - I know we have 2 iterator methods but that is some legacy stuff - I don't think new stuff needs to add 2 separate methods

poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFRow.java
pjfanning: I see - you can ignore the other comments about removed iterator() methods

poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/PackageRelationshipCollection.java
daniel-shuy: Sure

poi/src/main/java/org/apache/poi/hssf/usermodel/HSSFWorkbook.java
daniel-shuy: Sorry, this was automatically formatted by my IDE, will remove it

poi/src/main/java/org/apache/poi/ss/usermodel/Workbook.java
daniel-shuy: ok, will remove it, I agree its better to have 1 method

poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/PackageRelationshipCollection.java
daniel-shuy: Done

poi/src/main/java/org/apache/poi/hssf/usermodel/HSSFWorkbook.java
daniel-shuy: Done

poi/src/main/java/org/apache/poi/ss/usermodel/Workbook.java
daniel-shuy: Done

pjfanning:
@daniel-shuy generally looks good - but could you add some test coverage for some of the new methods? We don't necessarily need full coverage but some regression tests would be a pre-req for a merge.

poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFWorkbook.java
```diff
@@ -1269,6 +1272,19 @@ public String getSheetName(int sheetIx) {
...
+     * Returns a spliterator of the sheets in the workbook
+     * in sheet order. Includes hidden and very hidden sheets.
+     *
+     * @return a spliterator of the sheets.
+     *
+     * @since POI 5.2.0
+     */
+    @Override
+    public Spliterator<Sheet> spliterator() {
+        return new SheetSpliterator();
```
pjfanning: Is this better than `Spliterators.spliterator(sheets, Spliterator.ORDERED)`?

daniel-shuy: Its actually the same, because `SheetSpliterator` is delegating to `sheets.spliterator()`, which is actually calling `Spliterators.spliterator(this, Spliterator.ORDERED)`.

IMO its still better to delegate to the backing `Collection`'s `spliterator()`, so that if the backing `Collection` changes, there is no risk of the `Spliterator`'s characteristics going out of sync.

I added `SheetSpliterator` to be consistent with `sheetIterator()`, which creates an instance of `SheetIterator` to delegate `sheets.iterator()`. But now that I think about it, its probably better to simply do a cast, reducing an object instantiation, eg.
```java
@Override
@SuppressWarnings("unchecked")
public Spliterator<Sheet> spliterator() {
    return (Spliterator<Sheet>)(Spliterator<? extends Sheet>) sheets.spliterator();
}
```

pjfanning: I don't think you'll need to cast.

daniel-shuy: Unfortunately `sheets.spliterator()` returns `Spliterator<XSSFSheet>`

pjfanning: can you try omitting the cast and see if it compiles? I just added spliterators in another project and was suprised that I didn't need to cast but it worked without the cast - https://github.com/pjfanning/excel-streaming-reader/commit/4ac5fb854498bf18a900c5ef95da6f07a37a7570

daniel-shuy: I tried, it doesn't work. Using `Spliterators.spliterator(this, Spliterator.ORDERED)` avoids having to cast because [Spliterators.spliterator(Collection, int)](https://docs.oracle.com/javase/8/docs/api/java/util/Spliterators.html#spliterator-java.util.Collection-int-) takes in `Collection<? extends T>` and returns `Spliterator<T>`. If only `Iterable<T>#spliterator()` returned `Spliterator<? extends T>` instead of `Spliterator<T>` 🤦‍♂️

pjfanning: fair enough - use the cast

daniel-shuy: Done

PR #293, opened 2022-01-07
issue with rounding in multiply function

pjfanning:
https://bz.apache.org/bugzilla/show_bug.cgi?id=65792


pjfanning:
merged into POI using r1897066

PR #310, opened 2022-03-05
Allow sections to have distinct headers and footers

quimperval:
The below variable was added in the class XWPFHeaderFooterPolicy.java. The variable allows to save a header policy for each section in the document. This variable is used to add header and footer references.

Tests and one example were added.

```
// This allows to have a reference to any section and have distinct headers for
// sections.
private CTSectPr sectPr;
``

![MultiSectionHeaders](https://user-images.githubusercontent.com/37852154/156886953-0b1e278d-6dab-4bfd-a1a5-214b736b22a8.png)
`

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xwpf/model/XWPFHeaderFooterPolicy.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xwpf/model/XWPFHeaderFooterPolicy.java
@@ -16,6 +16,17 @@
 ==================================================================== */
 package org.apache.poi.xwpf.model;
 
+import com.microsoft.schemas.office.office.CTLock;
+import com.microsoft.schemas.office.office.STConnectType;
+import com.microsoft.schemas.vml.CTFormulas;
+import com.microsoft.schemas.vml.CTGroup;
+import com.microsoft.schemas.vml.CTH;
+import com.microsoft.schemas.vml.CTHandles;
+import com.microsoft.schemas.vml.CTPath;
+import com.microsoft.schemas.vml.CTShape;
+import com.microsoft.schemas.vml.CTShapetype;
+import com.microsoft.schemas.vml.CTTextPath;
+import com.microsoft.schemas.vml.STExt;
 import org.apache.poi.ooxml.POIXMLDocumentPart;
 import org.apache.poi.ooxml.POIXMLDocumentPart.RelationPart;
 import org.apache.poi.xwpf.usermodel.XWPFDocument;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: poi-ooxml/src/main/java/org/apache/poi/xwpf/model/XWPFHeaderFooterPolicy.java [hunk 1]
```diff
@@ -52,227 +41,236 @@ Licensed to the Apache Software Foundation (ASF) under one or more
...
  * the right headers and footers for the document.
  */
 public class XWPFHeaderFooterPolicy {
-    public static final Enum DEFAULT = STHdrFtr.DEFAULT;
-    public static final Enum EVEN = STHdrFtr.EVEN;
-    public static final Enum FIRST = STHdrFtr.FIRST;
-
-    private XWPFDocument doc;
-
-    private XWPFHeader firstPageHeader;
```
pjfanning: please do not reformat the file - use the existing format so the changes can be evaluated - we use spaces for indents - not tabs.

Leave the imports in the same order

Don't reformat the existing javadocs

quimperval: Hello

I'll follow what you pointed. 

Is ok if I close this pull request and open another with the right formatted data? 

I'm new in coding, so I'm not sure what is the next step from my side, apart from fixing the data.

pjfanning: You can create a new Pull Request if you prefer. Alternatively, you can add more commits to your existing branch and this Pull Request will be automatically updated.

quimperval: I did a new commit to the branch. 
Thanks for your support!

pjfanning:
Thanks - I added all the changes except for the example - let me have a think about that part.

pjfanning:
@quimperval could you review https://github.com/apache/poi/commit/83a90487cb315e5ee798a11bcff3938dba8b67ac ?

We generally aim to not use the generated CT classes and to use the XWPF classes. The commit above changes your tests to use XWPFRun to set the text inside the paragraphs.

carmceart:
@quimperval Im following your commit and repacked poi-ooxml:4.1.2. 
It doesn't work well.
The docx file worked out appears that last page is blank and  redundant.
Meanwhile, the docx transfers to pdf  with xdocreport ,that performs bad.
How can I comprehend the variable "sectPr"?
beg..

quimperval:
Hello @carmceart 

I've created a branch, from the commit just before the PR 310, and created a simple document with different sections. 

https://github.com/quimperval/poi/tree/simple-multi-section-document

You can find it in the path 

poi-examples/src/main/java/org/apache/poi/examples/xwpf/usermodel/SimpleMultiSectionDocument.java

As you can see in the example, for each paragraph  was added a section properties object, to store data related to section complementary information, in the case of the example to show the behavior, for odd pages the orientation was set as portrait and for even pages was set as landscape. 

The object sectPr stores the date related to page size, and orientation. 

I've opened the generated docx file in libre office and it is showing three pages, one for each section. 
libreOfficeMultiSectionDocument

And I've opened the file with microsoft office online, it is adding an extra page at the end of the three sections.
MsOfficeMultiSectionDocument

I guess Ms Office is adding an extra page because the sectPr object, related to the main body of the document, was not set. So, word adds the empty page. 

I guess it could be possible to get the last sectPr of a paragraph and assign it to the main body, and then replace the section properties object of the paragraph, where we got the sectPr, with a null value. I guess in this case word would be showing the last paragraph without sectPr with the main body sectPr object.

This happens regardless of PR 310.

I could try to create a way to do what I described above and submit as a PR, to prevent an extra page added by word. I need sometime to work on it.

FYI @pjfanning

quimperval:
@carmceart with PR 310 each sectPr will be able to store different properties of headers and footers, which could be different from the ones set for the main body sectPr object, apart from other properties like page size or orientation.

carmceart:
@quimperval 
Thank you so much!
Your commit can sovle this problem!
But next ,I need to transfer docx to pdf. So, I have to control the style with poi code,that is a terribale experience.
A better way I found out to replace "Header", is Using Table as basic layout,the table header row could be used as document header.
Thanks .

PR #312, opened 2022-03-11
close XSSFWorkbook on SXSSFWorkbook.write() method

appbakers:
Please check code change.
Why close is not called after underlying XSSFWorkbook is written?

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFWorkbook.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFWorkbook.java
@@ -967,6 +967,7 @@ public class SXSSFWorkbook implements Workbook {
         try {
             try (FileOutputStream os = new FileOutputStream(tmplFile)) {
                 _wb.write(os);
+                _wb.close();
             }
 
             //Substitute the template entries with the generated sheet data files


pjfanning:
That code change was deliberate as far as I know. Could you change your code to close the SXSSFWorkbook and also to call dispose on the SXSSFWorkbook after you do the write?

This seems to be related to https://bz.apache.org/bugzilla/show_bug.cgi?id=65949

appbakers:
Okay. It seems deliberate. I found error while executing poi test after this change.

so the close should not called inside SXSSFWorkbook.write() method.

I am closing this request.

> Task :poi-ooxml:test

TestSXSSFWorkbook > bug53515() FAILED
    org.opentest4j.AssertionFailedError at TestSXSSFWorkbook.java:354                                                                  
        Caused by: java.io.IOException at TestSXSSFWorkbook.java:354

TestSXSSFBugs > bug60197_NamedRangesReferToCorrectSheetWhenSheetOrderIsChanged() FAILED                           
    java.lang.RuntimeException at SXSSFITestDataProvider.java:76
        Caused by: java.io.IOException at POIXMLDocument.java:215

TestSXSSFBugs > testWriteDocumentTwice() FAILED
    java.io.IOException at POIXMLDocument.java:215

PR #313, opened 2022-03-20
poi-ooxml-lite: Check in list of classes and xsb-files to avoid the jar-to-test dependency

centic9:
This is a PR for discussing a change to the ooxml-lite packages which I would like to commit.

### Problem

Currently executing `./gradlew jar` takes a very looooong time. 

This is because we run all tests including the long-running integration-tests.

The root-cause is that we want to generate a list of all used classes/xsb files by running tests. This is then used to build  a minimized `poi-ooxml-lite` package which has a reduced size and thus is easier to include in some environments.

### Proposed solution

In order to avoid running all tests whenever building jars, this PR adds the list of classes/xsbs to the repository.

This means `./gradlew jar` can run without needing to invoke all tests, speeding it up a lot!

Whenever tests are run and lead to new entries to the files, they should be checked in along the actual changes to the code.

### Resluting improvements

* Obviously many targets run much more quickly now, especially "jar" will finish a lot more quickly.
** This is useful if you want to test with locally-built jar-files
* This is also what everybody else would expect when working with Apache POI sources, i.e. jar should finish quickly without running all tests

### Potential downsides

* When running tests in CI, changes to the new .clazz and .xsb files could happen similar to the Java 9-Class-files
** => Added steps to revert these two files if necessary
* When new clazz-files show up, they need to be added to the files.
** => Running `./gradlew test` or `./gradlew jenkins` locally will still run all tests locally and thus will update the files with new entries as necessary. Those will show up as local changes and should then be committed along with the changed code


build.xml
```diff
@@ -153,7 +153,7 @@ under the License.
     <property name="ooxml.testokfile" location="poi-ooxml/build/testokfile.txt"/>
 
     <property name="ooxml.lite.agent" location="build/dist/maven/poi-ooxml-lite-agent/poi-ooxml-lite-agent-${version.id}.jar"/>
-    <property name="ooxml.lite.report" location="build/ooxml-lite-report"/>
+    <property name="ooxml.lite.report" location="src/resources/ooxml-lite-report"/>
```
centic9, line 156: FYI, I did not change or check if Ant-build still does this correctly, we should really get rid of the Ant build now that we did a few Gradle-based releases.

pjfanning:
Seems like a good idea. What command do you use to regenerate the clazz and xsb list files (if you want to)?

centic9:
New entries to the files are added automatically whenever a test is enhanced and the Gradle-test-target is executed locally.

centic9:
This is applied now via https://github.com/apache/poi/commit/09e464bb5f6d8880e79a2beed45760ddfafb62a7

PR #314, opened 2022-03-25
Add equals() and hashCode() to PaneInformation

daniel-shuy:
Currently there is no way to compare `PaneInformation` equality except by manually comparing all its properties


pjfanning:
Thanks but could you write some test coverage?

daniel-shuy:
@pjfanning done

PR #318, opened 2022-04-02
fix bug: when transfer .ppt(which is encrypted) to .pdf by using poi and itexpdf,can cause NullPointException.

wanglunhui2012:
When i try to transfer .ppt to .pdf,that can work well.But When i try to transfer .ppt(Note: this file is encrypted) to .pdf,the file `DrawTextParagraph.java#getRenderableText()` can cause `NullPointException`.This pr can fix it.


pjfanning:
thanks - merged

PR #322, opened 2022-04-07
fix bug: when transfer .ppt(which include the object such as excel...…

wanglunhui2012:
When i try to transfer .ppt to .pdf,that can work well.But When i try to transfer .ppt(Note: this file include the object such as excel...) to .pdf,that can cause infinite loop.

I found one way to solve this by calling `graphics.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON);` in `HemfImageRenderer.java#drawImage()`

### DIFF
--- a/poi-scratchpad/src/main/java/org/apache/poi/hemf/draw/HemfImageRenderer.java
+++ b/poi-scratchpad/src/main/java/org/apache/poi/hemf/draw/HemfImageRenderer.java
@@ -107,6 +107,8 @@ public class HemfImageRenderer implements ImageRenderer, EmbeddedExtractor {
             return false;
         }
 
+        graphics.setRenderingHint(RenderingHints.KEY_ANTIALIASING, RenderingHints.VALUE_ANTIALIAS_ON);
+
         Charset cs = (Charset)graphics.getRenderingHint(Drawable.DEFAULT_CHARSET);
         if (cs != null && !charsetInitialized) {
             setDefaultCharset(cs);


pjfanning:
Do you have a reproducible test case? We can't just take random code changes with no evidence of the issue.

wanglunhui2012:
@pjfanning
I need some help.I wonder what `org.apache.poi.hwmf.record.HwmfPenStyle.HwmfLineDash.USERSTYLE` is?how it appear?how to deal with it.

kiwiwings:
I'm -1 for this change. you don't simply change graphics context properties without reverting those settings in the middle of drawing stuff.

Regarding the HwmfPenStyle, see the MS Spec [2.1.1.23 PenStyle Enumeration](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-wmf/2af5f7e4-9abb-4fb4-b428-d739c8b23829). IIRC you can specify a dash pattern. If you provide me a .WMF with this enumeration, I'll check this for you (disclaimer: although I've implemented the wmf renderer, I'm currently not very active with POI ...)

wanglunhui2012:
@kiwiwings 
In my case, `org.apache.poi.hwmf.draw.HwmfGraphics#draw(Shape)` will call `graphicsCtx.draw(shape);`, but this will take too long time, `HwmfLineDash lineDash = ps.getLineDash();` return `HwmfLineDash.USERSTYLE`, if don't deal it, it work very well.

wanglunhui2012:
In my cade, I change the `org.apache.poi.hwmf.draw.HwmfGraphics`, I do the change bellow:
```java
public void draw(Shape shape) {
    HwmfPenStyle ps = getProperties().getPenStyle();
    if (ps == null) {
        return;
    }
    HwmfLineDash lineDash = ps.getLineDash();
    if (lineDash == HwmfLineDash.NULL) {
        // line is not drawn
        return;
    }
    // --------- Change ---------
    if(lineDash == HwmfLineDash.USERSTYLE){
        return;
    }
    // --------- Change ---------
    BasicStroke stroke = getStroke();
    // first draw a solid background line (depending on bkmode)
    // only makes sense if the line is not solid
    if (getProperties().getBkMode() == HwmfBkMode.OPAQUE && (lineDash != HwmfLineDash.SOLID && lineDash != HwmfLineDash.INSIDEFRAME)) {
        graphicsCtx.setStroke(new BasicStroke(stroke.getLineWidth()));
        graphicsCtx.setColor(getProperties().getBackgroundColor().getColor());
        graphicsCtx.draw(shape);
    }
    // then draw the (dashed) line
    graphicsCtx.setStroke(stroke);
    graphicsCtx.setColor(getProperties().getPenColor().getColor());
    graphicsCtx.draw(shape);
}
```

wanglunhui2012:
> Do you have a reproducible test case? We can't just take random code changes with no evidence of the issue.

I found this happened when `Graphics2D` draw long path, infinite loop occur in `sun.dc.pr.PathDasher.appendLine()`:
```java
int width=960,height=540;
BufferedImage bufferedImage=new BufferedImage(width,height,BufferedImage.TYPE_INT_RGB);
Graphics2D graphics2D = bufferedImage.createGraphics();
graphics2D.setStroke(new BasicStroke(1, BasicStroke.CAP_BUTT,
        BasicStroke.JOIN_ROUND, 1.0f,
        new float[] {1.0f, 1.0f}, 0));
graphics2D.drawRect(0, 0, 2147483627, 2147483627);
```

kiwiwings:
Thank you for analyzing it that far - this is really helpful, but ...
Solving the infinite loop problem can be done in two ways - excluding/ignoring the unknown or trying to understand how to react on that userstyle structure. I don't know if I can successfully go the second approach, but I rather would try to process your .wmf instead of excluding the processing. If you like, you can provide it to me privately to my <username>@apache.org email.

wanglunhui2012:
@kiwiwings May I take another pr, which ignore the `USERSTYLE`? Because I see the comment in `org.apache.poi.hwmf.record.HwmfPenStyle`: 
```java
/**
 * The pen uses a styling array supplied by the user.
 * (this is currently not supported and drawn as solid ... no idea where the user
 * styling is supposed to come from ...)
 */
USERSTYLE(0x0007, null);
```
If drawn as solid, will cause infinite loop.

wanglunhui2012:
@pjfanning Sorry, the file is inner and I cannot find better way to reproducible.

kiwiwings:
> @kiwiwings May I take another pr, which ignore the `USERSTYLE`? Because I see the comment in `org.apache.poi.hwmf.record.HwmfPenStyle`

My approach to this would be: "Hey boss/customer, we have a problem with this .wmf. I've already proposed a simple solution, but they insist to fix it properly. Although they don't know if they are able to fix it, may I send them the file privately (as in, it doesn't get included in the open source project for everyone to see) nevertheless?"

PR #323, opened 2022-04-11
Make XSSFSheet's getRows(int, int, boolean) public instead of private

vladmashk:
It is a very useful method and it would be nice if we could use it.


pjfanning:
@vladmashk thanks. I'm neutral about this change. If you were to add some tests, then I'd be a bit more supportive.
This is easily implemented inside your own code as opposed to us bloating the API with more methods that need support and tests - but if this PR had tests, then maybe. We have no releases planned for a while btw.

PR #325, opened 2022-04-16
chore: Set permissions for GitHub actions

nathannaveen:
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

 Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

Signed-off-by: nathannaveen <42319948+nathannaveen@users.noreply.github.com>

### DIFF
--- a/.github/workflows/test-gradle.yml
+++ b/.github/workflows/test-gradle.yml
@@ -9,6 +9,9 @@ on:
   pull_request:
     branches: [ trunk ]
 
+permissions:
+  contents: read
+
 jobs:
   build:
 
--- a/.github/workflows/test-gradle.yml
+++ b/.github/workflows/test-gradle.yml
@@ -9,8 +9,7 @@ on:
   pull_request:
     branches: [ trunk ]
 
-permissions:
-  contents: read
+permissions: {}
 
 jobs:
   build:


pjfanning:
@nathannaveen thanks for drawing attention to this setting. I read https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/ and this seems to suggest that 'none' is the default - isn't that better than granting 'read' - since nothing in this yml file needs the github token?

nathannaveen:
> @nathannaveen thanks for drawing attention to this setting. I read https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/ and this seems to suggest that 'none' is the default - isn't that better than granting 'read' - since nothing in this yml file needs the github token?

I agree, and have updated the permissions to none by the recommended syntax from https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#overview.

PR #330, opened 2022-04-29
Patch averageif

sfisque:
adds missing support for singular AverageIf function

### DIFF
--- a/poi/src/main/java/org/apache/poi/ss/formula/functions/AverageIf.java
+++ b/poi/src/main/java/org/apache/poi/ss/formula/functions/AverageIf.java
@@ -28,26 +28,22 @@ import org.apache.poi.ss.formula.functions.CountUtils.I_MatchPredicate;
  * Handler for singular AverageIf which has different operand handling than
  * the generic AverageIfs version.
  */
-public class AverageIf
-extends Baseifs
-{
+public class AverageIf extends Baseifs {
+    
     public static final FreeRefFunction instance = new Averageifs();
 
     
     @Override
     public ValueEval evaluate( ValueEval[] _args, OperationEvaluationContext _ec )
     {        
-        if( _args.length < 2 ) 
-        {
+        if( _args.length < 2 ) {
             return ErrorEval.VALUE_INVALID;
         }
         
-        try 
-        {
-             AreaEval sumRange = null;
+        try {
+             AreaEval sumRange = convertRangeArg( _args[ 0 ] );
              
-            if( _args.length == 3 )
-            {
+            if( _args.length == 3 ) {
                 sumRange = convertRangeArg( _args[2] );
             }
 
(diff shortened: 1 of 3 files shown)


poi/src/main/java/org/apache/poi/ss/formula/functions/AverageIf.java
```diff
@@ -0,0 +1,133 @@
...
+import org.apache.poi.ss.formula.eval.ValueEval;
+import org.apache.poi.ss.formula.functions.CountUtils.I_MatchPredicate;
+
+/**
+ * Handler for singular AverageIf which has different operand handling than
+ * the generic AverageIfs version.
+ */
+public class AverageIf
+extends Baseifs
+{
```
centic9: We try to follow a code-style where brackets are on the previous line, can you re-format this class to follow that, also "extends" is usually on the same line as the class, i.e. 

public class AverageIf extends Baseifs {

poi/src/test/java/org/apache/poi/ss/formula/functions/TestAverageIf.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    private static void confirmError(ErrorEval errorEval, ValueEval[] args) {
+        ValueEval actualEval = invokeAverageif(args);
+        assertEquals(errorEval, actualEval);
+    }
+
+    /**
+     *  Example 1 from
+     *  https://support.microsoft.com/en-us/office/maxifs-function-dfd611e6-da2c-488a-919b-9b6376b28883
```
pjfanning: should this link be https://support.microsoft.com/en-us/office/averageif-function-faec8e2e-0dec-4308-af69-f5576d8ac642 ? the test data does not match either of the examples on that doc page though

poi/src/main/java/org/apache/poi/ss/formula/functions/AverageIf.java
sfisque: ok. i've never been a fan of K&R formatting, but i can make that change.

poi/src/test/java/org/apache/poi/ss/formula/functions/TestAverageIf.java
sfisque: yes.  the danger of cut/paste coding (i stole the UT from averageifs) for expedience.  i can make the change.

pjfanning: can you use the examples from https://support.microsoft.com/en-us/office/averageif-function-faec8e2e-0dec-4308-af69-f5576d8ac642 though - ie the data and test scenarios?

pjfanning: the M$ examples are rubbish - very focused on a small number of happy path cases and occasionally an exception case but at least using them as a guide is better than ignoring them and focusing a use case that suits you and maybe noone else

pjfanning:
could you cover example2 from M$ too?

pjfanning:
and you ignored the review comment asking you to respect the code format

sfisque-synodex:
ok.  but i'm not aware of "ignoring" the code format request.  i moved all the braces and other bits.

pjfanning:
I've merged what you've done so far after reformatting the code but with example2 being tested, the new code is suspect and liable for removal - https://github.com/apache/poi/commit/8a0d0d78b4aa6cdacec489b76d2849ecef87da97

sfisque-synodex:
i was in the process of adding ex.2 to the test.  should i continue?

pjfanning:
new code is incorrect - I have provided a pull request to your repo - if the example2 is left broken, i will remove all the new code

pjfanning:
Can you close this PR? - all this code is  merged to Apache trunk and I have been fixing your bugs and code formatting issues.

sfisque-synodex:
i do not appear to have that privilege.  i cannot find the close button.  you have my permission to close it if you have that privilege.

pjfanning:
I don't have a direct way to close this. I have un ugly way to close it but may avoid it. It's more important that you verify the apache poi trunk code to verify it works for you.

sfisque:
no worries, i forgot that i forked it under a different user-id.  closed

PR #334, opened 2022-06-04
Add SHRINK_TO_FIT and QUOTE_PREFIXED to CellUtil

funakawa718:
I found `CellUtil` doesn't support `shrinkToFit` and `quotePrefixed`.
Therefore, a `setCellStyleProperty()` on a cell with `shrinkTofit` set to `true` will revert to the default of `false`.

I've challenged myself to create a fix. Could you please have a moment to take a look at it?

### DIFF
--- a/poi/src/test/java/org/apache/poi/ss/util/BaseTestCellUtil.java
+++ b/poi/src/test/java/org/apache/poi/ss/util/BaseTestCellUtil.java
@@ -23,6 +23,7 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 
 import java.io.IOException;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


poi/src/test/java/org/apache/poi/ss/util/BaseTestCellUtil.java
```diff
@@ -104,6 +105,126 @@ void setCellStylePropertyBorderWithShortAndEnum() throws IOException {
...
+            assertEquals(FillPatternType.DIAMONDS, c.getCellStyle().getFillPattern());
+            assertEquals(f.getIndex(), c.getCellStyle().getFontIndex());
+            assertTrue(c.getCellStyle().getHidden());
+            assertEquals((short) 3, c.getCellStyle().getIndention());
+            assertFalse(c.getCellStyle().getLocked());
+            assertEquals((short) 45, c.getCellStyle().getRotation());
+            assertEquals(VerticalAlignment.CENTER, c.getCellStyle().getVerticalAlignment());
+            assertTrue(c.getCellStyle().getWrapText());
+            assertTrue(c.getCellStyle().getShrinkToFit());
+            assertTrue(c.getCellStyle().getQuotePrefixed());
```
funakawa718, line 224: Before the fix, `c.getCellStyle().getShrinkToFit()` would be `false` here and the test would fail.

pjfanning:
Thanks - merged with https://github.com/apache/poi/commit/d074e8b4eb3cd2c9eb996ef3a6b13391c54a8e51

funakawa718:
Thanks for the quick response and merge!

PR #338, opened 2022-06-29
Bug 51037 - apply default column style if no style set

pjfanning:
https://bz.apache.org/bugzilla/show_bug.cgi?id=51037

* still needs SXSSF support and testing
* needs more XSSF testing


pjfanning:
merged with https://github.com/apache/poi/commit/86cc8d1b4b4e84470d45a67a4abb7a034612fd6a

PR #339, opened 2022-07-15
use StringUtils's isNotBlank and isBlank instead of trim().length() or trim().isEmpty() etc

XenoAmess:
…


build.gradle
```diff
@@ -141,6 +141,7 @@ subprojects {
     version = '5.2.3-SNAPSHOT'
     ext {
         bouncyCastleVersion = '1.70'
+        commonsLang3Version = '3.12.0'
```
pjfanning: This appears in another PR you added and again, this does not seem important enough to add another dependency.

XenoAmess: Or you can just copy the functions needed, means isBlank and isNotBlank, from commons-lang3, which is an apache v2 licensed repo from apache, so get some codes from there seems resonable, as this is also an apache v2 licensed repo from apache

pjfanning: POI has its own StringUtil class - you can add extra methods there.

XenoAmess:
@pjfanning done. please review again.

pjfanning:
thanks - merged

PR #348, opened 2022-07-15
refine int cast long

XenoAmess:


pjfanning:
applied some of these changes - rebase and see if you want to get the rest merged

poi-ooxml/src/test/java/org/apache/poi/xssf/usermodel/TestSXSSFBugs.java
```diff
@@ -185,7 +185,7 @@ void test62872() throws Exception {
                 cell = row.createCell(j - 1);
 
                 //make some noise
-                cell.setCellValue(new Date(i * TEN_MINUTES + (j * TEN_MINUTES) / COLUMN_COUNT));
+                cell.setCellValue(new Date((long) i * TEN_MINUTES + (j * TEN_MINUTES) / COLUMN_COUNT));
```
pjfanning, line 188: this is test code - not worth changing

poi-scratchpad/src/main/java/org/apache/poi/hslf/record/Notes.java
```diff
@@ -29,7 +29,7 @@ Licensed to the Apache Software Foundation (ASF) under one or more
 public final class Notes extends SheetContainer
 {
     private byte[] _header;
-    private static long _type = 1008l;
+    private static long _type = 1008L;
```
pjfanning, line 32: why change the l to an L?

xzel23: It's often recommended to use L instead of l because in some fonts it's easy to confuse l and 1.

PR #351, opened 2022-07-15
add break to loop

XenoAmess:


pjfanning:
@XenoAmess I added a slightly different change using https://github.com/apache/poi/commit/186ad50d1816fc05b9a06b204e421fe5598fa331. Please close this PR if you are happy.

XenoAmess:
@pjfanning suggest change it to foreach, but it can do.

PR #352, opened 2022-07-15
use Paths

XenoAmess:


pjfanning:
This is a big diff - and I'm not sure what we gain from it - what is so wrong with FileOutputStream?

poi-examples/src/main/java/org/apache/poi/examples/xwpf/usermodel/BetterHeaderFooterExample.java
```diff
@@ -48,7 +49,7 @@ public static void main(String[] args) throws IOException {
             XWPFFooter foot = doc.createFooter(HeaderFooterType.DEFAULT);
             foot.createParagraph().createRun().setText("footer");
 
-            try (OutputStream os = new FileOutputStream(new File("header2.docx"))) {
+            try (OutputStream os = Files.newOutputStream(new File("header2.docx").toPath())) {
```
kdebski85, line 52: Instead of creating `File` object and then using `toPath()` to convert to `Path`, you can directly create `Path` object with `java.nio.file.Paths::get` static method.

cowwoc:
@pjfanning One of benefits of using `Path` instead of `File` is that it abstracts away whether you're reading normal (expanded) files, or files inside compressed files like JAR, ZIP, etc. Overall, it is more flexible and `Files` provides a lot of useful utility methods that you can build on.

PR #356, opened 2022-07-15
refine conditions

XenoAmess:


pjfanning:
I applied some of the changes using https://github.com/apache/poi/commit/773ad5e8251a43b822793cb265a04a80793be4c7 -  I'm not convinced by the other ones

poi/src/main/java/org/apache/poi/hpsf/TypedPropertyValue.java
```diff
@@ -255,7 +255,7 @@ static void skipPadding( LittleEndianByteArrayInputStream lei ) {
         for (int i=0; i<skipBytes; i++) {
             lei.mark(1);
             int b = lei.read();
-            if (b == -1 || b != 0) {
+            if (b != 0) {
```
XenoAmess, line 258: I think it obvious here... whenever `b == -1` there always be `b != 0`

pjfanning: committed this one just now

PR #360, opened 2022-07-15
bugfix rgba not set

XenoAmess:


pjfanning:
Thanks @XenoAmess - I committed this with a new test case.

PR #363, opened 2022-07-21
use async task to avoid calculating dowstream result in same thread

pjfanning:
Relates to https://bz.apache.org/bugzilla/show_bug.cgi?id=66152

This is work in progress.

* See the uncommented test in this PR to see how to reproduce the issue.
* By calculating the value of the cell that we rely on using an async task, we avoid creating a Java call stack that can get too big if cells depend on cells that depend on other cells, etc. - too many layers of nesting and we will stack overflow
* The main downside is that we have more threads - in this case, a ForkJoinPool
* We could make this configurable, maybe a static method on WorkbookEvaluator that allows you to enable/disable the use of the ForkJoinPool
  * configuring POI is a bit of a mess - many different classes with their own static methods
  * probably too late to try to centralise the config

### DIFF
--- a/poi/src/main/java/org/apache/poi/ss/formula/WorkbookEvaluator.java
+++ b/poi/src/main/java/org/apache/poi/ss/formula/WorkbookEvaluator.java
@@ -24,6 +24,7 @@ import java.util.IdentityHashMap;
 import java.util.Map;
 import java.util.Stack;
 import java.util.TreeSet;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.Future;
 
@@ -58,7 +59,9 @@ public final class WorkbookEvaluator {
 
     private static final Logger LOG = LogManager.getLogger(WorkbookEvaluator.class);
 
-    private static ForkJoinPool forkJoinPool = new ForkJoinPool();
+    private static ExecutorService executorService;
+    private static boolean useExecutorService = false;
+    private static final Object executorServiceLock = new Object();
 
     private final EvaluationWorkbook _workbook;
     private EvaluationCache _cache;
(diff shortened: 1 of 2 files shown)


rlwhitcomb:
Could this potentially just run out of threads and deadlock instead of stack overflow?

centic9:
I would opt for configuration, potentially even default off for backwards compatibility. 
Many users (including some of my own use-cases) may not want to have uncontrolled threads being spun off, e.g. in large WebServers or large and complex business applications where resource usage is guarded and regulated.

pjfanning:
> Could this potentially just run out of threads and deadlock instead of stack overflow?

Fork Join Pools use a very limited number of threads. It uses a work stealing concept - they don't work like Thread Pools.

Still, the plan is to make this non-default behaviour.

pjfanning:
The use of the fork join pool (when enabled) is not to parallelise the job. The task is run on the executor and we then block waiting for the task. For a FormulaEvaluator, you will only ever have 1 async task running at a time. The aim is to stop the Java call stack getting too big by avoiding recursion.

pjfanning:
I've done some more testing and this change does not work reliably. The EvaluationTracker class has a _evaluationFrames list and this assumes all cells are evaluated in order - which cannot be guaranteed when async logic is brought in. Reworking EvaluationTracker to account for async logic looks non-trivial.

centic9:
So this is mostly a workaround to avoid Stackoverflow due to recursion? 
The usual solution for this is to replace recursion with a stack+loop, is this possible here?

pjfanning:
Yes, it would be possible to rewrite the code to do a solution that doesn't need async tasks. It would probably be a much bigger change though.

The new algorithm would probably involve doing an initial pass to work out which cells the target cell relies on and then work out which cells those rely on until you end up with no new cells to check. You would need to maintain a Set of cells that you are yet to check and produce a List of cells in the order that they need to be calculated. On the 2nd pass, you would process the List of cells output from the 1st pass and calculate the results for each.

pjfanning:
I'm closing this. The issue generally is not a high priority. I may try another approach later on.

PR #367, opened 2022-08-06
Avoid long literals ending with lowercase 'l'. These literals may be …

arturobernalg:
Avoid long literals ending with lowercase 'l'. These literals may be confusing, as lowercase 'l' looks very similar to '1'. Use 'L' instead


pjfanning:
thanks - merged

PR #368, opened 2022-08-06
Remove unnecessarily escaped characters.

arturobernalg:


poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/PackagePartName.java
```diff
@@ -326,7 +326,7 @@ private static void checkPCharCompliance(String segment)
             /* Check rule M1.7 */
             if (decodedChar == '/' || decodedChar == '\\') {
                 throw new InvalidFormatException(
-                    "A segment shall not contain percent-encoded forward slash ('/'), or backward slash ('\') characters. [M1.7]");
+                        "A segment shall not contain percent-encoded forward slash ('/'), or backward slash ('') characters. [M1.7]");
```
pjfanning: this change is wrong

pjfanning: or at least - I think the aim is to produce the text `'\'` - the fix might be to use `'\\'`

arturobernalg: reverted.

pjfanning:
@arturobernalg we don't use git - we use svn. Usually, if the Github PR is worthwhile, I'll go through the effort of creating a svn commit. In this case, I don't think the new code is a big improvement. I'm going to opt out of this. Maybe some other POI dev will like the change and apply it.

PR #371, opened 2022-08-18
fix: when copy slide,new slide use old textrun object reference

FlyingPigQAQ:


pjfanning:
Please provide some description - and ideally a test case to go with this. It is not enough to modify code without describing why.

pjfanning:
It is also better to raise a bug first before providing PRs. See https://bz.apache.org/bugzilla

FlyingPigQAQ:
> It is also better to raise a bug first before providing PRs. See https://bz.apache.org/bugzilla

Hello, I read the buglist and found an issue with the same problem but not fixed. The address is as follows. https://bz.apache.org/bugzilla/show_bug.cgi?id=65473

FlyingPigQAQ:
Unit tests were added.

pjfanning:
merged - thanks

PR #377, opened 2022-08-28
Fix issue with orphaned (in package) images and notes post slide removal.

gffloodg:
When removing a slide from a loaded (XMLSlideShow) presentation, it is observed that the resulting output, whilst removing the actual slides leaves dangling (orphaned) artefacts in the written package. i.e. the notes slides and image data are left in the package. This results in larger than needed package output - something that is considerably visible when the original deck is of a large size (e.g. 100MB+) 

This PR resolves the issue by ensuring that the package part relations are correctly removed for note and picture relations when a slide is remove from a deck.


pjfanning:
Thanks - merged

PR #389, opened 2022-10-18
Insert paragraphs and tables into `XWPFDocuments` recursively

AntonOellerer:
Up until now, the `XWPFDocument#insertNew[Paragraph|Table](XmlCursor)` methods were only able to insert the elements if they were present in the `XWPFDocument`-internal lists representing the top-level objects not nested in other elements  (`XWPFDocument.paragraphs`/`XWPFDocument.tables`).

This commit updates the insertion code for both methods such that objects can also be inserted if the `XmlCursor` does not point
to an element in the top-level list but to a nested object (for example a paragraph in a table).
This is done by first figuring out the path to the cursor by repeated calls to `XmlCursor#toParent()`, and then traversing down
the `XWPFDocument` internal structure to find the parent object where the new paragraph or table should be inserted.

Tests verifying this behavior have been added.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFDocument.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFDocument.java
@@ -33,6 +33,8 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Optional;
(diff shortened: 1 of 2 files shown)


pjfanning:
Thanks - merged

AntonOellerer:
Hey, do you already have any idea when this/poi 5.2.4 will be released?

pjfanning:
This is not the place to discuss releases. Try the dev@poi.apache.org mailing list.

PR #403, opened 2022-12-01
ignoring errors on ReadOnlySharedStringsTable count parsing, nobody i…

yuval-perelman:
ignoring errors on ReadOnlySharedStringsTable count parsing, nobody is using it. If from some reason the value of the input is invalid, there is no reason to fail the process especially since all the spreadsheet readers (Excel, Libre, Google...),  can handle it.


pjfanning:
What is the provenance of the xlsx that has a malformed SST file? 
I'd personally be reluctant to add this unless malformed SSTs were common.
There is a 2nd POI class called SharedStringsTable. That needs be kept in sync functionality wise. Also, the Boolean variable, malformedCount, is badly named.

matafagafo:
> What is the provenance of the xlsx that has a malformed SST file? I'd personally be reluctant to add this unless malformed SSTs were common. 

I want to put my 2 cents here.
We use POI to read user uploaded files, and we already received some XLSX files with invalid counts, bellow a sample..

POI generates `NumberFormatException` parsing the file.
Trace:
```
java.lang.NumberFormatException: For input string: "4294749708" 
     at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
     at java.lang.Integer.parseInt(Integer.java:583)
     at java.lang.Integer.parseInt(Integer.java:615)
     at org.apache.poi.xssf.eventusermodel.ReadOnlySharedStringsTable.startElement(ReadOnlySharedStringsTable.java:244)

``````
The user file  SST was:
`<sst xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main" count="4294749708" uniqueCount="566">`

After opening with Excel (**without** errors or warnings showed by Excel) and Saving As, the SST was corrected to: 
`<sst xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main" count="2452" uniqueCount="566">`

I hope this helps to answer your question.

pjfanning:
I added https://github.com/apache/poi/commit/fdffe5f66d1c4c09b1edf77f009c3950f60b9ed0 as a temporary solution - I also added your test case after that.

pjfanning:
Another option is to use https://github.com/pjfanning/poi-shared-strings instead - the shared strings table implementations there are tolerant of large count values. I've added a test case to that lib to validate this. It is likely to be a few months before there is an official POI release with my recent poi-ooxml changes.

yuval-perelman:
@pjfanning looks like its common enough for me not to be the only one who ran into this issue in a relatively short period of time in a pretty small forum, I have no way of knowing how common it actually is. I suspect that a common system that generates XLSX generates malformed SST files, but so far I got no answer from the client who sent this file as to how that file was created. Thanks a lot for the fix, highly appreciated.

PR #404, opened 2022-12-08
Fix issue with text runs not honouring specific styling if they are i…

gffloodg:
…inside a table cell.

When rendering PPTX containing tables, it can be seen that text runs within cells with certain styles (b / i / color) are not correctly rendered. 

This is caused by a bug in the creation of table cell text runs.
This PR fixes the issue and adds tests to verify the change.


pjfanning:
Thanks - merged

dinox0r:
This causes the following exception in POI 5.2.4:
```none
java.lang.NullPointerException: Cannot invoke "org.openxmlformats.schemas.drawingml.x2006.main.CTTextCharacterProperties.isSetB()" because "rPr" is null
	at org.apache.poi.xslf.usermodel.XSLFTableCell$XSLFCellTextRun.isBold(XSLFTableCell.java:769)
```

pjfanning:
@higuaro I think this might be fixed since then - see https://github.com/apache/poi/commit/360c05d9e3496cb6b49e73732615fd9daedf40aa

Hopefully, a POI 5.2.5 patch will be released in a few weeks.

PR #405, opened 2022-12-12
Use `Characters.toChars` instead of `StringBuilder.appendCodePoint` to improve performance

StanBertrand:


pjfanning:
Thanks - merged

PR #406, opened 2022-12-12
Add the ability to avoid adding cellStyle to improve writing performance

StanBertrand:
In [bug-51037] we apply the default column style if no style is set (https://svn.apache.org/viewvc?view=revision&revision=1902391)
We change :
```
    @Override
    public CellStyle getCellStyle()
    {
        if(_style == null){
            SXSSFWorkbook wb = (SXSSFWorkbook)getRow().getSheet().getWorkbook();
            return wb.getCellStyleAt(0);
        } else {
            return _style;
        }
    }
```
by : 
```
@Override
    public CellStyle getCellStyle()
    {
        if (_style == null) {
            CellStyle style = getDefaultCellStyleFromColumn();
            if (style == null) {
                SXSSFWorkbook wb = getSheet().getWorkbook();
                style = wb.getCellStyleAt(0);
            }
            return style;
        } else {
            return _style;
        }
    }
    private CellStyle getDefaultCellStyleFromColumn() {
        CellStyle style = null;
        SXSSFSheet sheet = getSheet();
        if (sheet != null) {
            style = sheet.getColumnStyle(getColumnIndex());
        }
        return style;
    }
```
When we try to write a lot of rows, we have this flamegraph : 
Capture d’écran 2022-12-12 à 16 49 05

The method `SXXSFCell.getCellStyle` can take approximatively 20% of `SheetDataWriter.writeCell`.

I add the `ignoreCellStyle` because if I want to write a simple excel file without style, I can skip this call.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/GZIPSheetDataWriter.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/GZIPSheetDataWriter.java
@@ -49,6 +49,14 @@ public class GZIPSheetDataWriter extends SheetDataWriter {
     }
 
     /**
+     * @param sharedStringsTable the shared strings table, or null if inline text is used
+     * @param ignoreCellStyle whether to use cell style
+     */
+    public GZIPSheetDataWriter(SharedStringsTable sharedStringsTable, boolean ignoreCellStyle) throws IOException {
+        super(sharedStringsTable, ignoreCellStyle);
+    }
+
+    /**
      * @return temp file to write sheet data
      * @deprecated no need for this be public, will be made private or protected in an upcoming release
      */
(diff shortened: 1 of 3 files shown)


pjfanning:
Thanks for the PR but could I highlight that leaving PRs with no description or background is not a good idea?

I would take some convincing that adding this extra complication to the SXSSF is worth it. Do you have any proof that cell styles are expensive? I'm very sceptical about this.

StanBertrand:
Sorry! I added a comment

pjfanning:
Could you trying calling setCellStyle(style) on each cell? You can have just 1 style instance that you set for all cells.
If you look at SXSSFCell.getCellStyle, it skips all the lookup code if you have pre-set the style.

pjfanning:
I also added https://github.com/apache/poi/commit/7454a722f0d408afc605cb1b3cbb2daa9c5c913c - this should improve the performance of SXSSFCell getColumnIndex.
getColumnIndex is one of the main reasons that the cell style code is slow.

StanBertrand:
Thanks 👍

PR #408, opened 2022-12-21
Increase performance of StylesTable.putStyle by 100%

mirraj2:
When exporting a large spreadsheet, this method took 2.3 seconds of execution time.  See VisualVM screenshot:

image

This change would ensure that only one linear lookup is made into the xfs collection instead of the current two lookups per invocation of this method.


pjfanning:
Thanks - merged

PR #409, opened 2022-12-23
Textruns not honouring highlight

gffloodg:
This PR adds support for getting/setting the highlight color in TextRuns. It also adds support for rendering of the same. As usual, appropriate tests also added.


poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFTextRun.java
```diff
@@ -157,6 +160,52 @@ private static void fetchFontColor(CTTextCharacterProperties props, Consumer<Pai
         }
     }
 
+    @Override
```
pjfanning, line 175: can you add javadoc on public API methods? include `@since POI 5.2.4`

poi-scratchpad/src/main/java/org/apache/poi/hslf/usermodel/HSLFTextRun.java
```diff
@@ -440,6 +440,23 @@ public void setFontColor(PaintStyle color) {
...
     }
 
+
+    @Override
+    public PaintStyle getHighlightColor() {
+        return null;
+    }
+
+    @Override
+    public void setHighlightColor(final Color color) {
```
pjfanning: why add these methods? - if this stuff is only supported on XSLFTextRun, then only that class should have these methods.

gffloodg: I agree, but the common rendering in DrawTextParagraph does not have access to the XSLF* classes, so cannot be cast - I chose this as a tradeoff. If you have a better suggestion then Im happy to adjust as needed.

poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFTextRun.java
gffloodg: Done

poi/src/main/java/org/apache/poi/sl/usermodel/TextRun.java
```diff
@@ -79,6 +80,37 @@ enum FieldType {
     void setFontColor(PaintStyle color);
 
 
+    /**
+     * Returns the font highlight (background) color for this text run.
+     * This returns a {@link SolidPaint}, or null if no highlight is set.
```
pjfanning: You could create a new interface in this package called `HightlightColorSupport` or something similar. Add these methods there instead. HSLFTextRun will not implement HightlightColorSupport. XSLFTextRun will.

DrawTextParagraph can check if the text run is an instanceof HightlightColorSupport.

gffloodg: Ok sure, let me take a look and update.

gffloodg: Code updated, ready for review

poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFTextRun.java
```diff
@@ -158,6 +162,78 @@ private static void fetchFontColor(CTTextCharacterProperties props, Consumer<Pai
...
+     * @param color The highlight (background) color to set.
+     *
+     * @see org.apache.poi.sl.draw.DrawPaint#createSolidPaint(Color)
+     * @since POI 5.2.4
+     */
+    @Override
+    public void setHighlightColor(final PaintStyle color) {
+        if (!(color instanceof SolidPaint)) {
+            LOG.atWarn().log("Currently only SolidPaint is supported!");
+            return;
```
pjfanning, line 226: I'm not sure - can we make this an exception? IllegalArgumentException maybe. Needs to be in the javadoc, one way or another.

poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFTextRun.java
```diff
@@ -158,6 +162,78 @@ private static void fetchFontColor(CTTextCharacterProperties props, Consumer<Pai
...
+     *
+     * @see org.apache.poi.sl.draw.DrawPaint#getPaint(java.awt.Graphics2D, PaintStyle)
+     * @see SolidPaint#getSolidColor()
+     * @since POI 5.2.4
+     */
+    @Override
+    public PaintStyle getHighlightColor() {
+        XSLFShape shape = getParagraph().getParentShape();
+        final boolean hasPlaceholder = shape.getPlaceholder() != null;
+        return fetchCharacterProperty((props, val) -> fetchHighlightColor(props, val, shape, hasPlaceholder));
```
pjfanning: can we rename the `val` variable to something more informative?

gffloodg: done

poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFTextRun.java
gffloodg: Updated and added tests

pjfanning:
Thanks - merged

PR #410, opened 2022-12-24
try to rebuild module-info classes even when java 8 is used

pjfanning:
We release using the Java 8 build but we rely on checked in copies of the module-info.classes. This is error-prone. Someone might have checked in a bad class file or the class file may not have been updated after recent changes to the module-info.java.


pjfanning:
merged with https://github.com/apache/poi/commit/edee60d87a0f3b6d54d056a8f63dd42a386eb593

PR #425, opened 2023-02-04
Catch `Throwable` when using the AWT classes

gastaldi:
`ERROR: java.lang.UnsatisfiedLinkError: no awt in java.library.path` is thrown when attempting to build a native image that depends on AWT on macOS.

See the following issue for more details:

- https://github.com/oracle/graal/issues/4124

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFSheet.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFSheet.java
@@ -63,7 +63,7 @@ public class SXSSFSheet implements Sheet, OoxmlSheetExtensions {
         setRandomAccessWindowSize(randomAccessWindowSize);
         try {
             _autoSizeColumnTracker = new AutoSizeColumnTracker(this);
-        } catch (InternalError e) {
+        } catch (Throwable e) {
             LOG.atWarn().log("Failed to create AutoSizeColumnTracker, possibly due to fonts not being installed in your OS", e);
         }
     }


poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFSheet.java
```diff
@@ -63,7 +63,7 @@ protected SXSSFSheet(SXSSFWorkbook workbook, XSSFSheet xSheet, int randomAccessW
         setRandomAccessWindowSize(randomAccessWindowSize);
         try {
             _autoSizeColumnTracker = new AutoSizeColumnTracker(this);
-        } catch (InternalError e) {
+        } catch (Throwable e) {
```
pjfanning, line 66: this code should not affect poi-scratchpad - but I agree that Throwable is a safer type to catch here

pjfanning:
Can you close this? - I committed https://github.com/apache/poi/commit/040181c42fdfbe15b706e8ae81f5e2f6fe87554c and there are a few more similar cases in the code that should probably be changed

pjfanning:
The catch you modified - that was only added recently and no release has ever included it. Are you sure that InternalError catch was not enough?

gastaldi:
Thanks, I noticed that `SheetUtil` also uses some AWT code in `org.apache.poi.ss.util.SheetUtil.getCellWidth` that could use this try-catch as well

pjfanning:
Can you share the stack traces you are getting? I'd prefer not to be catch Throwables all over the place.

gastaldi:
> The catch you modified - that was only added recently and no release has ever included it. Are you sure that InternalError catch was not enough?

Yes, I am sure. The thrown exception is `java.lang.UnsatisfiedLinkError`. Let me get the full stacktrace

gastaldi:
Here is the stacktrace without the system property set: 

```
Caused by: java.lang.UnsatisfiedLinkError: no awt in java.library.path
	at org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk.NativeLibrarySupport.loadLibraryRelative(NativeLibrarySupport.java:120)
	at java.base@17.0.5/java.lang.ClassLoader.loadLibrary(ClassLoader.java:50)
	at java.base@17.0.5/java.lang.Runtime.loadLibrary0(Runtime.java:818)
	at java.base@17.0.5/java.lang.System.loadLibrary(System.java:1989)
	at java.desktop@17.0.5/java.awt.Toolkit$2.run(Toolkit.java:1388)
	at java.desktop@17.0.5/java.awt.Toolkit$2.run(Toolkit.java:1386)
	at java.base@17.0.5/java.security.AccessController.executePrivileged(AccessController.java:168)
	at java.base@17.0.5/java.security.AccessController.doPrivileged(AccessController.java:318)
	at java.desktop@17.0.5/java.awt.Toolkit.loadLibraries(Toolkit.java:1385)
	at java.desktop@17.0.5/java.awt.Toolkit.initStatic(Toolkit.java:1423)
	at java.desktop@17.0.5/java.awt.Toolkit.<clinit>(Toolkit.java:1397)
	at java.desktop@17.0.5/java.awt.Font.<clinit>(Font.java:288)
	at java.desktop@17.0.5/java.awt.font.TextLayout.singleFont(TextLayout.java:468)
	at java.desktop@17.0.5/java.awt.font.TextLayout.<init>(TextLayout.java:530)
	at org.apache.poi.ss.util.SheetUtil.getDefaultCharWidth(SheetUtil.java:301)
	at org.apache.poi.xssf.streaming.AutoSizeColumnTracker.<init>(AutoSizeColumnTracker.java:117)
	at org.apache.poi.xssf.streaming.SXSSFSheet.<init>(SXSSFSheet.java:95)
	at org.apache.poi.xssf.streaming.SXSSFWorkbook.createAndRegisterSXSSFSheet(SXSSFWorkbook.java:726)
	at org.apache.poi.xssf.streaming.SXSSFWorkbook.createSheet(SXSSFWorkbook.java:720)
```

pjfanning:
Thanks - the exception name is enough for me - so UnsatisfiedLinkError is the issue

gastaldi:
> Thanks - the exception name is enough for me - so UnsatisfiedLinkError is the issue

Yes, as I mentioned in the first comment of this PR 😄

pjfanning:
So I updated SheetUtil in https://github.com/apache/poi/commit/040181c42fdfbe15b706e8ae81f5e2f6fe87554c - is that enough? That code had already been modified to catch UnsatisfiedLinkError (in April 2022).

gastaldi:
This is the stacktrace when the `org.apache.poi.ss.ignoreMissingFontSystem` is set to `true`: 

```java
Caused by: java.lang.NoClassDefFoundError: Could not initialize class java.awt.Font
	at java.desktop@17.0.5/java.awt.font.TextLayout.singleFont(TextLayout.java:468)
	at java.desktop@17.0.5/java.awt.font.TextLayout.<init>(TextLayout.java:530)
	at org.apache.poi.ss.util.SheetUtil.getCellWidth(SheetUtil.java:224)
	at org.apache.poi.ss.util.SheetUtil.getCellWidth(SheetUtil.java:185)
	at org.apache.poi.ss.util.SheetUtil.getColumnWidthForRow(SheetUtil.java:336)
	at org.apache.poi.ss.util.SheetUtil.getColumnWidth(SheetUtil.java:280)
	at org.apache.poi.ss.util.SheetUtil.getColumnWidth(SheetUtil.java:257)
	at org.apache.poi.xssf.streaming.SXSSFSheet.autoSizeColumn(SXSSFSheet.java:1621)
	at org.apache.poi.xssf.streaming.SXSSFSheet.autoSizeColumn(SXSSFSheet.java:1567)
```

gastaldi:
> So I updated SheetUtil in [040181c](https://github.com/apache/poi/commit/040181c42fdfbe15b706e8ae81f5e2f6fe87554c) - is that enough?

I think you also need to add catch any AWT operations in it, like `Font defaultFont = wb.getFontAt( 0);` in the `public static int getDefaultCharWidth(final Workbook wb)` method

pjfanning:
Can you provide a stack trace where getFontAt fails? If that is failing, I think there will be loads of changes needed.

Would it not be easier to install fonts yourself?

gastaldi:
The problem is that AWT is not available in a GraalVM native image built in MacOS because of https://github.com/oracle/graal/issues/4124, so I get `NoClassDefFoundError`s and the `java.lang.UnsatisfiedLinkError` above. I don't get any of these errors when running in JVM mode

xzel23:
Wouldn't it be better to guard awt calls by checking for headless mode instead of catching UnsatisfiedLinkError?

pjfanning:
@xzel23 this issue is hard to test. I ran the SXSSF tests with GraalVM on my Mac and had no issues (with and without this code change).

We have no POI road map for a POI 6 release that would allow us some latitude to refactor some code. Right now, we are sort of stuck with putting sticky plasters on the existing code base. In the end of the day catching Throwable instead of specific Errors means we are less likely to have someone with a different funky setup come along and say they have a different special Error happening. Since Errors are not declared in throws declarations, I think it is risky to depend on catching precise Errors.

For me, SXSSFWorkbook should never have had a hard dependency on java.awt code. The auto column sizing feature should have been something that users should have had to enable explicitly. But introducing new API methods has its own problems. Stackoverflow questions for POI indicate a very large proportion of the user base are using very old versions of the libs.

xzel23:
@pjfanning I see this as rather a GraalVM problem than a POI one. If you couldn't reproduce the problem on your Mac (the GraalVM issue has already been fixed on Windows but not on MacOS), it's probably because you used another GraalVM distribution. AFAIK, Bellsoft's has some support for AWT and this should work.

If OP could confirm this works using Bellsoft's NIK (their GraalVM distribution), I would rather recommend not making these changes in POI.

GraalVM folks seem to be working on fixing this in the original distribution, but so far it's not know what release it might get in.

gastaldi:
I managed to bypass the AWT issues by replacing the methods in the classes that use it.

See https://github.com/quarkiverse/quarkus-poi

pjfanning:
I reverted the bulk of my recent changes with https://github.com/apache/poi/commit/43551babf1cf628aa184a2f9445412815781dc9e - the new code still treats UnsatisfiedLinkError as meaning AWT is not set up - as well as the pre-existing InternalError check.

PR #428, opened 2023-02-10
Performance improvements when writing many cells with SXSSF

gtoison:
SXSSFCell.getColumnIndex() gets slow when writing thousands of columns because we linearly search through all the cells in a row. 
Since the column index is already known when creating a new cell we can just use it to track the left/right most column.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFRow.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFRow.java
@@ -141,7 +141,7 @@ public class SXSSFRow implements Row, Comparable<SXSSFRow>
         checkBounds(column);
         SXSSFCell cell = new SXSSFCell(this, type, column);
         _cells.put(column, cell);
-        _sheet.trackNewCell(cell);
+        _sheet.trackNewCell(column);
         return cell;
     }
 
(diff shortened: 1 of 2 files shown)


pjfanning:
* `new SXSSFCell(this, type, column)` sets _columnIndex 
* getColumIndex() uses _columnIndex  and only searches the row if _columnIndex is not set

ok - so this appears to be possibly new - added since POI 5.2.3 - so we may have fixed this issue but release of the fix is not done yet

-> see https://github.com/apache/poi/commit/7454a722f0d408afc605cb1b3cbb2daa9c5c913c

If you want to try this code out, you could try building POI yourself from latest sources or the jars from a recent build are at https://ci-builds.apache.org/job/POI/job/POI-DSL-1.8/834/artifact/

gtoison:
Ah that's great! Indeed I noticed the performance problem using 5.2.3 but did not see that some improvements were made
Thanks to the changes I suppose that this PR is not needed.
Unless you want to merge it anyway I will close the PR, let me know

pjfanning:
I don't to want to merge this - can you close it? I think the existing changes are sufficient.

gtoison:
I've tested the snapshot build and it seems better indeed, looking forward to 5.2.4!
Thank you!

PR #443, opened 2023-03-17
[Bug-66532] Improve performance of SheetDataWriter

rascmatt:
Simplify loop and avoid code point to string conversions.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SheetDataWriter.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SheetDataWriter.java
@@ -31,6 +31,7 @@ import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.util.Iterator;
+import java.util.PrimitiveIterator;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -397,33 +398,34 @@ public class SheetDataWriter implements Closeable {
             return;
         }
 
-        for (int i = 0; i < s.length(); i++) {
-            final int codepoint = s.codePointAt(i);
+        final PrimitiveIterator.OfInt iter = s.codePoints().iterator();
+        while(iter.hasNext()) {
+            final int codepoint = iter.nextInt();
 
             switch (codepoint) {
-                case 60: // <
+                case '<':
                     _out.write("&lt;");
                     break;
-                case 62: // >
+                case '>':
                     _out.write("&gt;");
                     break;
-                case 38: // &
+                case '&':
                     _out.write("&amp;");
                     break;
-                case 34: // "
+                case '"':
                     _out.write("&quot;");
                     break;
                 // Special characters
-                case 10: // \n
+                case '\n':
                     _out.write("&#xa;");
                     break;
-                case 13: // \r
+                case '\r':
                     _out.write("&#xd;");
                     break;
-                case 9:  // \t
+                case '\t':
                     _out.write("&#x9;");
                     break;
-                case 160: // NO-BREAK SPACE
+                case '\u00A0': // NO-BREAK SPACE
                     _out.write("&#xa0;");
                     break;
                 default:
@@ -438,7 +440,6 @@ public class SheetDataWriter implements Closeable {
                             _out.write(c);
                         }
                     } else if (chars.length == 2){
-                        i++;
                         _out.write(chars);
                     } else {
                         throw new IllegalArgumentException("Not a valid code point: " + codepoint);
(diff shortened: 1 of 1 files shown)


rascmatt:
Following you will find the results of a JMH benchmark I ran with 3 different versions. Unfortunately I was unable to get JMH running with Gradle, so I ran this in an external Maven Project.
 
The latest released version (5.2.3): 'Main.benchOriginal'
An unreleased improvement (#405): 'Main.bench_405'
And the version proposed in this commit: 'Main.bench_66532'

Benchmark            Mode  Cnt        Score        Error  Units
Main.benchOriginal  thrpt   15   520433.760 ±  81525.743  ops/s
Main.bench_405      thrpt   15   579381.912 ±  30395.196  ops/s
Main.bench_66532    thrpt   15  3029062.360 ± 141908.012  ops/s

poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SheetDataWriter.java
```diff
@@ -397,46 +397,51 @@ protected void outputEscapedString(String s) throws IOException {
             return;
         }
 
-        for (Iterator<String> iter = CodepointsUtil.iteratorFor(s); iter.hasNext(); ) {
-            String codepoint = iter.next();
+        for (int i = 0; i < s.length(); i++) {
+            final int codepoint = s.codePointAt(i);
```
pjfanning: this code is wrong - s.length() is not the number of codepoints - it is the number of chars - in a string.

see https://www.w3resource.com/java-tutorial/string/string_codepointcount.php

pjfanning (changes requested):

poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SheetDataWriter.java
rascmatt: I might need to improve the readability of the loop, however I think this is covered by line 441, where (in case of a character pair) the counter is additionally increased.

pjfanning: loops should not increment in multiple places (ideally) - if you use the codePointCount for the limit of the loop then the count will be correct

poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SheetDataWriter.java
```diff
@@ -397,46 +397,51 @@ protected void outputEscapedString(String s) throws IOException {
...
+
             switch (codepoint) {
-                case "<":
+                case 60: // <
```
pjfanning: could you test with `'<'`? - this should have similar performance but would be more readable

pjfanning:
> Following you will find the results of a JMH benchmark I ran with 3 different versions. Unfortunately I was unable to get JMH running with Gradle, so I ran this in an external Maven Project.
> 
> The latest released version (5.2.3): 'Main.benchOriginal' An unreleased improvement (#405): 'Main.bench_405' And the version proposed in this commit: 'Main.bench_66532'
> 
> Benchmark Mode Cnt Score Error Units Main.benchOriginal thrpt 15 520433.760 ± 81525.743 ops/s Main.bench_405 thrpt 15 579381.912 ± 30395.196 ops/s Main.bench_66532 thrpt 15 3029062.360 ± 141908.012 ops/s

could you share your benchmark code? a GitHub project or some such

poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SheetDataWriter.java
rascmatt: Yes, but then the codePointAt() method will not receive the correct index. I switched to using the codePoints iterator, which does not seem to have a big performance impact.

poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SheetDataWriter.java
rascmatt: Good idea. Done.

rascmatt:
https://github.com/rascmatt/poi-benchmark
The (updated) benchmark with code and results.

pjfanning:
thanks - looks good. Let me look over this but it looks like this should be safe to merge over the coming days.

pjfanning:
I added https://github.com/apache/poi/commit/0275daa5deae2e0069badd1f46268abb43fbc3dc - not exactly what is in your PR. Could you try this out in your benchmark?

rascmatt:
Thank you, I added this version to the benchmark and it looks good. In addition I added a benchmark for a more integrated setting (https://github.com/rascmatt/poi-benchmark/commit/d7fa4a8b2842d3a80306b537407863ad58ceb645). It shows a speedup of about ~45% for the SheetDataWriter.

PR #462, opened 2023-05-03
Update the number of property blocks in the POIFS header block

ebourg:
This fixes https://bz.apache.org/bugzilla/show_bug.cgi?id=66590

### DIFF
--- a/poi/src/main/java/org/apache/poi/poifs/property/PropertyTable.java
+++ b/poi/src/main/java/org/apache/poi/poifs/property/PropertyTable.java
@@ -219,6 +219,9 @@ public final class PropertyTable implements BATManaged {
        if(getStartBlock() != stream.getStartBlock()) {
           setStartBlock(stream.getStartBlock());
        }
+
+       // Update the number of property blocks in the header
+       _header_block.setPropertyCount(countBlocks());
     }
 
     private void populatePropertyTree(DirectoryProperty root) throws IOException {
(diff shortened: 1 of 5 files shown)


centic9:
Thanks for the changes. Looks good on a brief look. 

Can you add a sample .msi-file to directory test-data/poi-fs and a small unit-test which verifies opening it so we verify the intended feature as well?

ebourg:
I did modify the existing tests to ensure the header is properly updated. I don't think adding an actual msi file will improve the test coverage, it's just a CFB file like any other one.

pjfanning:
I don't think we should merge this without an MSI test. We have zero test coverage of MSIs.

-1 from me

ebourg:
I can contribute one for the tests (I've built one for the Jsign tests: https://github.com/ebourg/jsign/blob/master/jsign-core/src/test/resources/minimal.msi), but a msi file is really not different from other formats handled by POIFS, and this issue is mostly related to the correctness of the header information, that's not MSI specific.

ebourg:
@pjfanning Any chance you could withdraw your -1 without a MSI test file? As I stated, I detected this issue with MSI files but this is really not MSI specific, all CFB/COM/OLE files are expected to have the number of directory sectors in their header to be spec compliant.

pjfanning:
why not contribute a test?

ebourg:
The pull request does contain proper test coverage, it expands the existing tests checking the header sector (`TestPOIFSStream.createWriteRead()`, `TestPOIFSStream.addBeforeWrite()` and `TestHeaderBlockReading.testConstructors()`) with extra assertions for the directory sector count field. I could add a MSI specific test but that would just duplicate the existing tests.

pjfanning:
Ok - this is merged now - will review the build later to see if anything fails

ebourg:
Thank you!

PR #473, opened 2023-06-14
add getCTSettings method in XWPFSettings

ningpp:
characterSpacingControl(compressPunctuation) and other is very useful


pjfanning:
Seems useful. Merged with https://github.com/apache/poi/commit/6943cc9a9c8528ca47bba2a499e408e41f0c893e

PR #485, opened 2023-07-08
A fix for bug 66682.  

opeongo:
This fixes the problem where values from boolean formula cells in binary excel files read with the event model are not handled.  These cells are always yield empty values regardless of the evaluation of the formula.  The simple fix is to use the same cell handler as for literal boolean values.

Tested and it works.  I will add a test case if someone can give me some direction as to where it should go.


pjfanning:
https://github.com/apache/poi/blob/trunk/poi-ooxml/src/test/java/org/apache/poi/xssf/binary/TestXSSFBSheetHyperlinkManager.java is a good test to copy.

If you want to read in an xlsb file, copy it in `test-data/spreadsheet`. That's where TestXSSFBSheetHyperlinkManager is reading its input file from.

PR #487, opened 2023-07-21
Round up seconds in CellElapsedFormatter

aschott-looker:
Seconds should round up in the elapsed formatter. If you have a value like `0.0001` and format it with `[h]\h m\m s\s`, you get "0h 0m 8s" whereas Excel outputs "0h 0m 9s".


pjfanning:
I reverted this after it broke one of the existing tests associated with the xlsx file.

aschott-looker:
@pjfanning - Sorry about that! There is something odd with that file. If you download it, make a trivial change (e.g. add some text somewhere), save it and run the test, it fails.

I've opened https://github.com/apache/poi/pull/489 instead which has the same code change but just adding a unit test elsewhere to verify the behavior.

PR #489, opened 2023-07-24
Round up seconds in CellElapsedFormatter

aschott-looker:
Seconds should round up in the elapsed formatter. If you have a value like 0.0001 and format it with [h]\h m\m s\s, you get "0h 0m 8s" whereas Excel outputs "0h 0m 9s".

### DIFF
--- a/poi/src/test/java/org/apache/poi/ss/format/TestCellFormat.java
+++ b/poi/src/test/java/org/apache/poi/ss/format/TestCellFormat.java
@@ -1046,4 +1046,11 @@ class TestCellFormat {
                 .map(CellFormatPart.NAMED_COLORS::get)
                 .forEach(Assertions::assertNotNull);
     }
+
+    @Test
+    void testElapsedSecondsRound() {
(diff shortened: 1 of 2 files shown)


poi/src/test/java/org/apache/poi/ss/format/TestCellFormat.java
```diff
@@ -1046,4 +1046,11 @@ void testNamedColors() {
                 .map(CellFormatPart.NAMED_COLORS::get)
                 .forEach(Assertions::assertNotNull);
     }
+
+    @Test
+    void testElapsedSecondsRound() {
```
pjfanning, line 1051: * this looks like a better way to test this
* the tests where we read xlsx files and match their results can be easily broken and the PRs are hard to review because you can't see the diff in the xlsx files

aschott-looker:
Closed by https://github.com/apache/poi/commit/c6aa3a1318271b25fe90a34fca3d6f1de0a813fc

PR #496, opened 2023-08-05
Bug 66855 Formula parser incorrectly handles sheet name containing multiple single quotes in a row

kleb-prog:
Small fix for the bug

https://bz.apache.org/bugzilla/show_bug.cgi?id=66855


poi/src/main/java/org/apache/poi/ss/formula/FormulaParser.java
```diff
@@ -1182,8 +1182,6 @@ private SheetIdentifier parseSheetName(boolean isSndPartOfQuotedRange) {
                         if (look == '\''){
                             // Any single quotes which were already present in the sheet name will be converted to double single quotes ('')
                             // so switch back to single quote
-                            sb.appendCodePoint(look);
-                            nextChar();
```
kleb-prog, line 1186: It's not necessary to do it here because on this step variable "look" has second single quote of the pair of single quotes and will be appended to string builder on next iteration of the loop

poi/src/test/java/org/apache/poi/hssf/model/TestFormulaParser.java
```diff
@@ -1357,8 +1357,8 @@ void testUnionOfFullCollFullRowRef() throws IOException {
                 UnionPtg.class
         );
 
-        wb.createSheet("Sh't1");
-        ptgs = parse("'Sh''t1'!$A:$A,'Sh''t1'!$1:$4", wb);
```
pjfanning: could you add a new test and leave this one as is? It must be useful to keep the test for `"Sh't1"`.

kleb-prog: Yes, I'll do it

kleb-prog: Done

pjfanning:
thanks - merged with https://github.com/apache/poi/commit/d15286103648b5cbfadfc6eeec132f224b49e2ec

PR #503, opened 2023-08-17
Fully replace content of XWPFTableCell on `setText`

AntonOellerer:
Up until now, invoking `XWPFTableCell#setText(String)` appended the string to a paragraph if one already existed in the cell, contrary to what would be expected.
This is fixed in this commit by first removing all runs and iruns from the paragraph in the cell.

Fixes https://bz.apache.org/bugzilla/show_bug.cgi?id=66988


poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFTableCell.java
```diff
@@ -423,6 +423,8 @@ public String getText() {
 
     public void setText(String text) {
         XWPFParagraph par = paragraphs.isEmpty() ? addParagraph() : paragraphs.get(0);
+        par.getRuns().clear();
+        par.getIRuns().clear();
```
pjfanning: Could you add a new `appendText` method that works the old way? Add javadoc to that method saying `@since POI 5.2.4` and javadoc to this method describing the change in behaviour in POI 5.2.4. 

Could you add a unit test for the appendText method?

AntonOellerer: Done, testing also exposed an issue in how the `setText` was done, which has been fixed

pjfanning:
This broke an existing test called test63624 - I'll rewrite that test

PR #514, opened 2023-09-01
[bug-63934] Fix parsing of structure references

rascmatt:


pjfanning:
https://bz.apache.org/bugzilla/show_bug.cgi?id=63934

PR #523, opened 2023-09-21
fix test locale

avl93:
Some tests rely on system locale being US and fails on other locales, this fix makes them run successfully on any locale.


poi/src/test/java/org/apache/poi/ss/formula/functions/TestDays.java
```diff
@@ -37,6 +39,11 @@ Licensed to the Apache Software Foundation (ASF) under one or more
  */
 public class TestDays {
 
+    @BeforeAll
+    static void setup() {
+        Locale.setDefault(Locale.US);
```
pjfanning: POI has its own LocaleUtil. Could you use that instead?

avl93: Sure, done

PR #536, opened 2023-10-14
Complex scripts support in XWPFRun

moalhaddar:
Bugzilla ticket: https://bz.apache.org/bugzilla/show_bug.cgi?id=67735

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
@@ -75,7 +75,6 @@ public class XWPFRun implements ISDTContents, IRunElement, CharacterRun {
     private final String pictureText;
     private final IRunBody parent;
     private final List<XWPFPicture> pictures;
-    private ScriptType scriptType = ScriptType.NON_COMPLEX;
 
     /**
      * @param r the CTR bean which holds the run attributes
@@ -140,28 +139,6 @@ public class XWPFRun implements ISDTContents, IRunElement, CharacterRun {
     @Deprecated
     public XWPFRun(CTR r, XWPFParagraph p) {
         this(r, (IRunBody) p);
-    }
-
-    /**
-     * Gets the current script type for this run.
-     *
-     * @return the current script type.
-     */
-    public ScriptType getScriptType() {
-        return this.scriptType;
-    }
-
-    /**
-     * Sets the script type for this run.
-     *
-     * @param scriptType to be set.
-     */
-    public void setScriptType(ScriptType scriptType){
-        this.scriptType = scriptType;
-    }
-
-    private boolean isComplexScript() {
-        return getScriptType() == ScriptType.COMPLEX;
     }
 
 
@@ -273,26 +250,31 @@ public class XWPFRun implements ISDTContents, IRunElement, CharacterRun {
     }
 
     /**
-     * Whether the bold property shall be applied to the currently specified
-     * {@link #scriptType script type} characters in the contents of this
-     * run when displayed in a document.
+     * Whether the bold property shall be applied to the non-complex
+     * characters in the contents of this run when displayed in a document.
      *
-     * @return {@code true} if the bold property is applied
+     * @return {@code true} if the bold property non-complex scripts is applied
      */
     @Override
     public boolean isBold() {
         CTRPr pr = getRunProperties(false);
-        if (isComplexScript()) {
-            return pr != null && pr.sizeOfBCsArray() > 0 && isCTOnOff(pr.getBCsArray(0));
-        } else {
-            return pr != null && pr.sizeOfBArray() > 0 && isCTOnOff(pr.getBArray(0));
-        }
+        return pr != null && pr.sizeOfBArray() > 0 && isCTOnOff(pr.getBArray(0));
     }
 
     /**
-     * Whether the bold property shall be applied to the currently specified
-     * {@link #scriptType script type} characters in the contents of this
-     * run when displayed in a document.
+     * Whether the bold property shall be applied to the complex
+     * characters in the contents of this run when displayed in a document.
+     *
+     * @return {@code true} if the bold property for complex scripts is applied
+     */
+    public boolean isComplexScriptBold() {
+        CTRPr pr = getRunProperties(false);
+        return pr != null && pr.sizeOfBCsArray() > 0 && isCTOnOff(pr.getBCsArray(0));
+    }
+
+    /**
+     * Whether the bold property shall be applied to the non-complex
+     * characters in the contents of this run when displayed in a document.
      * <p>
      * This formatting property is a toggle property, which specifies that its
      * behavior differs between its use within a style definition and its use as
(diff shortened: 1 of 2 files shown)


poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
```diff
@@ -1447,6 +1566,19 @@ public enum FontCharRange {
         hAnsi /* high ansi */
     }
 
+    /**
+     * @see <a href="https://www.ecma-international.org/publications-and-standards/standards/ecma-376/">OOXML Spec</a>
```
pjfanning: Can you pull out the text? Or tell us exactly where to find these sections. The ECMA specs are an insane mess with zips inside zips. 4 separate parts.

This enum should be its own class but the question is whether it is docx specific or needs to appear in a more general package.

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
```diff
@@ -905,10 +1006,28 @@ public void setFontSize(int size) {
...
+    public void setFontSize(double size, ScriptType scriptType) {
         CTRPr pr = getRunProperties(true);
-        CTHpsMeasure ctSize = pr.sizeOfSzArray() > 0 ? pr.getSzArray(0) : pr.addNewSz();
-        ctSize.setVal(bd.multiply(BigDecimal.valueOf(2)).setScale(0, RoundingMode.HALF_UP).toBigInteger());
+        BigDecimal bd = BigDecimal.valueOf(size);
+        BigInteger val = bd.multiply(BigDecimal.valueOf(2)).setScale(0, RoundingMode.HALF_UP).toBigInteger();
+        boolean isComplexScript = scriptType == ScriptType.COMPLEX;
+
+        if (isComplexScript) {
+            CTHpsMeasure ctCsSize = pr.sizeOfSzCsArray() > 0 ? pr.getSzCsArray(0) : pr.addNewSzCs();
```
pjfanning: what are you going to do if the array already exists and it has more than 1 array element?

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
```diff
@@ -256,8 +256,23 @@ public void setLang(String lang) {
      */
     @Override
     public boolean isBold() {
+        return isBold(ScriptType.NON_COMPLEX);
```
pjfanning: Is there any chance that the the run itself could be marked as having ScriptType.COMPLEX? It feels very messy for the API caller to have to specify this on all the methods like isBold, isItalic, etc.

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
moalhaddar: The sections doesn't mention what a complex script is because it's something related to the typesetting. However, these sections are titled:
- 17.3.2.2 bCs (Complex Script Bold)
- 17.3.2.17 iCs (Complex Script Italics)
- 17.3.2.38 sz (Non-Complex Script Font Size)
- 17.3.2.39 szCs (Complex Script Font Size)

and in them they specify that these properties must have an explicit tag with it's value for them to work with complex scripts. I'm not sure if i should include the entire paragraphs or only a link to the PDF zip file? (it's 1 file only).

https://www.ecma-international.org/wp-content/uploads/ECMA-376-1_5th_edition_december_2016.zip

What would you suggest?

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
moalhaddar: The current behavior of POI does exactly this for the regular fonts. So i repeated the implementation with the complex scripts.

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
moalhaddar: I see. You mean as in an internal state for the script type with a setter and a getter? I think i can do that

pjfanning:
I thoroughly dislike the M$ formats.

Could I suggest this?
* leave the existing methods as is - no changes
* remove the ScriptType enum
* create separate methods for the complex script settings, eg for 'bold'
  * setComplexScriptBold(boolean)
  * isComplexScriptBold()
* similar separate methods for italic and font size

pjfanning:
as an aside, would you have any interest in trying to fix https://bz.apache.org/bugzilla/show_bug.cgi?id=63741 ?

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
```diff
@@ -911,6 +1047,29 @@ public void setFontSize(double size) {
...
+     * <p>
+     * If this element is not present, the default value is to leave the value
+     * applied at previous level in the style hierarchy. If this element is
+     * never applied in the style hierarchy, then any appropriate font size may
+     * be used for the non-complex characters.
+     * </p>
+     *
+     * @param size The font size as number of point measurements.
+     * @see #setFontSize(int)
+     * @since POI 5.0.0
```
pjfanning: Could you use `@since POI 5.2.5` on all the new public methods? Including any new public methods that have no `@since` setting.

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
```diff
@@ -849,7 +929,22 @@ public int getFontSize() {
...
-     * Specifies the font size which shall be applied to all non complex script
+     * Specifies the font size which shall be applied to the complex script
+     * characters in the contents of this run when displayed.
+     *
+     * @return value representing the font size (non-integer size will be rounded with half rounding up,
+     * -1 is returned if size not set)
+     * @deprecated use {@link #getComplexScriptFontSizeAsDouble()}
+     */
+    @Deprecated
+    @Removal(version = "6.0.0")
```
pjfanning: for the new code only, could we could just `public double getComplexScriptFontSize()`? No `getComplexScriptFontSizeAsDouble` needed - let's not add new methods that have deprecated or removal annotations on them.

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
moalhaddar: Done

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
moalhaddar: Done

pjfanning:
Thanks - merged with https://github.com/apache/poi/commit/bc870ab76497d3f968a32282104fb170bea7a77f

PR #541, opened 2023-11-02
Bug: Auto sizing not working properly due to rounding error when calculating defaultCharWidth

nme-xx:
The XSSFSheet#autoSizeColumn method doesn't work properly for some specific cell values and font metrics.

The following code snippet is an example for this behavior:
```java
String sheetName = "Sheet 1";
String cellText = "Auto sizing incorrect! :(";

try (XSSFWorkbook workbook = new XSSFWorkbook()) {

    final XSSFSheet sheet = workbook.createSheet(sheetName);

    final XSSFCell cell = sheet.createRow(0).createCell(0);

    cell.setCellValue(cellText);

    final XSSFFont font = workbook.createFont();
    font.setFontName("Arial");
    font.setFontHeight(15);
    font.setBold(true);

    final XSSFCellStyle style = workbook.createCellStyle();
    style.setFont(font);
    cell.setCellStyle(style);

    sheet.autoSizeColumn(0, false);

    try (FileOutputStream fOS = new FileOutputStream("./test-excel.xlsx")) {
        workbook.write(fOS);
    }
}
```

Generated excel file:
image

The bug is caused by a rounding error in SheetUtil#getDefaultCharWidth:
https://github.com/apache/poi/blob/5533d35cefd15e98fd988a458ac41c5e8aad06f7/poi/src/main/java/org/apache/poi/ss/util/SheetUtil.java#L301-L302

Rounding is not necessary at all, removing the operation and converting the return type of getDefaultCharWidth() to double solves the issue.

The column width calculation that causes the error (and uses a value returned by SheetUtil#getDefaultCharWidth) occurs in SheetUtil#getCellWidth:
https://github.com/apache/poi/blob/5533d35cefd15e98fd988a458ac41c5e8aad06f7/poi/src/main/java/org/apache/poi/ss/util/SheetUtil.java#L244

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/AutoSizeColumnTracker.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/AutoSizeColumnTracker.java
@@ -47,7 +47,7 @@ import org.apache.poi.util.Internal;
 */
 @Internal
 /*package*/ class AutoSizeColumnTracker {
-    private final int defaultCharWidth;
+    private final double defaultCharWidth;
     private final DataFormatter dataFormatter = new DataFormatter();
 
     // map of tracked columns, with values containing the best-fit width for the column
--- a/poi/src/main/java/org/apache/poi/ss/util/SheetUtil.java
+++ b/poi/src/main/java/org/apache/poi/ss/util/SheetUtil.java
@@ -118,7 +118,7 @@ public class SheetUtil {
      * @param useMergedCells    whether to use merged cells
      * @return  the width in pixels or -1 if cell is empty
      */
-    public static double getCellWidth(Cell cell, int defaultCharWidth, DataFormatter formatter, boolean useMergedCells) {
+    public static double getCellWidth(Cell cell, double defaultCharWidth, DataFormatter formatter, boolean useMergedCells) {
         List<CellRangeAddress> mergedRegions = cell.getSheet().getMergedRegions();
         return getCellWidth(cell, defaultCharWidth, formatter, useMergedCells, mergedRegions);
     }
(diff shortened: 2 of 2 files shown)


pjfanning:
I added some changes in https://github.com/apache/poi/commit/8819952b2f7210d4201a1b1d75d420abc67d8c2c 

I didn't want to break backward compatibility so I've added new methods and deprecated the ones that they replace.

It's hard to automate tests because the calculated widths depend on the font setup that testers have installed.

If you can spot a mistake or could try out my change that would be great. I will double check the code change later and it is possible it might break a few tests.

nme-xx:
I can't see any additional failing tests (except the ones that fail due to my locale but aren't related to the issue).
My example seems to work too:
image
Do you know when the next version will be released yet?

pjfanning:
It will be a few weeks before POI 5.2.5 is released. We are concentrating on releasing XMLBeans 5.2.0 first and once that is released, we should be in a position to start a POI release vote.

nme-xx:
Alright, thank you for addressing the issue so quickly!

pjfanning:
@NiklasMelznerExxcellent Some users are complaining that this change is still creating cells that are not wide enough. Would you have any interest in having another look?

I'm wondering whether it might be worth just arbitrarily increasing the width of each cell by some hardcoded width. Maybe we could make this configurable but default to some non-zero value.

niklasmelzner:
I guess you're referring to [this bug](https://bz.apache.org/bugzilla/show_bug.cgi?id=68094)?
I'll have a look at it next week.

ThoSap:
Was this fixed in Apache POI 5.3.0 which was released two days ago?

centic9:
Changes from this PR were applied as part of Apache POI 5.3.0, see [r1913676](https://svn.apache.org/viewvc?view=rev&rev=1913676), however some manual tests showed that the actual result is dependent on the application used to display the workbook, so columns may still be either too small, see https://bz.apache.org/bugzilla/show_bug.cgi?id=68094 for some details.

ThoSap:
With Apache POI 5.2.3 the auto sizing works properly with Excel, I'm using that version currently as newer ones (did not try 5.3.0 yet) are broken.
I don't know about LibreOffice though.

pjfanning:
https://svn.apache.org/viewvc?view=rev&rev=1913676 was in POI 5.2.5. So far the only suggested solution is to have a configurable extra width that is added to all cell widths. There are too many variables to get a works in all cases solution (depending on what fonts are installed on the machine used to run the POI code and which spreadsheet tool is used to view the generated spreadsheet).

pjfanning:
I created #657 as an outline

PR #552, opened 2023-11-23
Fix Bug 68200 

Tong-Huang:
Hi Team,

Here some cases, `calculateRate` will return `Double.NaN` -> [#NUM!],
(https://github.com/apache/poi/blob/trunk/poi/src/main/java/org/apache/poi/ss/formula/functions/Rate.java#L99)

But it can get the right result in Numbers

case1:

RATE(22,30000,20000,-82257625,0,0.1)

Result: Double.NaN

whereas in Numbers I get: 0.35397960290713076

--

case2:

RATE(22,10000,10000,-313562750,0,0.1)

Result: Double.NaN

whereas in Numbers I get: 0.35397960290713076

Following the implementation of numpy-financial, I try to fix the bug.

https://github.com/numpy/numpy-financial/blob/d02edfb65dcdf23bd571c2cded7fcd4a0528c6af/numpy_financial/_financial.py#L582

### DIFF
--- a/poi/src/main/java/org/apache/poi/ss/formula/functions/Rate.java
+++ b/poi/src/main/java/org/apache/poi/ss/formula/functions/Rate.java
@@ -74,11 +74,14 @@ public class Rate implements Function {
     }
 
     private static double _g_div_gp(double r, double n, double p, double x, double y, double w) {
-        double t1 = Math.pow(r+1, n);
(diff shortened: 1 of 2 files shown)


xzel23:
Link to Apache POI Bug Tracker: https://bz.apache.org/bugzilla/show_bug.cgi?id=68200

PR #581, opened 2024-02-01
Updated TLS to v1.2 for TimeStampSimpleHttpClient

wreicherz:
simple update of chosen algo. ("TLS"); vs. ("TLSv1.2");


pjfanning:
Are you making this change just because it looks wrong or because you have hit an issue?

With POI, we work with real use cases because the Microsoft specs can be hard to interpret.

pjfanning:
the logic you are changing only appears to happens when ignoreHttpsCertificates = true - are you changing to this true (it defaults to false)

Ideally you should not ignore the HTTPS certs

wreicherz:
> Are you making this change just because it looks wrong or because you have hit an issue?

To prevent incompatibility in the future, due to the deprecated state of the algo

No current issues whatsoever, (very) low priority.

PR #582, opened 2024-02-03
Increase performance of CodepointsUtil.java by 5x

mirraj2:
This code path gets invoked millions of times for my pretty basic excel file b/c it is called for every single cell.

I guess java's native streams are pretty slow, because this sped things up by a LOT

### DIFF
--- a/poi/src/main/java/org/apache/poi/util/CodepointsUtil.java
+++ b/poi/src/main/java/org/apache/poi/util/CodepointsUtil.java
@@ -28,7 +28,7 @@ public class CodepointsUtil {
      * @return iterator with Strings representing the codepoints
      */
   public static Iterator<String> iteratorFor(String text) {
-    PrimitiveIterator.OfInt iter = text.codePoints().iterator();
+    PrimitiveIterator.OfInt iter = primitiveIterator(text);
     return new Iterator<String>() {
       @Override
       public boolean hasNext() {
--- a/poi/src/main/java/org/apache/poi/util/CodepointsUtil.java
+++ b/poi/src/main/java/org/apache/poi/util/CodepointsUtil.java
@@ -26,13 +26,21 @@ public class CodepointsUtil {
     /**
      * @param text to iterate over
      * @return iterator with Strings representing the codepoints
-     * @see #primitiveIterator(String) a more performant iterator
      */
-    public static Iterator<String> iteratorFor(String text) {
-        return text.codePoints()
-                .mapToObj(codePoint -> new String(Character.toChars(codePoint)))
-                .iterator();
-    }
+  public static Iterator<String> iteratorFor(String text) {
+    PrimitiveIterator.OfInt iter = text.codePoints().iterator();
+    return new Iterator<String>() {
+      @Override
+      public boolean hasNext() {
+        return iter.hasNext();
+      }
+
+      @Override
+      public String next() {
+        return Character.toString(iter.next());
+      }
+    };
+  }
 
     /**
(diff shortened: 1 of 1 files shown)


xzel23:
I think another (and maybe the main) reason might be the two step conversion and the creating of both a new character array and a new string instance creation in every step in the old code whereas in your code, you use Character.toString(int codepoint).  Could you try out what happens in the old code if you replace `mapToObj(codePoint -> new String(Character.toChars(codePoint)))` with `mapToObj(Character::toString)`?

However, that doesn't help much since the method you used was introduced in Java 11 whereas POI 4.x still is compatible with Java 8.

So this cannot be applied at the moment. I have been inactive for quite some time on POI, but I think it would be about time to switch to (at least) Java 11 for POI 5 so this would be possible.

Please change your code to this and report back how this compares to the old version:
```
  public static Iterator<String> iteratorFor(String text) {
    PrimitiveIterator.OfInt iter = primitiveIterator(text);
    return new Iterator<String>() {
      @Override
      public boolean hasNext() {
        return iter.hasNext();
      }

      @Override
      public String next() {
        return new String(Character.toChars(codePoint));
      }
    };
```

pjfanning:
Do you have you know what code is calling this iteratorFor method? Most POI code already uses the primitiveIterator method instead.

xzel23:
Another thing is that the new code iterates over chars, not the codepoints. That does not work for some CLK characters, so you need to revert the second commit.

poi/src/main/java/org/apache/poi/util/CodepointsUtil.java
```diff
@@ -26,13 +26,21 @@ public class CodepointsUtil {
...
+    PrimitiveIterator.OfInt iter = primitiveIterator(text);
+    return new Iterator<String>() {
+      @Override
+      public boolean hasNext() {
+        return iter.hasNext();
+      }
+
+      @Override
+      public String next() {
+        return Character.toString(iter.next());
```
pjfanning, line 40: was this method not added after Java 8? We require Java 8 support.

pjfanning: see https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Character.html#toString(int)

only added in Java 11

pjfanning:
https://github.com/apache/poi/commit/e45579801975c995bd846f41f9318cc11b56ffb3 removes 1 of the 2 remaining uses of  iteratorFor.

xzel23:
This demonstrates the problem:
```
jshell> "𠜎𠜱𠝹𠱓𠱸𠲖𠳏!".toCharArray()
$8 ==> char[15] { '?', '?', '?', '?', '?', '?', '?', '?', '?', '?', '?', '?', '?', '?', '!' }

jshell> "𠜎𠜱𠝹𠱓𠱸𠲖𠳏!".codePoints().toArray()
$9 ==> int[8] { 132878, 132913, 132985, 134227, 134264, 134294, 134351, 33 }
```
Iterating over characters as you do will create a stream consisting of 15 Strings whereas the correct result should have only 8.

xzel23:
Sorry, I didn't see that the primitiveIterator already handles codepoints correctly. I removed the comments regarding codepoints. So you just need to change the Character.toString().

pjfanning:
I added this with https://github.com/apache/poi/commit/2469b89ee62f1d8647d0e2a445f81ce347ced3f5 - thanks

mirraj2:
thanks @pjfanning

PR #586, opened 2024-02-08
Bug 68183: SXSSFWorkbook should dispose of temporary files when close() is called

InfiniteLoop90:
This way consumers can use the workbook in a try-with-resources statement and the temp files will always be cleaned up without having to remember to explicitly call SXSSFWorkbook#dispose().

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFWorkbook.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/SXSSFWorkbook.java
@@ -100,6 +100,10 @@ import org.apache.poi.xssf.usermodel.XSSFWorkbook;
  *
  * Carefully review your memory budget and compatibility needs before deciding
  * whether to enable shared strings or not.
+ *
+ * <p>To release resources used by this workbook (including disposing of the temporary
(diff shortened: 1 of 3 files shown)


InfiniteLoop90:
Let me know if there’s anything else needed from me. Thanks!

centic9:
This seems to break some test-cases, can you try to adjust them?

pjfanning:
* we can't change this behaviour unless the release is a major release
    * https://github.com/apache/poi/commit/eebb3717e04341aa041be14f8322ac4dd835c286 and related changes would also probably mean that the next release can't just be patch release (5.2.6) 
* `dispose` should be deprecated

InfiniteLoop90:
Thanks for the feedback.

I made the following changes:

* Changed the `TestSXSSFWorkbookWithCustomZipEntrySource#validateTempFilesAreEncrypted` test so that it now passes.
* Added a new test to assert that temp files are disposed of when closing the SXSSFWorkbook without needing to call `SXSSFWorkbook#dispose`.
* Deprecated `SXSSFWorkbook#dispose`.

Ready for re-review when ready. Thanks again.

pjfanning:
thanks - merged

PR #590, opened 2024-02-16
Unnecessary Object creation when logging InternalSheet.java

rh-id:


poi/src/main/java/org/apache/poi/hssf/model/InternalSheet.java
```diff
@@ -541,12 +540,11 @@ public ConditionalFormattingTable getConditionalFormattingTable() {
     public void setDimensions(int firstrow, short firstcol, int lastrow, short lastcol)
     {
         LOGGER.atDebug().log("Sheet.setDimensions");
-        LOGGER.atDebug().log(() -> new SimpleMessage(
```
pjfanning, line 544: -1 from me. This lambda function only runs when debug is enabled. Debug is not normally enabled. 

With your change the expensive String concat will always run, regardless of whether the text needs to be output.

xzel23: Correct. And while we are at it: since we are at Log4J2 now, I think we should instead better use something like this:
```
        LOGGER.debug("Sheet.setDimensions");
        LOGGER.debug("firstrow{}firstcol{}lastrow{}lastcol{}", firstrow, firstcol, lastrow, lastcol);

```

So, get rid of the `atDebug()`where not necessary and use the MessageFormat.

pjfanning: if we do change this line, I suggest that we improve the format of its output - eg 
```
LOGGER.debug("firstrow: {} firstcol: {} lastrow: {} lastcol: {}", firstrow, firstcol, lastrow, lastcol);
```

rh-id:
Hi @pjfanning @xzel23 ,
Thanks for the review, I had updated the commit.

xzel23:
@pjfanning Looks good to me. What do you think?

pjfanning:
Thanks @rh-id, @xzel23 - merged

PR #593, opened 2024-02-17
Optimize XDDFChartData.java LOGGER

rh-id:


poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFChartData.java
```diff
@@ -170,8 +172,7 @@ public void replaceData(XDDFDataSource<?> category, XDDFNumericalDataSource<? ex
             if (categoryData != null && values != null) {
                 int numOfPoints = category.getPointCount();
                 if (numOfPoints != values.getPointCount()) {
-                    LOGGER.warn("Category and values must have the same point count, but had " +
-                            numOfPoints + " categories and " + values.getPointCount() + " values.");
+                    LOGGER.atWarn().log("Category and values must have the same point count, but had {} categories and {} values.", box(numOfPoints), box(values.getPointCount()));
```
xzel23: You get a 

- +1 for using the message format
- 0 for using `.atWarn().log(...)` instead of just `.warn(...)` - the fluent API is useful for doing some things that are nice, but not used in POI so this just makes the code longer and less concise.
- -1 for using `Unbox.box(...)`- in fact I don't believe this makes even the slightest difference in 99% of the scenarios and unless I see a thorough jmh test suite, I rather believe that this has a negative impact on performance (`Unbox.box()` does the string conversion every time, even if the result is not used because the log level is not enabled - just to prevent a simple boxing of a primitive wrapper which is probably the most tuned operation in JVMs for decades). IMHO that's the worst idea since what once lead to log4shell.

Last but not least, be aware that this is about logging a warning, maybe the last thing that needs any optimising - we should rather focus on minimising overhead for logging at debug and lower levels. And please do not use Unbox.box, it's the root of all evil (quote by Donald E. Knuth, read it up if you never heard about it).

So, just use the MessageFormat and revert the other changes.

xzel23 (changes requested):

rh-id:
Hi @xzel23, thanks for the review. I had updated the changes

xzel23 (approved):

PR #596, opened 2024-02-19
Improve readability and remove redundant cast at HSSFChart.java

rh-id:


poi/src/main/java/org/apache/poi/hssf/usermodel/HSSFChart.java
```diff
@@ -113,7 +113,7 @@ private HSSFChart(HSSFSheet sheet, ChartRecord chartRecord) {
     public void createBarChart( HSSFWorkbook workbook, HSSFSheet parentSheet )
     {
 
-        List<org.apache.poi.hssf.record.Record> records = new ArrayList<>();
+        List<Record> records = new ArrayList<>();
```
pjfanning, line 116: this is needed - Java 14+ has a java.lang.Record and we need to differentiate our Record class

pjfanning (changes requested):
most of this code is deliberate - have you tried building this with Java 17 or 21?

rh-id:
I'm using Java 21 and yes gradle build success.
I thought this is not deliberate because it has import Record class and no usage of java.lang.Record.
Perhaps we should just close this PR?

PR #598, opened 2024-02-20
Optimize AffineTransform.createTransformedShape call at DrawShape.java

rh-id:
If we look at [javadoc ](https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/java/awt/geom/AffineTransform.html#createTransformedShape(java.awt.Shape))

`AffineTransform.createTransformedShape(pSrc)` return `a new Shape object that defines the geometry of the transformed Shape, or null if pSrc is null.`

So this PR restructure the `AffineTransform.createTransformedShape` so that it checks the source whether it is null first instead of calling the method and check the return Shape value.

### DIFF
--- a/poi/src/main/java/org/apache/poi/sl/draw/DrawShape.java
+++ b/poi/src/main/java/org/apache/poi/sl/draw/DrawShape.java
@@ -192,8 +192,8 @@ public class DrawShape implements Drawable {
         }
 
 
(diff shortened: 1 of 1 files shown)


pjfanning:
* poi test coverage is incomplete and we have a long history of breaking stuff with small optimisations
* poi is a 20 year old project and is well past peak development
* I would much prefer to see changes driven by tests than making small changes that have no new test coverage

PR #601, opened 2024-02-29
XDGF: replace an elliptical arc with a line when a control point is co-linear with begin and end

komarovd95:
We've faced an issue with elliptical arcs in VSDX files - when a control point of the elliptic arc is co-linear (i.e. lies on the same line) with begin and end points of the arc, it leads to very huge numbers when calculating coordinates of the center of the arc. 

It happens because slopes of two tangents (class `EllipticalArcTo` lines 198 and 201) are equal to each other which leads to division by zero (or by value that is very near to zero).

As the solution it's proposed to check co-linearity before calculating an arc. I added this condition check into `EllipticalArcTo#createEllipticalArc` method to also cover corner-cases with `ArcTo` (which uses the method under the hood) and with `RelEllipticalArcTo`. 

Below you can find an example that high-lights the problem. 
Screenshot 2024-02-29 at 16 11 00

And after the fix: 
Screenshot 2024-02-29 at 16 12 11

The file itself (just change extension to `.vsdx` since GitHub does not allow to upload VSDX files)
[Single Icon.zip](https://github.com/apache/poi/files/14450770/Single.Icon.zip)

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xdgf/usermodel/shape/ShapeDebuggerRenderer.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xdgf/usermodel/shape/ShapeDebuggerRenderer.java
@@ -19,6 +19,7 @@ package org.apache.poi.xdgf.usermodel.shape;
 
 import java.awt.Font;
 import java.awt.Graphics2D;
+import java.awt.geom.Path2D;
 
 import org.apache.poi.xdgf.usermodel.XDGFShape;
 
@@ -39,9 +40,9 @@ public class ShapeDebuggerRenderer extends ShapeRenderer {
     }
 
     @Override
-    protected void drawPath(XDGFShape shape) {
+    protected Path2D drawPath(XDGFShape shape) {
 
-        super.drawPath(shape);
+        Path2D path = super.drawPath(shape);
         if (_debugAcceptor == null || _debugAcceptor.accept(shape)) {
 
             // show numbers to associate shapes with ids.. doesn't always work
@@ -62,6 +63,7 @@ public class ShapeDebuggerRenderer extends ShapeRenderer {
             _graphics.scale(1, -1);
         }
 
+        return path;
     }
 
 }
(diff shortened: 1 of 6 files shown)


poi-ooxml/src/main/java/org/apache/poi/xdgf/usermodel/shape/ShapeDebuggerRenderer.java
```diff
@@ -40,9 +39,9 @@ public void setDebugAcceptor(ShapeVisitorAcceptor acceptor) {
     }
 
     @Override
-    protected Path2D drawPath(XDGFShape shape) {
```
pjfanning: does this compile? Regardless, I don't think we should change protected method signatures in a patch release. Could you consider adding new private methods and leave the protected methods alone?

komarovd95: Hey @pjfanning! Good catch. I completely forgot that `ShapeRenderer` class might be used outside of `poi` codebase itself. 

Can you please check out the last revision? I reverted changing the method signature and preserved its behaviour. Now, it draws all the sections but returns only first drawn path.

komarovd95:
Hi @pjfanning! Can you please take a look one more time?

poi-ooxml/src/test/java/org/apache/poi/xdgf/usermodel/section/geometry/TestEllipticalArcTo.java
```diff
@@ -0,0 +1,76 @@
+package org.apache.poi.xdgf.usermodel.section.geometry;
```
pjfanning, line 18: could you add license headers to the new files?

komarovd95: Just pushed the changes. Headers are added to this file and `TestGeometryUtils.java`.

pjfanning:
thanks - merged

PR #607, opened 2024-03-14
Support SVGs in XWPF

Etienne-Gautier:
# Context
It is currently not possible to add SVGs to XWPF documents ... but it is possible in other formats such as XSLF ([here](https://github.com/Etienne-Gautier/poi/blob/25f00ba7c2db9bd897cffa67d12b1cfa31583516/poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFRelation.java#L240)).

This PR adds this feature + a basic test.


pjfanning:
merged - thanks

PR #617, opened 2024-04-08
fix for bug63189

ohkwon79:
This is fix for https://bz.apache.org/bugzilla/show_bug.cgi?id=63189

But, this doesn't seem to be a complete solution. 

Only when cloning a sheet, it has been patched so that only the hyperlinks on the drawing are copied properly.

We need to additionally check where the same problem occurs due to incorrect handling of hyperlink relation or other reference relations.

For a complete solution, it seems desirable to handle referenceRelation separately as I mentioned in bugzilla.


pjfanning (commented):
I applied this patch locally but the test fails for me.

```
            PackageRelationship linkRel2 = drawing2.getRelationPartById(linkRelId2).getRelationship();
            assertEquals(linkRelationType, linkRel2.getRelationshipType());
            assertEquals(linkTargetUrl, linkRel2.getTargetURI().getRawFragment());
```

This part of the test fails because drawing2.getRelationPartById(linkRelId2) is null.
The new code in XSSFWorkbook executes but the relations do not seem to be saved with the right ids.

pjfanning:
I think the reason the test might have appeared to work is because the test in the PR has a catch block that justs prints the stack trace but then swallows the exception. We cannot accept tests that swallow exceptions.

ohkwon79:
I removed catch block in the testcase. and I changed how to check hyperlink relation object.

However, unlike the original, the hyperlink relation is not included in the Relation of POIXMLDocumentPart.

This is because POIXMLDocumentPart only supports file-level association objects, and a proper solution would be possible only if it supports not only Relation to DocumentPart but also Relation to Reference.

And I think adding this feature would go beyond bug patch level.

pjfanning:
We'll accept big changes. We can just call the next release 5.3.0 or 6.0.0.

ohkwon79:
I added HyperlinkRelationship and created a fix using it.

Additionally, I added implementations of the following interfaces:
https://github.com/dotnet/Open-XML-SDK/blob/main/src/DocumentFormat.OpenXml.Framework/Features/IReferenceRelationshipsFeature.cs

The second commit is a modification to the part that reads and writes relationship information. This commit includes fixes to the tests affected by it.

pjfanning:
Thanks. I'm currently travelling but will have a look in a few days.

poi-ooxml/src/main/java/org/apache/poi/ooxml/HyperlinkRelationship.java
```diff
@@ -0,0 +1,29 @@
+package org.apache.poi.ooxml;
```
pjfanning, line 17: can you add an Apache source header here?

poi-ooxml/src/main/java/org/apache/poi/ooxml/POIXMLDocumentPart.java
```diff
@@ -18,12 +18,7 @@ Licensed to the Apache Software Foundation (ASF) under one or more
 
 import java.io.IOException;
 import java.net.URI;
-import java.util.ArrayList;
```
pjfanning: can you revert this change to this imports?

poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/internal/marshallers/ZipPartMarshaller.java
```diff
@@ -21,20 +21,14 @@ Licensed to the Apache Software Foundation (ASF) under one or more
...
 import org.apache.poi.openxml4j.exceptions.OpenXML4JException;
-import org.apache.poi.openxml4j.opc.PackageNamespaces;
-import org.apache.poi.openxml4j.opc.PackagePart;
-import org.apache.poi.openxml4j.opc.PackagePartName;
-import org.apache.poi.openxml4j.opc.PackageRelationship;
-import org.apache.poi.openxml4j.opc.PackageRelationshipCollection;
-import org.apache.poi.openxml4j.opc.PackagingURIHelper;
-import org.apache.poi.openxml4j.opc.StreamHelper;
-import org.apache.poi.openxml4j.opc.TargetMode;
+import org.apache.poi.openxml4j.opc.*;
```
pjfanning: can you keep the existing imports - instead of using wildcards?

poi-ooxml/src/test/java/org/apache/poi/openxml4j/opc/TestRelationships.java
```diff
@@ -323,10 +323,10 @@ private void assert_50154(OPCPackage pkg) throws Exception {
 
         PackageRelationship rId1 = drawingPart.getRelationship("rId1");
         URI parent = drawingPart.getPartName().getURI();
-        URI rel1 = parent.relativize(rId1.getTargetURI());
-        URI rel11 = PackagingURIHelper.relativizeURI(drawingPart.getPartName().getURI(), rId1.getTargetURI());
-        assertEquals("'Another Sheet'!A1", rel1.getFragment());
-        assertEquals("'Another Sheet'!A1", rel11.getFragment());
+//        URI rel1 = parent.relativize(rId1.getTargetURI());
```
pjfanning, line 327: why is this commented out?

poi-ooxml/src/main/java/org/apache/poi/ooxml/HyperlinkRelationship.java
ohkwon79: Done

poi-ooxml/src/main/java/org/apache/poi/ooxml/POIXMLDocumentPart.java
ohkwon79: Done

poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/internal/marshallers/ZipPartMarshaller.java
ohkwon79: Done

poi-ooxml/src/test/java/org/apache/poi/openxml4j/opc/TestRelationships.java
ohkwon79: This is because PackageRelationship.getTargetURI() has been modified to not resolve based on the source URI in the case of hyperlinks. (PackageRelationship.java)

```
        if (PackageRelationshipTypes.HYPERLINK_PART.equals(relationshipType)) {
            return targetUri;
        }
```

Because the targetURI of an internal hyperlink is not based on a physical file and only has a fragment, it is inappropriate to obtain the absolute path through resolve(). So, it does not resolve and returns only the fragment as is.

If targetURI is only a fragment like "#XXX", PackagingURIHelper.relativizeURI() returns "../" and thus fails the test. parent.relativize() normally returns a URI with only the fragment.

There is also a way to modify PackagingURIHelper.relativizeURI() to properly handle URIs with only fragments, like URI.relativize().

However, in my personal opinion, I think that it is inappropriate to resolve() or relativize() hyperlink relationships based on sourceURI. That's why I commented out this part.

ohkwon79: I added brief comments about this to PackageRelationship.java and TestRelations.java.

pjfanning:
Thanks @ohkwon79 - I merged this but I might make some changes. I still want to re-review the commented out sections of the tests.

PR #640, opened 2024-05-27
Fix ClassCastException inserting a paragraph into header or footer (bug 69042)

AntonOellerer:
Up until now, when trying to insert a paragraph into a header or footer via the `XWPFDocument.insertNewParagraph(XmlCursor)` method, a `ClassCastException` was thrown as the code
 incorrectly assumed that the object in the path to the cursor is a table in all cases.
To fix this, the method was updated so that the case where the path object is the header or footer is handled appropriately.
https://bz.apache.org/bugzilla/show_bug.cgi?id=69042


poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFDocument.java
```diff
@@ -912,6 +920,53 @@ private void insertElementIntoCell(XWPFTableCell tableCell, IBodyElement iBodyEl
         }
     }
 
+    private void insertIntoTables(IBodyElement iBodyElement, Deque<XmlObject> path, CTTbl ctTbl, List<XWPFTable> tables) {
+        for (XWPFTable xwpfTable : tables) {
+            if (ctTbl == xwpfTable.getCTTbl()) {
+                insertElementIntoTable(xwpfTable, iBodyElement, path);
+            }
```
pjfanning, line 930: shouldn't you break when you match in the `if`?

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFDocument.java
```diff
@@ -912,6 +920,53 @@ private void insertElementIntoCell(XWPFTableCell tableCell, IBodyElement iBodyEl
...
+        }
+    }
+
+    private void insertIntoHeadersFooters(IBodyElement iBodyElement, CTHdrFtrImpl ctHdrFtrImpl) {
+        for (XWPFHeader header : headers) {
+            if (ctHdrFtrImpl == header._getHdrFtr()) {
+                insertIntoHeaderFooter(iBodyElement, header);
```
pjfanning, line 937: same issue as above about break call when you match in the if and likewise in next for loop.

poi-ooxml/src/test/java/org/apache/poi/xwpf/TestXWPFBugs.java
```diff
@@ -25,9 +25,7 @@ Licensed to the Apache Software Foundation (ASF) under one or more
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-
 import javax.crypto.Cipher;
-
```
pjfanning: Can you add back this new line?

AntonOellerer:
Thank you for the review, I added `breaks`/`returns` where necessary!

AntonOellerer:
I did some more work around this problem and discovered that the problem might actually lie somewhere else, as `XWPFDocument.insertNewParagraph(XmlCursor)` should not be invoked if the parent object is actually a `XWPFHeaderFooter`, instead we should use `XWPFHeaderFooter.insertNewParagraph(XmlCursor)`.
What is your opinion on this @pjfanning?

pjfanning:
My preference is to make as few changes as possible to POI. It is a 20 year old project and it is very easy to break stuff when attempting to make changes.

PR #653, opened 2024-07-03
fixing a possible merge issue

tohidemyname:


centic9:
Thanks for the suggestion, this is applied now.

Also updated JavaDoc a bit. 

Note: We prevent creation of sheets where the name only differs in upper-/lowercase, but a "broken" input file might lead to slightly different behavior after this, but we now handle this consistently across different formats.

PR #655, opened 2024-07-05
Fixing a search error

tohidemyname:
searchText retruns wrong index.


centic9:
Thanks for the suggestion, applied the actual fix now.

Note: I did not change to getRList() as this can have negative impact on performance, we saw that calling getRList() may  introduce additional overhead.

Also did not add the appendText() method, please add a separate PR with accompanying tests if you would like to add such a method.

PR #657, opened 2024-07-07
SXSSF: support setting an arbitrary extra width value for column widths

pjfanning:
see #541 

It's hard to calculate a default width for SXSSF columns. The current calculated widths might be too small. This setting lets users add an arbitrary extra width to the generated column widths. It is one value for all columns.

There might be an argument to allow this setting to be set for individual columns.

* The value must be set before you start adding rows the sheet for it to be properly applied.
* Negative values are allowed but not encouraged

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/AutoSizeColumnTracker.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/AutoSizeColumnTracker.java
@@ -60,6 +60,8 @@ import org.apache.poi.util.Internal;
     // Using a HashSet instead of a TreeSet because we don't care about order.
     private final Set<Integer> untrackedColumns = new HashSet<>();
     private boolean trackAllColumns;
+    // arbitraryExtraWidth is the extra width added to the best-fit column width (since POI 5.3.1)
+    private double arbitraryExtraWidth = 0.0d;
 
     /**
      * Tuple to store the column widths considering and not considering merged cells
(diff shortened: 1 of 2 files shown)


pjfanning:
merged with https://github.com/apache/poi/commit/41453f3916fefd469cfa554dcb5ca069d221b331

jkmcl:
@pjfanning May I ask why using the new method `setArbitraryExtraWidth` introduced by this PR makes no difference in the code below?

```java
private static void createFile(Path path, boolean extraWidth) throws IOException {
    try (var workbook = new SXSSFWorkbook()) {
        var sheet = workbook.createSheet();
        if (extraWidth) {
            // *** Call the new method here ***
            sheet.setArbitraryExtraWidth(SheetUtil.getDefaultCharWidthAsFloat(workbook) * 2);
        }
        sheet.trackColumnForAutoSizing(0);

        var row = sheet.createRow(0);
        row.createCell(0).setCellValue("This is a very long text that will exceed default column width for sure.");
        sheet.autoSizeColumn(0);

        try (var stream = Files.newOutputStream(path)) {
            workbook.write(stream);
        }
    }
}

@Test
void mainTest() throws IOException {
    var outDir = Path.of("target/test-output");
    Files.createDirectories(outDir);

    createFile(outDir.resolve("no_extra_width.xlsx"), false);
    createFile(outDir.resolve("extra_width.xlsx"), true);
}
```

The above `@Test` method creates two files with almost the same code path except that `setArbitraryExtraWidth` is not called when creating `no_extra_width.xlsx`. I was expecting the width of the first column in this file to be shorter but it turned out to be the same in both files:

image

Does some condition need to be met for the extra width to be honored?

FYI, I ran the code on a Windows 10 machine with Eclipse Temurin JDK 17.0.13. The complete Maven project containing the code can be found here: https://github.com/jkmcl/poi-sample

Thanks.

pjfanning:
@jkmcl It looks like the 5.4.0 arbitrary extra width code doesn't work. I reimplemented it in https://github.com/apache/poi/commit/9eda2463986e817d8c4ebbc6f51d33c06311a86c

Would you be able to try that change?

The original 5.4.0 change was never really completed or reviewed.

You could also try:
```
sheet.setColumnWidth(0, sheet.getColumnWidth(0) + 100);
```

If you want to widen a column after calling autoSizeColumn(0).

jkmcl:
@pjfanning Yes, my test code created a file with a wider column when `setArbitraryExtraWidth` was called using https://github.com/apache/poi/commit/9eda2463986e817d8c4ebbc6f51d33c06311a86c

Thanks!

PR #670, opened 2024-08-12
XWPF: fix _getText in XWPFRun(ignore w:delInstrText, convert w:noBreakHyphen to "‑")

fangd1997:
1. w:delInstrText also need to be ignored
2. w:noBreakHyphen needs to be converted to "‑"
@pjfanning Please help review the code

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
@@ -1521,10 +1521,10 @@ public class XWPFRun implements ISDTContents, IRunElement, CharacterRun {
 
         if (o instanceof CTText) {
             final Node node = o.getDomNode();
-            // Field Codes (w:instrText, defined in spec sec. 17.16.23)
+            // Field Codes (w:instrText, defined in spec sec. 17.16.23 and w:delInstrText, defined in spec sec. 17.16.13)
             //  come up as instances of CTText, but we don't want them
             //  in the normal text output
-            if (!("instrText".equals(node.getLocalName()) && XSSFRelation.NS_WORDPROCESSINGML.equals(node.getNamespaceURI()))) {
+            if (!("instrText".equals(node.getLocalName()) && !("delInstrText".equals(node.getLocalName()) && XSSFRelation.NS_WORDPROCESSINGML.equals(node.getNamespaceURI()))) {
                 String textValue = ((CTText) o).getStringValue();
                 if (textValue != null) {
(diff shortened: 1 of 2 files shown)


pjfanning:
It is my understanding that noBreakHyphens are displayed optionally - if, for instance, you display a paragraph in a narrow space. Most people extracting this text will have no interest in seeing the hyphen. For me, this behaviour would have to be optional and default to the existing behaviour.

 For delInstrText, please provide a sample docx file that can be used to test with.

fangd1997:
[TestNoBreakHyphenAndDelInstrText.docx](https://github.com/user-attachments/files/16592924/TestNoBreakHyphenAndDelInstrText.docx)
image
the result in test: 
image
In this example, the "-" in the first `Figure 1-1` is noBreakHypen, and I think it is necessary to display it.

The second `Figure 1-1` has delInstrText, just like InstrText, it shouldn't actually be displayed

fangd1997:
The Sample File: [TestNoBreakHyphenAndDelInstrText.docx](https://github.com/user-attachments/files/16592924/TestNoBreakHyphenAndDelInstrText.docx)
image
the result in test: 
image
In this example, the "-" in the first `Figure 1-1` is noBreakHypen, and I think it is necessary to display it.
FYI: https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.wordprocessing.nobreakhyphen?view=openxml-3.0.1
The second `Figure 1-1` has delInstrText, just like InstrText, it shouldn't actually be displayed

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
```diff
@@ -1564,6 +1564,9 @@ private void _getText(XmlObject o, StringBuilder text) {
             final Node node = o.getDomNode();
             if (XSSFRelation.NS_WORDPROCESSINGML.equals(node.getNamespaceURI())) {
                 switch (node.getLocalName()) {
+                    case "noBreakHyphen":
+                        text.append('‑');
```
pjfanning, line 1568: is this a simple minus character or is the Unicode Non-Breaking-Hyphen char (https://www.compart.com/en/unicode/U+2011)?

fangd1997: Due to the inability to directly copy the ‑ character from Word, I chose Unicode Non-Breaking-Hyphen char(/u2011), which has the same meaning, as the replacement for w:noBreakHyphen

pjfanning:
Thanks - merged

PR #672, opened 2024-08-15
XWPF: support removing style by pos and getting all XWPFStyle list

fangd1997:
1. support getting all XWPFStyle list
2. support removing specified style by pos
@pjfanning


pjfanning:
Thanks - merged

PR #673, opened 2024-08-15
XWPF: generate classes for word10 xsds

fangd1997:
@pjfanning 
The main purpose is to add support for paraId and textId in CTP
support http://schemas.microsoft.com/office/word/2010/wordml
FYI：https://learn.microsoft.com/en-us/openspecs/office_standards/ms-docx/9704b59f-bc49-4618-ac66-41beb82a0d7f
diff in wml.xsd:
import word10.xsd and word12.xsd into wml.xsd
image
1. add attribute w14:anchorId into CT_Object and CT_Picture
image
2. add element w15:dataBinding and element w14:checkbox into CT_SdtPr
image
3. add w14:paraId, w14:textId and w14:noSpellErr into CT_P
image
4. add w14:paraId and w14:textId into CT_Row
image


poi-ooxml-full/src/main/xmlschema/org/apache/poi/schemas/OfficeOpenXML-XMLSchema-Transitional.zip
pjfanning, line 1: why is this being removed?

fangd1997: I did not remove this file, I just modified wml.xsd in this zip file.
The specific diff can be found in my first comment.

PR #674, opened 2024-08-16
XWPF: improve XWPFSDT

fangd1997:
Split XWPFSDT into  XWPFSDTBlock, XWPFSDTRow, XWPFSDTCell, XWPFSDTRun


pjfanning:
This breaks backward compatibility. I can not consider merging this without a detailed description of why we need to break compatibility.

This approach of raising loads of PRs but not raising issues in our bug tracker is very bad practice. Large changes should be discussed before they are implemented.

fangd1997:
Okay, I will close this PR
I originally planned to improve the support for SDT, as xWPF's support for SDT is less usable

PR #691, opened 2024-09-13
Add a check to verify that the tmp/poifiles directory still exists

girgen:
There are mechanisms outside the application, such as periodic routines in the operating system, that will clean tmp directories of old files. Make sure we recreate the directory if this happens.

The refers to bug 69323 and a regression introduced on commit 1def0cba387d8c86
https://bz.apache.org/bugzilla/show_bug.cgi?id=69323


pjfanning:
I guess that this is an improvement but looking at the code there are still possible race conditions - where the dir exists when a method call starts but gets deleted during execution. This is a real edge case. I'll probably merge this more or less as is and worry about edge cases another time.

pjfanning:
Just as a warning, there are no POI releases planned in the near future - if this issue is affecting you in production, you can register your own strategy instance that contains this check.

PR #692, opened 2024-09-20
D* functions are incompatible with the diamond operator

LukSpiewak:
The DSUM formula evaluation returns 0.0 when the condition contains the diamond operator <>.

## Problem
The code misinterprets the <> operator as a 'less than' comparison when it appears at the beginning of a condition.

## Modifications 
I added support for the diamond operator for both text and numeric data.


pjfanning:
Thanks - merged

LukSpiewak:
Hello, thank you @pjfanning for quick review and merge!
When do you plan to release the next version?

pjfanning:
There are no releases scheduled. Maybe in a few months.
You have options like building the jar for yourself or using the APIs in POI to register custom function implementations.
I would be especially interested in seeing wider use of the APIs to add custom function implementations as the current model of people expecting us to do releases after every commit is not sustainable.

sfisque:
is there a link to a nightly build that could be offered to "early adopters"?

pjfanning:
We have a web site and a page about this.

https://poi.apache.org/devel/nightly.html

PR #700, opened 2024-09-27
Make the default temporary directory contain the username to avoid permission issues when deploying applications on the same server with multiple users

TigerZCoder:
When deploying applications for multiple users on the same server, according to the current poi behavior, a fixed temporary directory named 'poifiles' is always created in <java.io.tmpdir> by default, which will cause 'java.io.IOException: Permission denied' exceptions in applications run by other users. Here I add the <user.name> as part of the directory name like the jdk perf dir 'hsperfdata_<userid>' to avoid the permission issue.

### DIFF
--- a/poi/src/main/java/org/apache/poi/util/DefaultTempFileCreationStrategy.java
+++ b/poi/src/main/java/org/apache/poi/util/DefaultTempFileCreationStrategy.java
@@ -47,6 +47,9 @@ public class DefaultTempFileCreationStrategy implements TempFileCreationStrategy
     /** To use files.deleteOnExit after clean JVM exit, set the <code>-Dpoi.delete.tmp.files.on.exit</code> JVM property */
     public static final String DELETE_FILES_ON_EXIT = "poi.delete.tmp.files.on.exit";
 
+    /** JVM property for the current username. */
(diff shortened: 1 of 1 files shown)


pjfanning:
I don't support this. The temp file support is pluggable. You can plug in your implementation. Most people are happy with the default - as it is already coded.

You can also set the -Djava.io.tmpdir system property to whatever you like.

xzel23:
How about using Files.createTempDirectory(POIFILES)?

TigerZCoder:
Yes, I know it's pluggable and we solved this with a custom plugin. But how about a friendlier out-of-the-box behavior?

pjfanning:
POI is 20 years old and has thousands of users and noone else is complaining about this. Changing behaviour on users is a bad thing.
Some users may have Java security set ups that ban calls to System.getProperty(JAVA_PROP_USER_NAME) - I used to work in a team where this was the case.
If you want to supply a 2nd implementation of TempFileCreationStrategy - maybe called UserNameAwareTempFileCreationStrategy - then we could add that as a non-default implementation that users could switch to if they like.

TigerZCoder:
OK. This is indeed a behavior that should be kept to avoid surprises. I think it is a good suggestion to provide an alternative implementation. I will modify the code and submit it later.

PR #704, opened 2024-10-05
Make the default temporary directory contain the username to avoid permission issues when deploying applications with multiple users on the same server

TigerZCoder:
When deploying applications for multiple users on the same server, according to the current poi behavior, a fixed temporary directory named 'poifiles' is always created in <java.io.tmpdir> by default, which will cause 'java.io.IOException: Permission denied' exceptions in applications run by other users. Here I add the <user.name> as part of the directory name like the jdk perf dir 'hsperfdata_' to avoid the permission issue.

Following @pjfanning 's suggestion, I created a new class to avoid modifying the default behavior.

Actually it's for https://github.com/apache/poi/pull/700

### DIFF
--- a/poi/src/main/java/org/apache/poi/util/UserNameAwareDefaultTempFileCreationStrategy.java
+++ b/poi/src/main/java/org/apache/poi/util/UserNameAwareDefaultTempFileCreationStrategy.java
@@ -28,8 +28,7 @@ import static org.apache.poi.util.TempFile.JAVA_IO_TMPDIR;
  * that avoids permission issues when deploying applications with multiple users on the same server.
  * Other than adding the username to the temporary directory, all other behavior is the same as the superclass.
  *
- * @author Tiger Zhao
- * @since 2024-10-05 22:41
+ * @since POI 5.3.1
  */
 public class UserNameAwareDefaultTempFileCreationStrategy extends DefaultTempFileCreationStrategy {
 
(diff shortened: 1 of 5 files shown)


poi/src/main/java/org/apache/poi/util/UserNameAwareDefaultTempFileCreationStrategy.java
```diff
@@ -0,0 +1,57 @@
...
+
+import static org.apache.poi.util.TempFile.JAVA_IO_TMPDIR;
+
+/**
+ * Username-aware subclass of {@link DefaultTempFileCreationStrategy}
+ * that avoids permission issues when deploying applications with multiple users on the same server.
+ * Other than adding the username to the temporary directory, all other behavior is the same as the superclass.
+ *
+ * @author Tiger Zhao
+ * @since 2024-10-05 22:41
```
pjfanning: Thanks but can you remove this author tag. We don't use tags like this in our shared code.

`@since POI 5.3.1` too - since is for version numbers not times.

pjfanning:
Can you write a test?

pjfanning (changes requested):
the word default should not appear in the new file name - this is not a default strategy

TigerZCoder:
I named the subclass UserNameAwareDefaultTempFileCreationStrategy because it behaves the same as DefaultTempFileCreationStrategy except for the poifiles directory name. But it's OK to remove the "default" to distinguish it from the default strategy, I'll rename it.

pjfanning:
Thanks - merged

TigerZCoder:
Thanks

PR #709, opened 2024-10-15
using hashmap is not thread-safe, may cause death cycle in concurrent use

venderxiu:
using hashmap is not thread-safe, may cause death cycle in concurrent use

### DIFF
--- a/poi/src/main/java/org/apache/poi/util/BitFieldFactory.java
+++ b/poi/src/main/java/org/apache/poi/util/BitFieldFactory.java
@@ -19,12 +19,13 @@
 package org.apache.poi.util;
 
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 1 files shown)


pjfanning:
POI is not designed for multithreaded use. I guess we can make some changes but I'd prefer to see some POC to demo the issue first. I don't see the point making single changes when we are fairly likely to have other issues too.

pjfanning:
I can see that this is a static field. That definitely seems like a bad idea - and makes it more important to fix.

pjfanning:
Thanks - merged with ab68210d2ace09d73a9468d908e61ab56750bb31

RaymondLam1:
Nice 👍

jinsihou19:
niubi

yvan-ouyang:
niubi🐂

PR #718, opened 2024-10-26
Bug 69411: Changed XSSFReader#getSheetsData to return a SheetIterator instead of Iterator<InputStream>

InfiniteLoop90:
This would cut down on the common need to cast the returned value to get access to the sheet name, sheet comments, etc.


InfiniteLoop90:
Closing as per maintainer's comment here about breaking API compatibility: https://bz.apache.org/bugzilla/show_bug.cgi?id=69411#c1

PR #727, opened 2024-11-19
Do not replace the @DSTAMP@ token when generating Version.java

ebourg:
The timestamp was removed from `Version.java` last year (23caf67c58fb5e6ecbe50eb1a24b02ef40585629), but the Ant and Gradle builds still attempt to replace the `@DSTAMP@` token. This PR removes this now unnecessary replacement.


ebourg:
Thanks!

PR #728, opened 2024-11-19
Write the number of blocks used by the property table in the header of files using 4K blocks only

ebourg:
This is a follow up to the PR #462 for Bug 66590, the header field must be updated only if the file uses 4K blocks. If small blocks are used the field must remain nul.


pjfanning:
This causes a lot of test failures.

Example:
Test testWritePropertyCount4096() FAILED

ebourg:
Indeed, I misused writeOutAndReadBack, the tested files are modified but shouldn't.

ebourg:
Fixed, the tests pass now.

ebourg:
Thank you

PR #730, opened 2024-11-20
Misplaced comment in DirectoryNode

ebourg:
There is a misplaced comment in the `DirectoryNode` class, the `iterator()` and `spliterator()` methods aren't part of the `POIFSViewable` interface and should be under the `/* **********  END  begin implementation of POIFSViewable ********** */` comment.


centic9:
Thanks for the change, I ended up reformatting the whole file some more to better fit the coding style in other areas.

PR #770, opened 2025-03-11
Bugfix setBulletStyle in XSLFTextParagraph

GermanAlejo:
Having a 0 here throws IllegalArgumentException, the method setBulletAutoNumber requires 1 as minimun as argument.


pjfanning:
This won't be accepted unless there is a detailed description of why this is needed and test cases are needed.

GermanAlejo:
I discovered this bug while using the library myself. The method setBulletStyle from XSLFTextParagraph class calls setBulletAutoNumber method, which expects a int as a parameter (startAt) this is hardcoded so minimun value allowed is 1, if the value is less than 1 then the method will throw IllegalArgumentException

image

The problem is method setBulletStyle has a minimun hardcoded value of 0 when calling this method, which means that each time we call it with a AutoNumberingScheme argument it will always throw an IllegalArgumentException, this renders this method quite useless, changing the 0 by 1 would solve the issue.

GermanAlejo:
I have added a couple of tests, however I am not knowledgeable in testing so please let me know If this should be done in any way different or if it is insufficient and I will update the PR. 

Let me know aswell if the explanation provided is not enough.

pjfanning:
Thanks @GermanAlejo - merged. I will have a look out for issues and if there are, I may need to revert this.

PR #775, opened 2025-03-19
Allow some OPC compliance checks to be tuned via optional configuration

krrg:
(Feature Proposal)

Currently POI strictly enforces compliance with the OPC specs and provides no mechanism by which these checks can be disabled by the library consumer. Unfortunately, there do exist documents in the wild that do not _strictly_ adhere to one or more of these specifications but which still need to be parsed. I have observed a number of documents "in the wild" over the years which are not strictly compliant, but otherwise can be parsed successfully were it not for the strict compliance checking.

I would like to introduce a new `OPCComplianceFlags` class which consumers may optionally include when opening an OPCPackage. This will allow users to optionally disable the current strict checking behavior at a granular level for M4.2 through M4.5, allowing them to parse documents that are technically non-compliant, but which are otherwise valid. 

This change should be completely backwards compatible. All existing public APIs are retained and will default to the existing behavior of strict OPC enforcement if no compliance flag parameter is passed. I have added a few additional tests to ensure these flags work if specified.

Thanks in advance for any feedback you might have!

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/OPCPackage.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/OPCPackage.java
@@ -151,6 +151,7 @@ public abstract class OPCPackage implements RelationshipSource, Closeable {
      * @param opcComplianceFlags Enable or disable specific OPC compliance flags.
      *                           This is useful to allow parsing of certain non-compliant documents.
      * @throws OpenXML4JRuntimeException if there are issues creating properties part
+     * @since POI 5.4.1
      */
     OPCPackage(PackageAccess access, OPCComplianceFlags opcComplianceFlags) {
         if (getClass() != ZipPackage.class) {
@@ -202,6 +203,7 @@ public abstract class OPCPackage implements RelationshipSource, Closeable {
      * @throws InvalidFormatException
      *             If the specified file doesn't exist, and a parsing error
      *             occur.
+     * @since POI 5.4.1
      */
     public static OPCPackage open(String path, OPCComplianceFlags opcComplianceFlags) throws InvalidFormatException {
         return open(path, defaultPackageAccess, opcComplianceFlags);
@@ -232,6 +234,7 @@ public abstract class OPCPackage implements RelationshipSource, Closeable {
      * @throws InvalidFormatException
      *             If the specified file doesn't exist, and a parsing error
      *             occur.
+     * @since POI 5.4.1
      */
     public static OPCPackage open(File file, OPCComplianceFlags opcComplianceFlags) throws InvalidFormatException {
         return open(file, defaultPackageAccess, opcComplianceFlags);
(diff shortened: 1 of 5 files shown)


poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/OPCPackage.java
```diff
@@ -141,14 +141,26 @@ public abstract class OPCPackage implements RelationshipSource, Closeable {
...
+
+    /**
+     * Constructor.
+     *
+     * @param access Package access.
+     * @param opcComplianceFlags Enable or disable specific OPC compliance flags.
+     *                           This is useful to allow parsing of certain non-compliant documents.
+     * @throws OpenXML4JRuntimeException if there are issues creating properties part
+     */
+    OPCPackage(PackageAccess access, OPCComplianceFlags opcComplianceFlags) {
```
pjfanning, line 156: can you add `@since POI 5.4.1` to new non-private methods and constructors?

pjfanning: it's ok with new classes just to put this `@since` on the class level content and not on methods/constructors.

poi-ooxml/src/test/java/org/apache/poi/openxml4j/opc/compliance/TestOPCComplianceCoreProperties.java
```diff
@@ -295,4 +305,50 @@ void testNoCoreProperties_saveInPlace() throws Exception {
...
         assertTrue(tmp.delete());
     }
+
+    /**
+     * Test M4.2 rule not enforced when appropriate OPCComplianceFlag is set.
+     */
+    @Test
+    void testDisableEnforceCompatibilityMarkup() throws IOException {
+        testComplianceFlagOverridePreventsException(
+                "DoNotUseCompatibilityMarkupFAIL.docx",
```
pjfanning, line 315: these files - should they be in the PR or are they already in the repo? It doesn't look right that they have no relative path info.

pjfanning: if these are new files that need to be added to the repo (test-data dir and it has subdirs based on file type), it will probably break the poi-integration tests which run automatic tests on all files in test-data dir and invalid files will need to be explicitly skipped

krrg: These files already exist and are already being used in the neighboring tests. The existing tests in this file check that we _do_ enforce compliance. My additional tests confirm that when a specific compliance setting (or settings) are turned off, the file then parses without error.

I'm using the same `OpenXML4JTestDataSamples.openSampleStream` method the other tests use, which I believe handles the actual prefixing, etc.

I ran all of the tests with gradle test, and everything passed on my machine including all the `poi-integration` tests, so I think we're okay? Happy to adjust anything if I missed something.

pjfanning:
Thanks - merged

PR #806, opened 2025-05-10
Avoiding needless `java.awt.Color` dependency in `HSSFColor`

JaroslavTulach:
I am currently working on project called [Enso](https://github.com/enso-org/enso). We are making [heavy use of Apache POI](https://github.com/enso-org/enso/blob/59eb1f2aeb6369fe2abde95f66385a3927064a9e/build.sbt#L1395) to parse Excel documents our users are processing.

[Excel in Enso](https://community.ensoanalytics.com/c/enso101/sections/245392/lessons/914097 "Excel in Enso")

We are also using [GraalVM Native Image](http://graalvm.org) to compile our application to _native executable_. Recently we realized in https://github.com/enso-org/enso/pull/12843 that it is necessary to **avoid dependency on `java.desktop`** module.

### Why Avoiding `java.desktop` Dependency?

Since Java9 the JDK offers tool called [jlink](https://openjdk.org/jeps/282). It can be used to create a smaller JDK ready for execution in the cloud and other restricted environments:
- one can specify set of JDK modules to include 
- often it is useful to exclude `java.desktop`
- in https://github.com/enso-org/enso/pull/12843 exclusion of `java.desktop` was necessary
- as GraalVM was crashing when `java.desktop` was included

Excluding `java.desktop` is a good practice anyway - reduces bloat of the final application by tens of megabytes.

### Patching `HSSFColor`

- to resolve all the needs of https://github.com/enso-org/enso/pull/12843
- I had to [copy the code of HSSFColor](https://github.com/enso-org/enso/pull/12843/files#r2081091554) into our repository
- and modify it to avoid instantiating `java.awt.Color`

Such a solution was good short-term workaround, but I **don't want to maintain a fork** of POI! Hence my goal is to find a way to include the necessary fixes in your project in the long term. Hence the creation of this PR. Can you please consider its acceptance?

### Technicalities

- this PR replaces the RGB handling code in `HSSFColor`
- originally this code relied on `java.awt.Color`
   - however that creates _unnecessary dependency_ on `java.desktop`
- the PR proposes direct manipulation with RGB `int` compatible with handling in `java.awt.Color`
- the _behavior remains the same_
- dependency of `HSSFColor` on `java.desktop` is removed

### Other Fixes

- there are [other places in POI where dependency](https://github.com/enso-org/enso/pull/12843) on `java.desktop` deserves to be removed 
-  however that's a long term direction
- I am starting with this simple `HSSFColor` proposal 
- if it is accepted, I create subsequent PRs addressing other `java.desktop` dependencies

### DIFF
--- a/poi/src/main/java/org/apache/poi/hssf/util/HSSFColor.java
+++ b/poi/src/main/java/org/apache/poi/hssf/util/HSSFColor.java
@@ -353,16 +353,15 @@ public class HSSFColor implements Color {
                Integer.toHexString(getBlue()*0x101)).toUpperCase(Locale.ROOT);
     }
 
-    final short getBlue() {
+    private final short getBlue() {
         return (short) (rgb & 0xff);
     }
-    final short getGreen() {
+    private final short getGreen() {
         return (short) ((rgb >> 8) & 0xff);
     }
-    final short getRed() {
+    private final short getRed() {
         return (short) ((rgb >> 16) & 0xff);
     }
-
 
     @Override
     public boolean equals(Object o) {
--- a/poi/src/main/java/org/apache/poi/hssf/util/HSSFColor.java
+++ b/poi/src/main/java/org/apache/poi/hssf/util/HSSFColor.java
@@ -42,7 +42,7 @@ public class HSSFColor implements Color {
     private static Map<Integer,HSSFColor> indexHash;
     private static Map<HSSFColorPredefined,HSSFColor> enumList;
 
-    private final java.awt.Color color;
+    private final int rgb;
     private final int index;
     private final int index2;
 
@@ -110,7 +110,7 @@ public class HSSFColor implements Color {
         private final HSSFColor color;
 
         HSSFColorPredefined(int index, int index2, int rgb) {
-            this.color = new HSSFColor(index, index2, new java.awt.Color(rgb));
+            this.color = new HSSFColor(index, index2, rgb);
         }
 
         /**
@@ -145,7 +145,7 @@ public class HSSFColor implements Color {
          * @return (a copy of) the HSSFColor assigned to the enum
          */
         public HSSFColor getColor() {
-            return new HSSFColor(getIndex(), getIndex2(), color.color);
+            return new HSSFColor(getIndex(), getIndex2(), color.rgb);
         }
     }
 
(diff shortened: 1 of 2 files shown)


poi/src/main/java/org/apache/poi/hssf/util/HSSFColor.java
```diff
@@ -145,21 +145,45 @@ public String getHexString() {
...
+     * new HSSFColor(index, index2, color.getRGB());
+     * </pre>
+     * or specify RGB directly.
+     *
+     * @param index
+     * @param index2
+     * @param color
+     * @deprecated prefer constructor that specifies RGB directly as a value
+     */
+    @Deprecated
```
pjfanning: why deprecate a constructor if simply prefer another constructor?

poi/src/main/java/org/apache/poi/hssf/util/HSSFColor.java
```diff
@@ -324,11 +348,22 @@ public short getIndex2() {
...
     public String getHexString() {
-        return (Integer.toHexString(color.getRed()*0x101) + ":" +
-               Integer.toHexString(color.getGreen()*0x101) + ":" +
-               Integer.toHexString(color.getBlue()*0x101)).toUpperCase(Locale.ROOT);
+        return (Integer.toHexString(getRed()*0x101) + ":" +
+               Integer.toHexString(getGreen()*0x101) + ":" +
+               Integer.toHexString(getBlue()*0x101)).toUpperCase(Locale.ROOT);
     }
 
+    final short getBlue() {
```
pjfanning: would there a be a benefit to making these methods public?

So far, you don't appear to use these methods outside this class, so they could even be made private.

myeyestore:
The documentation in this repository is excellent! Very clear and comprehensive.

poi/src/main/java/org/apache/poi/hssf/util/HSSFColor.java
JaroslavTulach: - I didn't want to change the API - unless necessary
- there are no usages of these methods outside of this class thus I am making them private in 449dba6
- the `color` field wasn't publicly accessible either (luckily)

poi/src/main/java/org/apache/poi/hssf/util/HSSFColor.java
JaroslavTulach: - probably a _hidden thought_ in my mind
   - I'd like the usage of `java.awt.Color` to disappear one day
   - thus deprecating
- but the "Color constructor" doesn't need to be _deprecated_
- another thought: 
   - the new constructor doesn't need to be public
   - that way we could minimize the publicly visible API changes to zero
   - I said I prefer not to [change the API](https://github.com/apache/poi/pull/806#discussion_r2083226065) unless necessary, then
   - keeping the "Color constructor" and hiding the RGB one would live to that promise...

### Three Options

1. deprecate "Color constructor" and introduce public "RGB constructor"
2. leave "Color constructor" and make the "RGB constructor" (package) private
3. leave "Color constructor" and introduce public  "RGB constructor"

Any of those options works for Enso use-case. I can do any of them. My slight preference goes to option no. 1.

pjfanning:
I'd prefer to make the new constructor public but not to deprecate the old one.

poi/src/main/java/org/apache/poi/hssf/util/HSSFColor.java
JaroslavTulach: > I'd prefer to make the new constructor public but not to deprecate the old one.

Done in 32fea65

pjfanning:
thanks - merged

JaroslavTulach:
> thanks - merged

I see https://github.com/apache/poi/commit/8bac1b33f6ddfc7dbc2b83c28419026f651b5c63 - thanks for accepting my change. It will make Enso better. When do you expect next release binary artifacts to be uploaded to Maven central?

PS: I was hoping to get a commit authored by me in the POI history to expand on my [ASF contributions](https://projects.apache.org/committee.html?netbeans). Maybe next time.

pjfanning:
We don't use git in POI. We use subversion. This GitHub project is a readonly mirror.

There is no release expected in the near future. The POI community is small and releases need multiple people to do reviews and get involved in the votes.

You can build the jars from source or we have regular builds in CI - https://ci-builds.apache.org/

JaroslavTulach:
> There is no release expected in the near future. 

Your [web indicates](https://poi.apache.org/changes.html) you may perform a release by end of the calendar year (it shows `Version 5.4.2 (2025-??-??)`). True? 

> releases need multiple people ... get involved in the votes.

As a PMC member of [Apache NetBeans](https://people.apache.org/phonebook.html?unix=netbeans) I can give you non-binding +1 ;-)

PR #808, opened 2025-05-16
feat: read trackChange

donglining:
read trackChange  in sdtcontent


pjfanning:
Thanks but can you add tests? Have you run the existing tests? Can you provide some background about the change?

pjfanning:
-1 from me

I think the change tracking should not be hacked as if is a text run. If we were to add such changes, I think they should not be ignored by the parser by default and the change data should be stored in dedicated classes. The last thing we need is to have our POI Word parser returning deleted paragraphs and text as if they were still visible in the doc.

donglining:
> Thanks but can you add tests? Have you run the existing tests? Can you provide some background about the change?

>  Currently, the change tracking in XWPFSDTContent cannot be read. I can  add some tests
I believe change tracking should be readable, allowing users to determine whether to display it. In XWPFParagraph, insertion (ins) change tracking is currently read as run text.

like this ,  the  'EF'  ins trackChange  can‘t  read ,  

image

pjfanning:
> -1 from me
> 
> I think the change tracking should not be hacked as if is a text run. If we were to add such changes, I think they should not be ignored by the parser by default and the change data should be stored in dedicated classes. The last thing we need is to have our POI Word parser returning deleted paragraphs and text as if they were still visible in the doc.

I will try try to veto this change. The change tracking should not be treated like a text run. It needs it own dedicated handling.

donglining:
> > -1 from me
> > I think the change tracking should not be hacked as if is a text run. If we were to add such changes, I think they should not be ignored by the parser by default and the change data should be stored in dedicated classes. The last thing we need is to have our POI Word parser returning deleted paragraphs and text as if they were still visible in the doc.
> 
> I will try try to veto this change. The change tracking should not be treated like a text run. It needs it own dedicated handling.

like  XWPFParagraph   , change tracking treated like a text run.
image

pjfanning:
If you add test coverage then this can be reviewed but it is bad practice to provide PRs with no description and no tests.

donglining:
> If you add test coverage then this can be reviewed but it is bad practice to provide PRs with no description and no tests.

Got it. I'll proceed with adding test cases.

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFSDTContent.java
```diff
@@ -123,8 +134,12 @@ public String getText() {
                 text.append(((XWPFSDT) o).getContent().getText());
                 addNewLine = true;
             } else if (o instanceof XWPFRun) {
-                text.append(o);
-                addNewLine = false;
+                XWPFRun xRun = (XWPFRun) o;
+                // don't include the text if reviewing is enabled and this is a deleted run
+                if (xRun.getCTR().getDelTextArray().length == 0) {
```
pjfanning, line 139: use isEmpty() instead of length == 0

pjfanning:
I downloaded the docx in the PR. Word shows the suggested changes but they are just suggestions. They are not the saved state of the document.

donglining:
> I downloaded the docx in the PR. Word shows the suggested changes but they are just suggestions. They are not the saved state of the document.

Yes, but the INS (insertion) track changes in XWPFParagraph are being read as TextRun objects. I want them to have consistent behavior (i.e., either all treated as tracked changes or all as final text).

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFSDTContent.java
donglining: I will make the corresponding changes here.

donglining: but  Here, it's an array, not a collection.

PR #823, opened 2025-06-14
Allowing custom TempFileCreationStrategy per document

kelemen:
As is now, we can only set a global `TempFileCreationStrategy` which assumes that all documents within the same process has the same security profile (and quota) which may or may not be the case. Currently it is not possible to workaround this limitation in a multithreaded application.

This PR intends to fix this issue by allowing to set a `TempFileCreationStrategy` for `SXSSFWorkbook` and `ZipPackage` instances.

Implementation notes:

- I have not deleted the `BinaryRC4Encryptor.BinaryRC4CipherOutputStream(DirectoryNode)` constructor because it is protected and in theory might be used (though given that other implementations have the counterparts of this class private, I'm guessing this visibility is unintended).
- When adjusting the `OPCPackage.open` methods (to avoid duplicating logic), I'm relying on a new `OPCPackageOpener` class. However, strictly speaking my implementation is not 100% backward compatible (can be adjusted of course if needed), because the new `OPCPackageOpener` immediately throws NPE when it receives null `PackageAccess` or `OPCComplianceFlags`, while the original implementation does not. Rather, the original implementation will simply store the nulls, and will (likely) throw an exception on first use. Should I adjust my changes to be 100% backward compatible, or is this an acceptable risk for this project?
- Not directly related to this PR, but while making the adjustments, I have noticed that `DefaultTempFileCreationStrategy` is buggy (at least I doubt its behavior is intended): When it detects that the temporary directory does not exist, then it will reset the directory to what is in `java.io.tmpdir` even if a different directory was specified in the constructor. I'm mentioning this only, because this issue can be a security risk (if someone changed the tmp dir due to security reasons), and this PR might make this issue more visible.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/OPCPackageOpener.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/OPCPackageOpener.java
@@ -1,0 +1,364 @@
+/* ====================================================================
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
+   The ASF licenses this file to You under the Apache License, Version 2.0
+   (the "License"); you may not use this file except in compliance with
+   the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+==================================================================== */
+
+package org.apache.poi.openxml4j.opc;
+
+import static org.apache.poi.openxml4j.opc.ContentTypes.PLAIN_OLD_XML;
+import static org.apache.poi.openxml4j.opc.ContentTypes.RELATIONSHIPS_PART;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Date;
+import java.util.Objects;
+import java.util.Optional;
+
+import org.apache.poi.openxml4j.exceptions.InvalidFormatException;
+import org.apache.poi.openxml4j.exceptions.InvalidOperationException;
+import org.apache.poi.openxml4j.opc.internal.InvalidZipException;
+import org.apache.poi.openxml4j.opc.internal.PackagePropertiesPart;
+import org.apache.poi.openxml4j.opc.internal.ZipContentTypeManager;
+import org.apache.poi.openxml4j.util.ZipEntrySource;
+import org.apache.poi.util.IOUtils;
+import org.apache.poi.util.StringUtil;
+import org.apache.poi.util.TempFileCreationStrategy;
+
+/**
+ * Defines a class to open OPC packages with particular configuration.
+ *
+ * @since POI 5.5.0
+ */
+public final class OPCPackageOpener {
+    private static final OPCComplianceFlags defaultOpcComplianceFlags = OPCComplianceFlags.enforceAll();
+
+    /**
+     * Package access.
+     */
+    private PackageAccess packageAccess;
+
+    /**
+     * Whether OPC compliance requirements are checked (e.g., M4.2, M4.3, M4.4, and M4.5)
+     */
+    private OPCComplianceFlags opcComplianceFlags;
+
+    /**
+     * Strategy to create temporary files.
+     */
+    private TempFileCreationStrategy tmpStrategy;
+
+    public OPCPackageOpener() {
+        this.packageAccess = OPCPackage.defaultPackageAccess;
+        this.opcComplianceFlags = defaultOpcComplianceFlags;
+        this.tmpStrategy = TempFileCreationStrategy.getDefaultStrategy();
+    }
+
+    /**
+     * Sets the package access.
+     *
+     * @param packageAccess the package access. May not be {@code null}.
+     */
+    public void setPackageAccess(PackageAccess packageAccess) {
+        this.packageAccess = Objects.requireNonNull(packageAccess, "packageAccess");
+    }
+
+    /**
+     * Sets whether OPC compliance requirements are checked (e.g., M4.2, M4.3, M4.4, and M4.5).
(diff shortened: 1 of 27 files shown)


poi-examples/src/main/java/org/apache/poi/examples/xssf/streaming/SavePasswordProtectedXlsx.java
```diff
@@ -95,7 +96,7 @@ public static void save(final InputStream inputStream, final String filename, fi
             EncryptionInfo info = new EncryptionInfo(EncryptionMode.agile);
             Encryptor enc = Encryptor.getInstance(info);
             enc.confirmPassword(pwd);
-            opc.save(enc.getDataStream(fs));
+            opc.save(enc.getDataStream(fs, TempFileCreationStrategy.getDefaultStrategy()));
```
pjfanning: please revert this change - there is no need to complicate the example with this

poi/src/main/java/org/apache/poi/poifs/crypt/ChunkedCipherOutputStream.java
```diff
@@ -63,14 +64,14 @@ public abstract class ChunkedCipherOutputStream extends FilterOutputStream {
     private Cipher cipher;
     private boolean isClosed;
 
-    public ChunkedCipherOutputStream(DirectoryNode dir, int chunkSize) throws IOException, GeneralSecurityException {
+    public ChunkedCipherOutputStream(DirectoryNode dir, int chunkSize, TempFileCreationStrategy tmpStrategy) throws IOException, GeneralSecurityException {
```
pjfanning, line 71: add back the old constructor for compat reasons

poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/OPCPackageOpener.java
```diff
@@ -0,0 +1,362 @@
...
+import org.apache.poi.openxml4j.opc.internal.InvalidZipException;
+import org.apache.poi.openxml4j.opc.internal.PackagePropertiesPart;
+import org.apache.poi.openxml4j.opc.internal.ZipContentTypeManager;
+import org.apache.poi.openxml4j.util.ZipEntrySource;
+import org.apache.poi.util.IOUtils;
+import org.apache.poi.util.StringUtil;
+import org.apache.poi.util.TempFileCreationStrategy;
+
+/**
+ * Defines a class to open OPC packages with particular configuration.
```
pjfanning, line 42: add `@since POI 5.5.0` on all new public methods and classes

poi-ooxml/src/main/java/org/apache/poi/xssf/streaming/GZIPSheetDataWriter.java
```diff
@@ -56,7 +68,7 @@ public GZIPSheetDataWriter(SharedStringsTable sharedStringsTable) throws IOExcep
     @Deprecated
     @Removal(version = "6.0.0")
     public File createTempFile() throws IOException {
-        return TempFile.createTempFile("poi-sxssf-sheet-xml", ".gz");
+        return _tmpStrategy.createTempFile("poi-sxssf-sheet-xml", ".gz");
```
pjfanning: this method is not used in POI - so you don't need to update it

poi/src/main/java/org/apache/poi/poifs/crypt/ChunkedCipherOutputStream.java
kelemen: Ah sorry, I thought I was careful ... Will add it back.

kelemen: Although truth to be told this class is marked as `@Internal`, so I guess when I was adjusting this, that was in my head.

pjfanning:
Raise the issue about DefaultTempFileCreationStrategy in https://bz.apache.org/bugzilla

pjfanning:
Has this been tested in Java 8? The CI in this repo does not work and I don't want to waste time on this PR if it doesn't compile.

I'm on the fence about this change. Frankly, I don't understand why we need to maintain complicated temp file strategy code that only a few people need. The community is small and POI is a lib about parsing and writing Microsoft format files. I struggle to care about people's complicated temp file requirements.

kelemen:
- Made the adjustments as requested in the above comments.
- I have created a report at <https://bz.apache.org/bugzilla/show_bug.cgi?id=69714>.
- I have run `gradle build` locally with Java 8 set up (I have just verified that, if I add `var x = 1;`, then it doesn't compile), and there were no issues.

As for the rational: I'm not sure what you consider "community", but poi is very widely known and used (since Excel reports are quite common). I can't really estimate how many people would want this feature, but I did bump into it (i.e., I do need this feature), and found it quite sad that there was no workaround, so I figured I might as well volunteer.

pjfanning:
Thanks for raising the bugzilla issue. I was actually talking about this though.

> Not directly related to this PR, but while making the adjustments, I have noticed that DefaultTempFileCreationStrategy is buggy (at least I doubt its behavior is intended): When it detects that the temporary directory does not exist, then it will reset the directory to what is in java.io.tmpdir even if a different directory was specified in the constructor. I'm mentioning this only, because this issue can be a security risk (if someone changed the tmp dir due to security reasons), and this PR might make this issue more visible.

This needs to be recorded somewhere if you don't intend to work on it.

poi-examples/src/main/java/org/apache/poi/examples/xssf/streaming/SavePasswordProtectedXlsx.java
kelemen: reverted

kelemen:
> This needs to be recorded somewhere if you don't intend to work on it.

If you are willing to accept a fix about that, then I will happily fix it in a separate PR. But let me report it regardless.

pjfanning:
The small community that I am talking about is the active PMC members for POI. The code doesn't merge itself. The jars don't release themselves.
There is a pretty big backlog of PRs and bugzilla issues to review. At the moment, I am going to zone out on this PR because I see it as a low priority compared to other backlog items.

kelemen:
I have created <https://bz.apache.org/bugzilla/show_bug.cgi?id=69715> about `DefaultTempFileCreationStrategy`.

kelemen:
Closing this, because the actual problem was solved by #825, so this change is no longer necessary.

PR #824, opened 2025-06-14
Fixes DefaultTempFileCreationStrategy when explicitly specified directory gets deleted

kelemen:
This PR fixes <https://bz.apache.org/bugzilla/show_bug.cgi?id=69715>, and as a side effect also fixes some race conditions which could have resulted in an NPE or spurious an `IOException`.


pjfanning:
-1 from me

* the temp file strategy is pluggable so anyone who needs this can override the strategy in POI - we don't need to merge this

kelemen:
I assumed from your acknowledgment to my comment on #823 that you are ok with me fixing it.

Also, this is a clearly bug, I'm not sure why leaving bugs in the code is ok (not to mention the race condition as well).

PR #825, opened 2025-06-16
Thread local temp file strategy

kelemen:
Added support for thread local temp file strategy.

The only awkwardness here are checked exceptions likely thrown in the task (in fact, most likely IO is likely). It might be reasonable to create an interface instead of `Supplier` declaring the `IOException`.

### DIFF
--- a/poi/src/main/java/org/apache/poi/util/TempFile.java
+++ b/poi/src/main/java/org/apache/poi/util/TempFile.java
@@ -19,6 +19,8 @@ package org.apache.poi.util;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.Objects;
+import java.util.function.Supplier;
 
 /**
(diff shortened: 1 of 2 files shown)


pjfanning:
@kelemen I merged this more or less as is. I left out the boolean flag as I'm not convinced it will aid performance that much.
Bugzilla seems to have crashed but I was emailed your most recent comment.
I fixed up the set(null) scenario do a remove.

kelemen:
Thanks, I have seen it, but couldn't respond on BugZilla (for the same reason). I wasn't that convinced about the flag as well, just figured that you will remove it (or ask to remove it), if you don't want to complicate.

Just one note: Do you still want to retain your `setThreadLocalTempFileCreationStrategy` method? Obviously, it is not an issue for me, but it might limit your possibilities in the future, if for who knows what reason you need or want to move to scoped values (as scoped values only support the "safe" style only).

PR #829, opened 2025-06-20
Allow null values in XWPFParagraph.get/setAlignment().

jaragunde:
A null value in this field would indicate that the paragraph should follow the alignment provided by the style hierarchy.

Fixes: https://bz.apache.org/bugzilla/show_bug.cgi?id=69720


jaragunde:
Hi! A gentle ping. Anything I can do on my side to make review easier? Maybe I missed some contribution step (e.g. signing a contributor agreement)? Thanks in advance for your time 🙂

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFParagraph.java
```diff
@@ -506,13 +506,18 @@ public ParagraphAlignment getAlignment() {
...
+     * @param align the paragraph alignment to apply to this paragraph. It can
+     *              be null to unset it and fall back to the style hierarchy.
      */
     public void setAlignment(ParagraphAlignment align) {
         CTPPr pr = getCTPPr();
-        CTJc jc = pr.isSetJc() ? pr.getJc() : pr.addNewJc();
-        STJc.Enum en = STJc.Enum.forInt(align.getValue());
-        jc.setVal(en);
+        if (align == null) {
+            pr.unsetJc();
```
pjfanning: * is there some way that getCTPPr(init = false) could be used here? we have equivalent code elsewhere
* we don't want to create a CTPPr instance if none already exists for the null input case

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFParagraph.java
```diff
@@ -488,11 +488,11 @@ public String getFootnoteText() {
      * specified in the style hierarchy, then no alignment is applied to the
      * paragraph.
      *
-     * @return the paragraph alignment of this paragraph.
+     * @return the paragraph alignment of this paragraph. Can be null if not set.
```
pjfanning: This is a breaking change. I know you don't like the way things work but unfortunately, in a 20 year old project with lots of users, we can't just change API behaviours. We have no major releases planned, just patch releases - so if you want to break the API, this change will need to wait.

Can you instead consider adding a new method `public boolean isAlignmentSet()`? New methods won't cause backward compatibility issues.

pjfanning (changes requested):

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFParagraph.java
jaragunde: Thanks a lot for the insight! You are totally right, backwards API compatibility must be considered, and a method that starts returning null can cause a lot of headaches... I will implement this behavior as a new method instead.

What do you think about allowing `setAlignment(null)`? It sounds less dangerous from the compatibility POV, but maybe it looks a bit weird? Better with a new method `unsetAlignment()`?

pjfanning: I think your change for `setAlignment(null)` is actually ok. That change is more of a bug fix because `STJc.Enum en = STJc.Enum.forInt(align.getValue())` (existing code) will NPE if align=null.

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFParagraph.java
jaragunde: I added `getCTPPr(create)` in one of the patches I just uploaded, and I kept `getCTPPr()` with no arguments to call `getCTPPr(true)`, preserving the original behavior. I looked for similar implementations and found `XSLFPlaceholderDetails.getCTPlaceholder(create)`, although this kind of behavior didn't seem to be very widely used in the codebase.

I can submit it as a separate patch/PR, whatever is most convenient for you. Thanks!

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFParagraph.java
jaragunde: I reverted `getAlignment()` to the original behavior and implemented `isAlignmentSet()` as discussed. Thanks!

pjfanning (approved):
lgtm - thanks

jaragunde:
Thanks for your review!

PR #843, opened 2025-07-14
Add getters and setters for XWPFTable indentation.

jaragunde:
The element tblPr->tblInd represents "table indent from leading margin". It specifies the indentation which shall be added before the leading edge of the current table in the document (the left edge in a left-to-right table, and the right edge in a right-to-left table). This indentation should shift the table into the text margin by the specified amount.

This value is specified in the units applied via its type attribute. Any width value of type pct or auto for this element shall be ignored.

If this element is omitted, then the table shall inherit the table indentation from the associated table style. If table indentation is never specified in the style hierarchy, no indentation shall be added to the parent table. If the resulting justification on any table row is not left, then this property shall be ignored.

To implement this property, we included a getter and a setter for the property value, and another couple of getter/setter for the existence (and validity) or absence of the property itself, for clients to know they must fall back to the table style.

Suggestions about the API style are very welcome! We could have achieved the same with two methods `get/setIndent(Integer)` allowing null values to indicate that the element is absent or invalid.

Thanks in advance for your time.


poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFTable.java
```diff
@@ -306,6 +306,80 @@ public int getNumberOfRows() {
...
+
+    /**
+     * Set the indentation in 20ths of a point (twips).
+     * @see int getIndent()
+     * @param indent Indentation value (20ths of a point)
+     */
+    public void setIndent(int indent) {
+        CTTblPr tblPr = getTblPr();
+        CTTblWidth tblInd = tblPr.isSetTblInd() ? tblPr.getTblInd() : tblPr.addNewTblInd();
+        tblInd.setW(new BigInteger(Integer.toString(indent)));
```
pjfanning: use https://docs.oracle.com/javase/8/docs/api/java/math/BigInteger.html#BigInteger-java.lang.String-

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFTable.java
```diff
@@ -306,6 +306,80 @@ public int getNumberOfRows() {
...
+     * Check if some indentation value is set for the table.
+     *
+     * <p>If this attribute is omitted, then the table shall inherit the table indentation from
+     * the associated table style. If table indentation is never specified in the style hierarchy,
+     * no indentation shall be added to the parent table.</p>
+     *
+     * @return true if the indent value is set and is valid, false if it is not set or shall be
+     * ignored (e.g. due to invalid type).
+     */
+    public boolean isSetIndent() {
```
pjfanning: hasIndent?

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFTable.java
```diff
@@ -306,6 +306,80 @@ public int getNumberOfRows() {
...
+        CTTblPr tblPr = getTblPr();
```
pjfanning: getTblPr(false)

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFTable.java
```diff
@@ -306,6 +306,80 @@ public int getNumberOfRows() {
...
+     * <p>This element specifies the indentation which shall be added before the leading edge of
+     * the current table in the document (the left edge in a left-to-right table, and the right
+     * edge in a right-to-left table).</p>
+     * <p>If the table alignment is not left/start, this property shall be ignored.</p>
+     *
+     * @see boolean isSetIndent()
+     * @return indentation value as an integer (20ths of a point)
+     */
+    public int getIndent() {
+        CTTblPr tblPr = getTblPr();
```
pjfanning: getTblPr(false) -- we don't want to modify the data for a get call

poi-ooxml/src/test/java/org/apache/poi/xwpf/usermodel/TestXWPFTable.java
```diff
@@ -586,4 +584,66 @@ void testSetGetTableAlignment() throws IOException {
...
+            assertTrue(tbl.isSetIndent());
+            assertEquals(0, tbl.getIndent());
+            tbl.setIndent(-100);
+            assertTrue(tbl.isSetIndent());
+            assertEquals(-100, tbl.getIndent());
+            tbl.removeIndent();
+            assertFalse(tbl.isSetIndent());
+            assertEquals(0, tbl.getIndent());
+        }
+    }
```
pjfanning, line 648: can you add a new line at the end of this file?

pjfanning: I think rebasing your branch to the head of the trunk branch will sort this out. I added https://github.com/apache/poi/commit/d4fc5cd6c08338a4132a6348ee1b1b077f9527b8

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFTable.java
pjfanning: Are you sure that BigInteger is the right type here? setW takes an Object but I wouldn't be surprised if it needs to be a specific set of classes that need to be used.

There is an xsetW method thats take this type.
```
/**
 * An XML ST_MeasurementOrPercent(@http://schemas.openxmlformats.org/wordprocessingml/2006/main).
 *
 * This is a union type. Instances are of one of the following types:
 *     org.openxmlformats.schemas.wordprocessingml.x2006.main.STUnqualifiedPercentage
 *     org.openxmlformats.schemas.officeDocument.x2006.sharedTypes.STPercentage
 *     org.openxmlformats.schemas.officeDocument.x2006.sharedTypes.STUniversalMeasure
 */
public interface STMeasurementOrPercent
```

jaragunde: I think we are safe using BigInteger. I inspected the getter and I see it returns a BigInteger. Besides, `setWidth()` also makes use of a `CTTblWidth` object and it sets a BigInteger.

I checked the BigInteger API as suggested and noticed BigInteger.valueOf(int) which is a better fit.

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFTable.java
jaragunde: Done!

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFTable.java
jaragunde: Done

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFTable.java
jaragunde: Done

poi-ooxml/src/test/java/org/apache/poi/xwpf/usermodel/TestXWPFTable.java
jaragunde: And done!

pjfanning:
thanks - merged

PR #845, opened 2025-07-15
[EXPERIMENT] test with xmlbeans snapshot

pjfanning:
* an experimental build with a heavily modified XMLBeans jar
* changes are basically those in https://github.com/apache/xmlbeans/pull/21
* I published the snapshot using my own groupId because I have easy access to that


pjfanning:
I was able to make changes in XMLBeans that didn't break POI but I found the hoped for performance boost was not working out

PR #848, opened 2025-07-17
Support all possible alignment values for docx tables.

jaragunde:
Besides the start/center/end values specified in the OOXML standard, Word also uses "left" and "right" as values. We need to support this to prevent POI code from crashing with such documents.

Fixes: https://bz.apache.org/bugzilla/show_bug.cgi?id=69744


pjfanning:
Thanks - merged

PR #854, opened 2025-07-23
Add getters for CTSdt(Content)Cell objects.

jaragunde:
Currently the XWPFSdt(Content)Cell object just produces a text equivalent of the content. There is no way to access the inner XML node for further inspection; instead, one needs to go to the parent object and iterate over its XML children.

This would allow users to use the XWPFRow.getICells() to obtain XWPFTableCell or XWPFSDTCell objects, and then access the latter's inner XML node to get access to its full data.

We also defer the parsing of the text in XWPFContentCell until the getter is called. A user who will work with the inner XML is unlikely to need the text parsing done.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFSDTCell.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFSDTCell.java
@@ -29,9 +29,11 @@ import org.openxmlformats.schemas.wordprocessingml.x2006.main.CTSdtCell;
  */
 public class XWPFSDTCell extends XWPFAbstractSDT implements ICell {
     private final XWPFSDTContentCell cellContent;
+    private final CTSdtCell sdtCell;
 
     public XWPFSDTCell(CTSdtCell sdtCell, XWPFTableRow xwpfTableRow, IBody part) {
         super(sdtCell.getSdtPr(), part);
+        this.sdtCell = sdtCell;
         cellContent = new XWPFSDTContentCell(sdtCell.getSdtContent(), xwpfTableRow, part);
     }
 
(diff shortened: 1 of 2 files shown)


jaragunde:
This is a bit of a speculative PR on my side :-) I think it improves the project, even if slightly, and it matches the existing API (most objects have their corresponding CT getters). On the other hand, there is the related XWPFSdt class that supports three different types of Sdt node, each of them backed by a different CTSdt* class, so this change doesn't align very well with the other class.

Your insight would be very valuable. Thanks in advance for your time!

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFSDTCell.java
```diff
@@ -40,4 +42,7 @@ public ISDTContent getContent() {
         return cellContent;
     }
 
+    public CTSdtCell getSdtCell() {
```
pjfanning: Javadoc and `@since` - Include the CT in the method name to highlight that users are heading into generated XMLBean class territory.

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFSDTContentCell.java
```diff
@@ -103,12 +108,20 @@ private boolean isStartToken(XmlCursor cursor, String string) {
...
+        }
         return text;
     }
 
+    @Override
     public String toString() {
         return getText();
     }
+
+    public CTSdtContentCell getCTSdtContentCell() {
```
pjfanning, line 129: javadoc and `@since`

pjfanning:
merged - thanks

PR #859, opened 2025-07-24
Implement XWPFRun.getEmbeddedCharts().

jaragunde:
Returns a list of XWPFChart objects embedded in the run.

Works in a similar way to getEmbeddedPictures(), maintaining a list of XWPFChart objects. They are retrieved from the XWPFDocument object through their relation id.


pjfanning:
The code is consistent with the code style so I'll merge as is. I really wish that internal lists were not exposed via APIs and that copies were returned if we really need to expose the full list.

jaragunde:
I noticed some methods return an unmodifiableList and some don't; `XWPFParagraph.getRuns()` and `getIRuns()` are unmodifiable; `XWPFRun.getEmbeddedImages()` is not.

Most methods in `XWPFDocument` return unmodifiable lists, including `getCharts()`, I should probably have followed that example. I can definitely submit a follow-up.

PR #861, opened 2025-07-25
feat：addPicture support set behindDoc, x, y parameter

liyong473:
addPicture support set behindDoc, x, y parameter

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
@@ -59,6 +59,8 @@ import org.openxmlformats.schemas.drawingml.x2006.picture.CTPicture;
 import org.openxmlformats.schemas.drawingml.x2006.picture.CTPictureNonVisual;
 import org.openxmlformats.schemas.drawingml.x2006.wordprocessingDrawing.CTAnchor;
 import org.openxmlformats.schemas.drawingml.x2006.wordprocessingDrawing.CTInline;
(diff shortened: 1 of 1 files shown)


pjfanning:
* Can you remove the new method that takes `int pictureType`? I am deprecating this usage. See #865.
* Can you add a test case for the new method?

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFRun.java
```diff
@@ -1363,6 +1365,175 @@ public XWPFPicture addPicture(InputStream pictureData, PictureType pictureType,
...
+     * @param pictureType The {@link PictureType} of the picture
+     * @param width       width in EMUs. To convert to / from points use {@link org.apache.poi.util.Units}
+     * @param height      height in EMUs. To convert to / from points use {@link org.apache.poi.util.Units}
+     * @param isBehindDoc      The "behindDoc" attribute of the CTAnchor.
+     * @param x      X Position Offset in EMUs. To convert to / from points use {@link org.apache.poi.util.Units}
+     * @param y      Y Position Offset in EMUs. To convert to / from points use {@link org.apache.poi.util.Units}
+     * @throws InvalidFormatException If the format of the picture is not known.
+     * @throws IOException            If reading the picture-data from the stream fails.
+     */
+    public XWPFPicture addPicture(InputStream pictureData, PictureType pictureType, String filename, int width, int height, boolean isBehindDoc, int x, int y)
```
pjfanning, line 1410: can you add `@since POI 5.5.0`?

PR #862, opened 2025-07-25
Return unmodifiable lists from XWPFRun charts and pictures.

jaragunde:
The operations `XWPFRun.getEmbedded[Pictures|Charts]().add()` are not supported. Users must call `XWPFRun.add[Picture|Chart]()` instead.

This change will make the fact more explicit, with an error when they attempt to use the wrong operation.


jaragunde:
I understand that doesn't address your original concern (do not expose internal lists) but it's a corrective measure that's already in the codebase. I hope it's useful.

PR #869, opened 2025-07-30
check xwpf node depth

pjfanning:
ensure that we don't handle XML that is too complex


pjfanning:
I can revert this if anyone has objections

PR #870, opened 2025-08-01
add new XSSFParser

pjfanning:
POI has lots of RuntimeExceptions and it would be too much work to change to using checked exceptions. Thousands of method signatures would change and lots of new try/catch changes would be needed in intermediate methods.

We can take some of the most used code paths and provide wrapper APIs that use checked exceptions by wrapping any non-fatal exception/error as a checked exception.

### DIFF
--- a/poi-ooxml/build.gradle
+++ b/poi-ooxml/build.gradle
@@ -55,14 +55,14 @@ java {
 
 dependencies {
     api project(':poi')
-    api project(':poi-ooxml-full')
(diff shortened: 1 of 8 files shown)


pjfanning:
@tballison @centic9 What do you think of this new API that wraps the core XSSFWorkbook constructor?
* I can add tests before merging it
* I can create equivalents for XWPFDocument and XMLSlideshow
* Eventually, the parser could also support configurable parsing - enabling and disabling features for a particular parse task, etc.

PR #888, opened 2025-08-25
Fix Excel's Accounting Format for zero part where 0 is randomly added at the end

dat07cs:
This is an attempt to try to handle a common case where `?` is used as alignment placeholder.

Given the custom number format for zero: `_-$* \"-\"??_-`

Current formatted value: `    - 0 ` `<4 spaces><literal-dash><space><zero><space>`
Expected formatted value: `    -   ` `(<4 spaces><literal-dash><3 spaces>)`


pjfanning:
The existing code is very complicated. The new changes make it even more complicated. It does appear to fix 2 asserts in the tests so that is good and doesn't appear to break anything.
I'm still somewhat reluctant to merge because the new code is so hard to read.
Maybe someone else who reviews it will find it easier to decipher.

PR #909, opened 2025-09-25
Delay initialization of `fontRenderContext` field & co. by moving it into own inner class

JaroslavTulach:
- This PR is a followup to #806 
- I am currently working on project called [Enso](https://github.com/enso-org/enso). 
- We are making [heavy use of Apache POI](https://github.com/enso-org/enso/blob/59eb1f2aeb6369fe2abde95f66385a3927064a9e/build.sbt#L1395) (version `5.2.3`) to parse Excel documents our users are processing.

[Excel in Enso](https://community.ensoanalytics.com/c/enso101/sections/245392/lessons/914097 "Excel in Enso")

- we are using `jlink` library and remove `java.desktop` module from our JDK
- that is causing us problems (just like the one in #806)
- the most recent problem occured in https://github.com/enso-org/enso/pull/13842
- after an investigation we found a decent workaround
   - step by step investigation is described in https://github.com/enso-org/enso/issues/13983#issuecomment-3332650938
   - and https://github.com/enso-org/enso/issues/13983#issuecomment-3332708320
   - the process of patching is documented at https://github.com/enso-org/enso/issues/13983#issuecomment-3333600671
   - which finally lead to creation of this as described [here](https://github.com/enso-org/enso/issues/13983#issuecomment-3333646569)
- given there already is `org.apache.poi.ss.ignoreMissingFontSystem` property
- Enso is unlikely the only project facing problems of using POI on a _headless system_
- Enso is just taking it to _another level_ by using `jlink` to remove `java.desktop` module
- however the need seems to be there: please accept our refactoring into upstream, thanks.


pjfanning:
Can you rebase due to the code conflict?

JaroslavTulach:
> Can you rebase due to the code conflict?

Rebased & conflicts resolved. It was a bit tricky, but I have also gladly found out you changed default value of `ignoreMissingFontSystem` to `true`. That's handy! 

It seems we are working towards the same goal of being able to run on systems that don't have _access to fonts_.

poi/src/main/java/org/apache/poi/ss/util/SheetUtil.java
```diff
@@ -238,56 +233,87 @@ public static double getCellWidth(Cell cell, float defaultCharWidth, DataFormatt
...
+
+        private static float getDefaultCharWidthAsFloat(AttributedString str) {
+            TextLayout layout = new TextLayout(str.getIterator(), fontRenderContext);
+            return layout.getAdvance();
+        }
+        
+        /**
+         * Copy text attributes from the supplied Font to Java2D AttributedString
+         */
+        private static void copyAttributes(Font font, AttributedString str, @SuppressWarnings("SameParameterValue") int startIdx, int endIdx) {
```
pjfanning, line 302: I don't like this method - it's name is confusing and why does it need the `SuppressWarnings`?

I would suggest a name like `modifyAttributedString(AttributedString attributedString, ...)`

pjfanning (changes requested):

poi/src/main/java/org/apache/poi/ss/util/SheetUtil.java
JaroslavTulach: - It is a [copy of an existing](https://github.com/apache/poi/pull/909/files#r2389874483) POI method. 
- I just moved it to the `WithJavaDesktop` inner class
- it can be renamed to whatever you wish, if that guarantees integration of this PR
- PS: I don't like the method either - because it is referencing `java.awt` classes!

poi/src/main/java/org/apache/poi/ss/util/SheetUtil.java
```diff
@@ -456,22 +481,16 @@ private static double getColumnWidthForRow(
...
+                return false;
+            }
+            throw t;
+        }
     }
 
-    /**
-     * Copy text attributes from the supplied Font to Java2D AttributedString
-     */
-    private static void copyAttributes(Font font, AttributedString str, @SuppressWarnings("SameParameterValue") int startIdx, int endIdx) {
```
JaroslavTulach, line 468: This is the original version of `copyAttributes` that is being [moved to an innerclass](https://github.com/apache/poi/pull/909/files#r2388939736) in this PR.

pjfanning: ok - I'll merge this and tinker with this method myself

JaroslavTulach:
- Thanks for accepting my change. 
- It will make Enso better - once we upgrade to upcoming version.
   - but we may need few more tweaks...
   - by when do we have to provide them?
- the [history of changes](https://poi.apache.org/changes.html) indicates there will be `2025-??-??` release
- by when do we have to provide our fixes to make it into upcoming release?

pjfanning:
* there are no concrete plans for a release
* it may not even happen in 2025 but it probably will happen later rather than sooner
* if you want to submit more changes, submit them but I can't guarantee that we will wait for changes before doing a release

PR #910, opened 2025-09-25
Make getters public for XML beans in XWPFDefault*Style.

jaragunde:
The classes XWPFDefaultRunStyle and XWPFDefaultParagraphStyle only expose a subset of their possibles attributes. To let users access all the underlying data, we change the visibility of the XML bean getters to public.


pjfanning:
thanks - merged

PR #912, opened 2025-09-26
Prevent NullPointerException in XWPFTable.getWidthType().

jaragunde:
According to the spec in section "17.4.63 tblW (Preferred Table Width)", the element tblW is not compulsory and "If this element is omitted, then the cell width shall be of type auto."

We modify the getter to follow this behavior, preventing a NPE.

It implies a change of behavior in the situation when there is no tblPr, in that case it used to return NIL, but notice that this behavior was introduced in commit d4fc5cd6c08338a4132a6348ee1b1b077f9527b8 and has never been part of a release.


pjfanning (approved):
lgtm - merged

PR #916, opened 2025-10-03
Remove deprecated dispose() calls from SXSSF workbook tests

AdRiley:
The SXSSFWorkbook.dispose() method was deprecated in https://github.com/apache/poi/pull/586

This PR removes use of it from the tests in favour of SXSSFWorkbook.close() called via try-with-resource blocks.


pjfanning:
It is still useful to test deprecated code. I'm not sure that this change is all that useful. If it was a sample, I'd say that avoiding the deprecation is good - but a test, I'm not convinced.

AdRiley:
Fair enough. I like my tests to run without warnings so people pay attention to warnings and don't get accustomed to ignoring them.

PR #920, opened 2025-10-17
New api for reading xlsb

AdRiley:
This PR introduces a new API for reading data from xlsb files with their native types as opposed to the current methodology which returns everything as a formatted string.

This is done via the introduction of a new callback handler XSSFBSheetContentsHandler which has typed callbacks.

This change is backwards compatible with the existing API which is still present and works as it did before.

This PR also adds support for reading the error codes from the xlsb file and returning the actual error.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
@@ -89,9 +89,9 @@ public class XSSFBSheetHandler extends XSSFBParser {
     /**
      * Creates a handler that converts numeric and date cells to formatted strings via {@link
      * DataFormatter}.
-     *
-     * <p>This variant mirrors the SAX-based API from {@link XSSFSheetXMLHandler} so existing POI
-     * consumers can reuse their {@link XSSFSheetXMLHandler.SheetContentsHandler} implementations.
+     * 
+     * <p>Select this overload when the consumer expects formatted string values rather than raw
+     * cell representations.
      *
      * @param is XLSB worksheet stream to parse
      * @param styles table providing cell style and number format metadata
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
@@ -166,7 +166,7 @@ public class XSSFBSheetHandler extends XSSFBParser {
                 handleFmlaNum(data);
                 break;
             case BrtFmlaError:
-                handleCellError(data);
+                handleFmlaError(data);
                 break;
                 //TODO: All the PCDI and PCDIA
             case BrtEndSheetData:
@@ -185,10 +185,10 @@ public class XSSFBSheetHandler extends XSSFBParser {
         checkMissedComments(currentRow, cellBuffer.getColNum());
     }
 
-    private void handleStringCellValue(String formattedValue) {
+    private void handleStringCellValue(String val) {
         CellAddress cellAddress = getCellAddress();
         XSSFBComment comment = getCellComment(cellAddress);
-        handler.stringCell(cellAddress.formatAsString(), formattedValue, comment);
+        handler.stringCell(cellAddress.formatAsString(), val, comment);
     }
 
     private void handleDoubleCellValue(double val) {
@@ -196,6 +196,18 @@ public class XSSFBSheetHandler extends XSSFBParser {
         XSSFBComment comment = getCellComment(cellAddress);
         ExcelNumberFormat nf = getExcelNumberFormat();
         handler.doubleCell(cellAddress.formatAsString(), val, comment, nf);
+    }
+
+    private void handleErrorCellValue(int val) {
+        FormulaError fe;
+        try {
+            fe = FormulaError.forInt(val);
+        } catch (IllegalArgumentException e) {
+            fe = null;
+        }
+        CellAddress cellAddress = getCellAddress();
+        XSSFBComment comment = getCellComment(cellAddress);
+        handler.errorCell(cellAddress.formatAsString(), fe, comment);
     }
 
     private CellAddress getCellAddress() {
@@ -248,16 +260,14 @@ public class XSSFBSheetHandler extends XSSFBParser {
 
     private void handleCellError(byte[] data) {
         beforeCellValue(data);
-        int bErr = data[XSSFBCellHeader.length] & 0xFF;
-        FormulaError fe;
-        try {
-            fe = FormulaError.forInt(bErr);
-        } catch (IllegalArgumentException e) {
-            fe = null;
-        }
-        CellAddress cellAddress = getCellAddress();
-        XSSFBComment comment = getCellComment(cellAddress);
-        handler.errorCell(cellAddress.formatAsString(), fe, comment);
+        int val = data[XSSFBCellHeader.length] & 0xFF;
+        handleErrorCellValue(val);
+    }
+
+    private void handleFmlaError(byte[] data) {
+        beforeCellValue(data);
+        int val = data[XSSFBCellHeader.length] & 0xFF;
+        handleErrorCellValue(val);
     }
 
     private void handleBoolean(byte[] data) {
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
@@ -23,6 +23,8 @@ import java.util.Queue;
 
 import org.apache.poi.ss.usermodel.BuiltinFormats;
 import org.apache.poi.ss.usermodel.DataFormatter;
+import org.apache.poi.ss.usermodel.ExcelNumberFormat;
+import org.apache.poi.ss.usermodel.FormulaError;
 import org.apache.poi.ss.usermodel.RichTextString;
 import org.apache.poi.ss.util.CellAddress;
 import org.apache.poi.util.Internal;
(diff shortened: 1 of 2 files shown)


poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
```diff
@@ -55,6 +56,54 @@ public class XSSFBSheetHandler extends XSSFBParser {
...
+     *
+     * <p>Select this overload when the consumer expects the raw cell representation rather than
+     * formatted strings.
+     *
+     * @param is XLSB worksheet stream to parse
+     * @param styles table providing cell style and number format metadata
+     * @param comments optional comments table, may be {@code null}
+     * @param strings shared strings table used by the sheet
+     * @param sheetContentsHandler callback receiving native cell events
+     * @param formulasNotResults {@code true} to request formulas rather than cached results
```
pjfanning: If you are not implementing this param then remove it

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
```diff
@@ -55,6 +56,54 @@ public class XSSFBSheetHandler extends XSSFBParser {
...
+     * <p>Select this overload when the consumer expects formatted string values rather than raw
+     * cell representations.
+     *
+     * @param is XLSB worksheet stream to parse
+     * @param styles table providing cell style and number format metadata
+     * @param comments optional comments table, may be {@code null}
+     * @param strings shared strings table used by the sheet
+     * @param sheetContentsHandler callback receiving formatted string values
+     * @param dataFormatter formatter applied to numeric and date cells
+     * @param formulasNotResults {@code true} to request formulas rather than cached results
```
pjfanning: Remove this param

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
```diff
@@ -114,7 +170,7 @@ public void handleRecord(int id, byte[] data) throws XSSFBParseException {
                 handleFmlaNum(data);
                 break;
             case BrtFmlaError:
-                handleFmlaError(data);
+                handleCellError(data);
```
pjfanning: Why is this changed?

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
```diff
@@ -133,86 +189,107 @@ private void beforeCellValue(byte[] data) {
...
+    }
+
+    private CellAddress getCellAddress() {
+        return new CellAddress(currentRow, cellBuffer.getColNum());
+    }
+
+    private XSSFBComment getCellComment(CellAddress cellAddress) {
         XSSFBComment comment = null;
         if (comments != null) {
-            comment = comments.get(cellAddress);
```
pjfanning, line 140: Why change the formatting?

pjfanning (changes requested):

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
AdRiley: handleFmlaError and handleCellError were identical before and still are. I can put handleFmlaError back if you prefer and move the shared code to a helper method. (Which I guess would closer follow the existing pattern)

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
AdRiley: Done

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
pjfanning: The PR is big already. Can this be undone?

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
AdRiley: Done

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
AdRiley: Typo sorry. Fixed.

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
AdRiley: Yes. I put it back. I think it is cleaner now.

pjfanning:
The CI run had a test failure. I'm travelling and bandwidth is too low to read the full log.

AdRiley:
I broke this test

  Test test66682() FAILED

  org.opentest4j.AssertionFailedError: expected: `<ERROR>` but was: <#DIV/0!>
  
  As I am now returning the actual error rather than just the generic ERROR.
  
  Options are
  1. Make this a breaking change and fix the test
  2. Make it so the existing API continues to return ERROR for cells with errors
  
  I can't decide which is the better option? Either is easy to do.

pjfanning:
I prefer to not break any tests. You can add a method or param that flips the behaviour but users should opt into the breaking change.

This is from real world pain with changes leading to complaints from users who rely on the behaviour as is and don't expect it to change.

AdRiley:
OK I updated the code so the error behaviour works as it did before for the old string based API. Anyone who needs the actual error codes can use the new API.

poi-ooxml/src/test/java/org/apache/poi/xssf/eventusermodel/TestXSSFBReader.java
```diff
@@ -17,22 +17,35 @@ Licensed to the Apache Software Foundation (ASF) under one or more
...
+import org.apache.poi.ss.usermodel.ExcelNumberFormat;
+import org.apache.poi.ss.usermodel.FormulaError;
 import org.apache.poi.xssf.binary.XSSFBSharedStringsTable;
 import org.apache.poi.xssf.binary.XSSFBSheetHandler;
 import org.apache.poi.xssf.binary.XSSFBStylesTable;
 import org.apache.poi.xssf.usermodel.XSSFComment;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
```
pjfanning, line 34: Can you fix the import order? static imports go together at the bottom separated from the rest
Some sort of alpha order among the 2 groups of imports.

poi-ooxml/src/test/java/org/apache/poi/xssf/eventusermodel/TestXSSFBReader.java
```diff
@@ -216,4 +229,234 @@ public String toString() {
...
+    }
+
+    private static ArgumentMatcher<XSSFComment> commentWith(String author, String text) {
+        return comment -> comment != null
+                && author.equals(comment.getAuthor())
+                && comment.getString() != null
+                && text.equals(comment.getString().toString().trim());
+    }
+
+    void testXSSFBSheetContentsHandler(String fileName,
```
pjfanning: missing `@Test` annotation

poi-ooxml/src/test/java/org/apache/poi/xssf/eventusermodel/TestXSSFBReader.java
```diff
@@ -216,4 +229,234 @@ public String toString() {
...
+            }
+        }
+    }
+
+    @Test
+    void testBasicXSSFBSheetContentsHandler() throws Exception {
+        XSSFBSheetHandler.XSSFBSheetContentsHandler handler = mockSheetContentsHandler();
+        testXSSFBSheetContentsHandler("testVarious.xlsb", handler);
+  
+        InOrder ordered = inOrder(handler);
```
pjfanning, line 274: Why do we need mockito? Can't this be tested with the APIs as is?

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
```diff
@@ -55,6 +56,50 @@ public class XSSFBSheetHandler extends XSSFBParser {
...
+     * XSSFBSheetContentsHandler}.
+     *
+     * <p>Select this overload when the consumer expects the raw cell representation rather than
+     * formatted strings.
+     *
+     * @param is XLSB worksheet stream to parse
+     * @param styles table providing cell style and number format metadata
+     * @param comments optional comments table, may be {@code null}
+     * @param strings shared strings table used by the sheet
+     * @param sheetContentsHandler callback receiving native cell events
```
pjfanning, line 71: Can you add `@since POI 5.5.0` on new public methods and constructors?

poi-ooxml/src/test/java/org/apache/poi/xssf/eventusermodel/TestXSSFBReader.java
AdRiley: Done

poi-ooxml/src/test/java/org/apache/poi/xssf/eventusermodel/TestXSSFBReader.java
AdRiley: It was actually a poorly named method. Fixed now.

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
AdRiley: Done

poi-ooxml/src/test/java/org/apache/poi/xssf/eventusermodel/TestXSSFBReader.java
AdRiley: Because I need to test that certain callbacks get called on the XSSFBSheetContentsHandler. So I needed to either create a mock handler by hand (which is what the existing tests did) or use a third party mocking framework like mockito (that I think was already been used in other places in the code base).

pjfanning (approved):
lgtm - merged

PR #950, opened 2025-11-24
Fix NPE in XWPFParagraph.getCTSpacing().

jaragunde:
We recently changed the getters so they don't create the PPr object if it doesn't exist, but we missed adding one null check for the new situation.

Add a test to exercise all the getters in a case where there isn't a PPr object.


pjfanning (approved):
Lgtm

PR #954, opened 2025-11-26
Support SVGs in XSSF

focJSi:
Currently, adding SVGs to a XSSF Workbook is not supported.

I added this feature similarly to #607 (Support SVGs in XWPF) with a basic test.
Do you have any tipps / suggestions to write a better test for this feature?

I have locally tried it out, generating [svg_sample.xlsx](https://github.com/user-attachments/files/23775300/svg_sample.xlsx).

I have read [here](https://stackoverflow.com/questions/67666241/how-to-add-svg-image-to-xssfworkbook) that for full backwards compatibility, one would have to convert the SVG to a PNG, as Excel does it.
From my undestanding, the same should apply to the added feature in #607, so I decided to give it a try and open the PR similarly. Any thoughts on this?

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFWorkbook.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFWorkbook.java
@@ -122,6 +122,7 @@ public class XSSFWorkbook extends POIXMLDocument implements Workbook, Date1904Su
     public static final int PICTURE_TYPE_EPS = 10;
     public static final int PICTURE_TYPE_BMP = 11;
     public static final int PICTURE_TYPE_WPG = 12;
+	public static final int PICTURE_TYPE_SVG = 14;
 
     /**
      * The underlying XML bean
@@ -592,6 +593,12 @@ public class XSSFWorkbook extends POIXMLDocument implements Workbook, Date1904Su
      * @see Workbook#PICTURE_TYPE_JPEG
      * @see Workbook#PICTURE_TYPE_PNG
      * @see Workbook#PICTURE_TYPE_DIB
+	 * @see #PICTURE_TYPE_GIF
+	 * @see #PICTURE_TYPE_TIFF
+	 * @see #PICTURE_TYPE_EPS
+	 * @see #PICTURE_TYPE_BMP
+	 * @see #PICTURE_TYPE_WPG
+	 * @see #PICTURE_TYPE_SVG
      * @see #getAllPictures()
      */
     @Override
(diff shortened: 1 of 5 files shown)


poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFWorkbook.java
```diff
@@ -122,6 +122,7 @@ public class XSSFWorkbook extends POIXMLDocument implements Workbook, Date1904Su
     public static final int PICTURE_TYPE_EPS = 10;
     public static final int PICTURE_TYPE_BMP = 11;
     public static final int PICTURE_TYPE_WPG = 12;
+	public static final int PICTURE_TYPE_SVG = 14;
```
focJSi, line 126: `org.apache.poi.common.usermodel.PictureType.WDP` has `ooxmlId` 13, which is missing here.
Should these constants be refactored to use `org.apache.poi.common.usermodel.PictureType#ooxmlId` in a seperate PR?

poi-ooxml/src/test/java/org/apache/poi/xssf/usermodel/TestXSSFPictureData.java
```diff
@@ -143,6 +143,37 @@ void testNewPictFormat() throws IOException {
         }
     }
 
+	@Test
+	void testNewSvgFormat() throws IOException {
+		try (XSSFWorkbook wb = new XSSFWorkbook()) {
+			XSSFSheet sheet = wb.createSheet();
+			XSSFDrawing drawing = sheet.createDrawingPatriarch();
+
+			byte[] data = "test svg data".getBytes(LocaleUtil.CHARSET_1252);
```
rPraml: This is not a valid svg image data (which should not be relevant for this test).
But maybe you can use a valid svg string, so it will not break if someone adds a SVG->PNG converter for backward compatibility

https://dev.w3.org/SVG/tools/svgweb/samples/svg-files/decimal.svg, for example:
```xml
<svg viewBox='0 0 125 80' xmlns='http://www.w3.org/2000/svg'>
  <text y="75" font-size="100" font-family="serif"><![CDATA[10]]></text>
</svg>
```

focJSi: Agreed, adjusted the data

poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFWorkbook.java
pjfanning: can you add the WDP one here as a placeholder for a later refactor?

focJSi: Sure, I added a comment as a placeholder. Hope that was what you were looking for :)

PR #958, opened 2025-12-01
Bugzilla 69303: Fix RangeCopier's copyMergedRanges when copying within the same sheet

vo0ff:
Fixes https://bz.apache.org/bugzilla/show_bug.cgi?id=69303

Features, added by https://github.com/apache/poi/pull/179, namely option to copy merged ranges, while useful, has several issues:
1. Works by copying all merged regions, performs no checks if merged region is a part of source range.
2. Doesn't take into account possible destination's range X/Y offsets relative to source range.
3. Outright fails when trying to copy ranges within the same sheet.

### DIFF
--- a/poi/src/main/java/org/apache/poi/ss/util/CellRangeAddress.java
+++ b/poi/src/main/java/org/apache/poi/ss/util/CellRangeAddress.java
@@ -133,4 +133,17 @@ public class CellRangeAddress extends CellRangeAddressBase {
         }
         return new CellRangeAddress(a.getRow(), b.getRow(), a.getCol(), b.getCol());
     }
+
+    /**
+     * Shifts cell range by specified number of rows and columns.
+     *
+     * @param rows    rows to shift by.
+     * @param columns columns to shift by.
+     * @return copy of this {@link CellRangeAddress}, shifted by rows and columns.
+     */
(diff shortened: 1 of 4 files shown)


poi/src/main/java/org/apache/poi/ss/util/CellRangeAddress.java
```diff
@@ -133,4 +133,17 @@ public static CellRangeAddress valueOf(String ref) {
         }
         return new CellRangeAddress(a.getRow(), b.getRow(), a.getCol(), b.getCol());
     }
+
+    /**
+     * Shifts cell range by specified number of rows and columns.
+     *
+     * @param rows    rows to shift by.
+     * @param columns columns to shift by.
+     * @return copy of this {@link CellRangeAddress}, shifted by rows and columns.
```
pjfanning, line 142: `@since 6.0.0`

pjfanning: on all new public methods

vo0ff: Thanks, added

pjfanning (approved):
lgtm - merged

PR #969, opened 2025-12-15
Fix rules for table style application in XSLFTables.

jaragunde:
Table styles contain rules for first and last rows/columns, even and odd rows/columns... When one of these rules is empty, we are supposed to fall back to the "whole table" rules. The fallback must also be applied for the format that's not explicitly specified in the specific rules.

when the corresponding specific rule is missing some info.

Fallback must also be applied when the corresponding specific rule is missing some info. A couple of examples from the reproducer/test file included:

The included reproducer/test file contained a few problems related to table style application:
* The second style for horizontal/vertical banding (band2H, band2V) was never applied.
* In the table with horizontal banding enabled, the style band1H did not set a font color, POI returned a null font color instead of the color from wholeTable.
* In the table with horizontal banding enabled, the style band2H did not set a background color, POI returned null instead of the color specified in wholeTable.

This patches fixes the behaviors mentioned above, making POI behavior match the one from MS Office and LibreOffice.

For extra context, this is a screenshot of one of the tables in the attached document, as displayed by PowerPoint:
image

This is the table rendered by POI on a Graphics2D context, before this patch:
image

And this is the render after this patch:
image


jaragunde:
I see the failure due to use of `List.getFirst()` (java 21). I'll fix that.

pjfanning (approved):
lgtm - merged

PR #973, opened 2025-12-17
XDDFDataSourcesFactory.java replace anonymous classes with a real one for clarity

emmanueldufour:
removed some inline classes ( because they are hard to read and debug) and replaced these 4 anon classes with a single real class thereby shrinking code

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSourcesFactory.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSourcesFactory.java
@@ -16,7 +16,6 @@
  *    limitations under the License.
  * ====================================================================
  */
-
 package org.apache.poi.xddf.usermodel.chart;
 
 import org.apache.poi.ss.usermodel.CellType;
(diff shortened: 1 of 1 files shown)


poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSourcesFactory.java
```diff
@@ -439,6 +439,7 @@ enum Type {
 		public final Type type;
 
 		public XDDFCategoryDataSourceImpl(CTAxDataSource categoryDS) {
+			this.categoryDS=categoryDS;
```
pjfanning: can you keep the whitespace?

emmanueldufour: how ?

emmanueldufour: you mean just around this = sign ?

emmanueldufour: done

pjfanning (changes requested):
this file has been reformatted - appears to replace the spaces used for indents with tabs, leading to a 100% diff in the file

I will not review this unless the format is fixed. It is not acceptable to reformat files in PRs with code changes - it makes the PR impossible to review.

poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSourcesFactory.java
```diff
@@ -37,541 +36,483 @@
...
+			}
+			return null;
+		}
+
+		public String getFormatCode() {
+			return formatCode;
+		}
+		
+	    public boolean isNumeric() {
+              return (type == Type.FROM_NUMBER_REFERENCE || type == Type.FROM_NUMBER_LITERAL);
```
pjfanning: code will never be merged unless it is formatted correctly - 4 space indents and no tabs

emmanueldufour: done

emmanueldufour: btw in the future do not let whitespaces hold you back:
https://stackoverflow.com/questions/37007300/how-to-ignore-whitespace-in-github-when-comparing

PR #974, opened 2025-12-17
refresh XWPFChart.java cache from embedded values 

emmanueldufour:
Word keeps cached values that need be refreshed after the embedded workbook has been updated. This is just the first step refreshing categories and  series labels (tested for bar charts). 
see for ex _the related issues_:
 https://stackoverflow.com/questions/74703330/updated-excel-chart-in-word-document-not-getting-reflected-apache-poi
https://stackoverflow.com/questions/54681481/apache-poi-when-editing-chart-data-in-a-word-file-it-returns-to-the-data-defi

TODO ultimately a function should be provided to refresh the entire chart cache
TODO getWorkbookCells and getWorkbookCellVallue should be moved to XSSFWorkbook

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFChart.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFChart.java
@@ -40,6 +40,8 @@ import org.apache.poi.xddf.usermodel.chart.XDDFChartData;
 import org.apache.poi.xddf.usermodel.chart.XDDFDataSource;
 import org.apache.poi.xddf.usermodel.chart.XDDFDataSourcesFactory;
 import org.apache.poi.xddf.usermodel.chart.XDDFNumericalDataSource;
+import org.apache.poi.xssf.usermodel.XSSFSheet;
+import org.apache.poi.xssf.usermodel.XSSFWorkbook;
 import org.apache.xmlbeans.XmlException;
 import org.openxmlformats.schemas.drawingml.x2006.wordprocessingDrawing.CTInline;
 
@@ -70,7 +72,7 @@ public class XWPFChart extends XDDFChart {
      * constructor to
      * Create a new chart in document
      *
-     * @since POI 4.0.0
+     * @since 4.0.0
      */
     protected XWPFChart() {
         super();
@@ -81,7 +83,7 @@ public class XWPFChart extends XDDFChart {
      *
      * @param part the package part holding the chart data,
      *             the content type must be {@code application/vnd.openxmlformats-officedocument.drawingml.chart+xml}
-     * @since POI 4.0.0
+     * @since 4.0.0
      */
     protected XWPFChart(PackagePart part) throws IOException, XmlException {
         super(part);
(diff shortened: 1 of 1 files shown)


pjfanning:
Can you start with using descriptive names for both your PRs?

emmanueldufour:
> Can you start with using descriptive names for both your PRs?

done

pjfanning:
It is against ASF licensing policies to use code from stackoverflow.com due to the restrictive licensing that they apply to code on the site. You need to get the people who wrote the code to place the code in the public domain or license it with a more permissive license.
https://www.apache.org/legal/resolved.html#stackoverflow

emmanueldufour:
> It is against ASF licensing policies to use code from stackoverflow.com due to the restrictive licensing that they apply to code on the site. You need to get the people who wrote the code to place the code in the public domain or license it with a more permissive license. https://www.apache.org/legal/resolved.html#stackoverflow

I did not use code from stackoverflow. I merely quoted related issues this is supposed to help with.

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFChart.java
```diff
@@ -58,7 +70,7 @@ public class XWPFChart extends XDDFChart {
      * constructor to
      * Create a new chart in document
      *
-     * @since 4.0.0
+     * @since POI 4.0.0
```
pjfanning: don't change these values

emmanueldufour: yes sorry I didnt intend to but Im struggling with GIT, if you could revert it that would help

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFChart.java
```diff
@@ -258,12 +270,76 @@ public long getChartRightMargin(long margin) {
...
+	 * @param cell must include the sheet reference
+	 * @param wb
+	 * @return toString() value of the cell
+	 */
+	private static String getWorkbookCellValue(CellReference cell, Workbook wb) {
+		
+		return wb.getSheet( cell.getSheetName()).getRow(cell.getRow()).getCell(cell.getCol()).toString();
+		
+	}
+	/**
```
pjfanning: whitespace between methods - no need for all the whitespace inside the method - please try to make the code follow the existing style in classes

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFChart.java
pjfanning: what can I revert in your PR?

emmanueldufour: since 4.0.0  vs  since POI 4.0.0

pjfanning: I wasn't offering to do it. I was asking why I'm expected to do it.

pjfanning (changes requested):
* no tests
* code incorrectly formatted

PR #981, opened 2025-12-18
fix(xwpf): fix inconsistency between bodyElements and paragraphs (POI-980)

ybg163yx:
### What changed
Fixes an inconsistency between `bodyElements` and `paragraphs` caused by
`XWPFDocument#setParagraph`.

### Why
Since **REL_5_5_1**, `setParagraph` updates the underlying XML by copying
paragraph contents via XMLBeans, while the Java-side `paragraphs` list was
updated by directly replacing the element reference.

This mismatch can lead to an inconsistent internal state, causing
`getParagraphPos` to return `-1` and breaking `removeBodyElement`.

### How
- Update the XML representation using `CTBody#setPArray`
- Re-align the `paragraphs` list with the paragraph instance stored in
  `bodyElements`, which is created from the updated XML
- Ensure both internal collections remain consistent after calling
  `setParagraph`

### Tests
- Verified locally using a DOCX file that fails starting from REL_5_5_1

### Issue
This PR fixes **POI-980**.


pjfanning:
this PR needs to be rebased against the trunk branch

PR #986, opened 2025-12-31
XWPF Themes: allow public access of theme and add helpers for theme fonts

Etienne-Gautier:
# Context
I'd like to better support DOCX theme fonts, but currently they are inaccessible.

For anyone interested, a run can define fonts either
- directly `w:ascii="Times New Roman"`
- indirectly via a theme for example `w:asciiTheme="majorBidi"`

Currently the XWPFTheme doesn't expose major/minor fonts to resolve theme fonts, so I am fixing that in this PR.

Note: I spent a bit of time figuring out how the theme fonts are resolved and although it isn't implemented in this PR I thought I'd write it down somewhere. So for anyone interested, a run might look like this:
```xml
    <w:r>
      <w:rPr>
        <w:rFonts w:asciiTheme="majorBidi" w:hAnsiTheme="majorBidi" w:cstheme="majorBidi"/>
      </w:rPr>
      <w:t>This text uses Times New Roman via themes</w:t>
    </w:r>
```

With settings containing 
```xml
  <w:themeFontLang w:val="en-AU" w:bidi="ar-SA"/>
```

And a Theme like this:

```xml
    <a:fontScheme name="Office">
      <a:majorFont>
        <a:latin typeface="Aptos Display" panose="02110004020202020204"/>
        <a:ea typeface=""/>
        <a:cs typeface="Segoe UI"/>
        <a:font script="Jpan" typeface="游ゴシック Light"/>
        <a:font script="Arab" typeface="Times New Roman"/>
    </a:majorFont>
    <a:minorFont>
        <a:latin typeface="Aptos" panose="02110004020202020204"/>
        <a:ea typeface=""/>
        <a:cs typeface=""/>
        <a:font script="Jpan" typeface="游明朝"/>
        <a:font script="Arab" typeface="Arial"/>
    </a:minorFont>
```

This theme font can be interpreted by combining info from the document settings (`themeFontLang`) and themes. There is an explanation of that in the spec at `17.15.1.88 themeFontLang (Theme Font Languages)`. However that explanation doesn't really tell you how to map the locale code (for example `ar-SA`) from setting's `themeFontLang` to a script (for example `Arab`) used by the theme definition.

I found that:
- POI's `LocaleID.lookupByLanguageTag(localeName)` can read the locale.
- Then the logic from Libreoffice [here](https://github.com/LibreOffice/core/blob/master/sw/source/writerfilter/dmapper/ThemeHandler.cxx#L21) can easily be ported to java to map a `localeID.getLcid()` to a script. We can then lookup the major/minor theme font with that script

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFTheme.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFTheme.java
@@ -29,6 +29,7 @@ import org.openxmlformats.schemas.drawingml.x2006.main.CTColor;
 import org.openxmlformats.schemas.drawingml.x2006.main.CTColorScheme;
 import org.openxmlformats.schemas.drawingml.x2006.main.CTFontCollection;
 import org.openxmlformats.schemas.drawingml.x2006.main.CTOfficeStyleSheet;
+import org.openxmlformats.schemas.drawingml.x2006.main.CTSupplementalFont;
 import org.openxmlformats.schemas.drawingml.x2006.main.ThemeDocument;
 import org.openxmlformats.schemas.wordprocessingml.x2006.main.*;
 
@@ -146,8 +147,10 @@ public class XWPFTheme extends POIXMLDocumentPart {
      *
      */
     @SuppressWarnings("WeakerAccess")
-    public String getMajorFont(){
-        return _theme.getThemeElements().getFontScheme().getMajorFont().getLatin().getTypeface();
+    public String getMajorFont() {
+        CTFontCollection majorFonts = getMajorFonts();
+        return majorFonts == null || majorFonts.getLatin() == null ?
+                null : majorFonts.getLatin().getTypeface();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFTheme.java
```diff
@@ -67,6 +68,42 @@ public void importTheme(XSLFTheme theme) {
         _theme = theme.getXmlObject();
     }
 
+    public CTOfficeStyleSheet getCTOfficeStyleSheet() {
```
pjfanning, line 76: can you add javadoc with at least `@since 6.0.0` on new public methods?

poi-ooxml/src/main/java/org/apache/poi/xwpf/usermodel/XWPFTheme.java
```diff
@@ -67,6 +68,42 @@ public void importTheme(XSLFTheme theme) {
...
+        return _theme;
+    }
+
+    public String getMajorFontForScript(String script){
```
pjfanning: any chance of adding unit test coverage?

Etienne-Gautier:
@pjfanning I've added some javadoc and basic tests. Thanks for the quick review

PR #990, opened 2026-01-08
Support rendering transparent bitmaps in presentations.

jaragunde:
Add PictureShape.getAlpha() method and implementations for HSLF and XSLF. Then make use of it in DrawPictureShape to apply the right alpha value to the picture being drawn.

Fixed a bug in BitmapImageRenderer that considered alpha value 0 as "fully opaque", when it means "fully transparent" instead.

Finally, added a test for this feature in TestDrawPictureShape for XSLF. A test for HSLF could not be created because it was not possible to generate a test file with today's tools; MS Office removes the bitmap transparency effect when saving as .ppt, and LibreOffice blends it into the bitmap.


poi/src/main/java/org/apache/poi/sl/usermodel/PictureShape.java
```diff
@@ -47,4 +47,6 @@ public interface PictureShape<
      * @return the clipping rectangle, which is given in percent in relation to the image width/height
      */
     Insets getClipping();
+
+    int getAlpha();
```
pjfanning, line 60: add javadoc including `@since 6.0.0`

poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFPictureShape.java
```diff
@@ -225,6 +225,10 @@ public Insets getClipping(){
             POIXMLUnits.parsePercent(r.xgetR()));
     }
 
+    public int getAlpha() {
+        return getBlip().sizeOfAlphaModFixArray() > 0 ? POIXMLUnits.parsePercent(getBlip().getAlphaModFixArray(0).xgetAmt()) : 100000;
```
pjfanning: why 100000? can this be documented?

jaragunde: Good question, I first found this in the code that applies alpha to bitmap fill in shapes: https://github.com/apache/poi/blob/fd66689944132934845f177bdd52b17822f600d1/poi/src/main/java/org/apache/poi/sl/draw/DrawPaint.java#L279

It obtains the alpha value from TexturePaint which expresses it in "100k percentage". I basically copied the getAlpha implementation from XSLFTexturePaint: https://github.com/apache/poi/blob/6670c48c305fca2aa30c4bdd01763c696f8f698e/poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFTexturePaint.java#L105

Opening up an OOXML file, alpha information is stored like this: `<a:alphaModFix amt="80000"/>`. This example is 80%. I guess it all comes down to "we do what OOXML does".

I've just taken a look at the standard, and I understood it's also possible to express it like a percentage string, like "80%" but I haven't found documents using it. I will forge one and come back.

jaragunde: Reporting back, I tested what happens when the document expresses the alpha value like this: `<a:alphaModFix amt="80%"/>`.

In that case, POIXMLUnits detects it ends with % and returns the integer value multiplied by 1000: https://github.com/apache/poi/blob/cc8b6d7591095c4b7e5185541bafcaf05e9ba50a/poi-ooxml/src/main/java/org/apache/poi/ooxml/util/POIXMLUnits.java#L144

So the returned alpha value is 80000 and the document is processed correctly. This is documented in `POIXMLUnits.parsePercent()`:
```
    /**
     * Office will read percentages formatted with a trailing percent sign or formatted
     * as 1000th of a percent without a trailing percent sign
     *
     * @return the percent scaled by 1000, so 100% = 100000
     */
 ```
The document is also correct for MS Office, displaying the expected alpha value. Interestingly, LibreOffice doesn't seem to interpret it correctly, and it sets alpha to 0 (transparent).

I added a comment clarifying this in `XSLFPictureShape`.

Comment (pjfanning) [reply poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFPictureShape.java]: It is not a percentage if its value is typically more than 100. We need to avoid the phrase 'percent'.

**Comment by pjfanning** [comment]
causes an NPE with one file.

```
        Test #1281 slideshow/missing-blip-fill.pptx XSLF FAILED

        org.opentest4j.AssertionFailedError: slideshow/missing-blip-fill.pptx - failed for handler XSLFFileHandler:  ==> Unexpected exception thrown: java.lang.NullPointerException
            at app/org.apache.poi.stress@6.0.0-SNAPSHOT-tests/org.apache.poi.stress.TestAllFiles.verify(TestAllFiles.java:317)
        Caused by: java.lang.NullPointerException
            at org.apache.poi.stress@6.0.0-SNAPSHOT-tests/org.apache.poi.stress.TestAllFiles.lambda$handleFile$1(TestAllFiles.java:242)
```

File: poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFPictureShape.java (hunk 2 above)
Comment (jaragunde) [reply poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFPictureShape.java]: I changed the wording of the comments I wrote. Instead of percentage, I call it "range between 0 and 100000".

Unfortunately, the use of "percentage scaled by 1000" is everywhere, in the method `parsePercent()` for example, and even the XML bean `STPositivePercentage` may either return a percentage in the hundreds, ended in `%`, or a percentage multipled by 1000.

**Comment by jaragunde** [comment]
Sent more commits, to address the NPE and the comments regarding the usage of "percentage" in the wording.

File: poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFPictureShape.java [hunk 3]
```diff
@@ -225,6 +225,16 @@ public Insets getClipping(){
             POIXMLUnits.parsePercent(r.xgetR()));
     }
 
+    public int getAlpha() {
+        // If not specified, we return 100000, which means fully opaque.
+        int FULLY_OPAQUE = 100000;
```
pjfanning: Can this be made a final static var?

poi/src/main/java/org/apache/poi/sl/draw/DrawPictureShape.java
```diff
@@ -66,6 +67,9 @@ public void drawContent(Graphics2D graphics) {
                 ImageRenderer renderer = getImageRenderer(graphics, ct);
                 if (renderer.canRender(ct)) {
                     renderer.loadImage(dataBytes, ct);
+                    if (0 <= alpha && alpha < 100000) {
+                        renderer.setAlpha(alpha/100000.f);
```
pjfanning: Can this use a static var with the 100000 value?

pjfanning (changes requested):

jaragunde:
Thanks for your reviews! I created static final variables for 0 and 100000 alpha values.

pjfanning (approved):

pjfanning:
merged - thanks

Issue #919, opened 2025-10-17
BUG: Shape with empty paragraph & textrun generates invalid XML


sascha08-15:
POI creates invalid pptx files if we call .clearText and do not create a paragraph and textrun in it. The later is the workaround; however, this IS a bug. 

Test to reproduce below, just open the created file in PowerPoint (on Mac) it will complain about invalid xml.

```import org.apache.poi.xslf.usermodel.*;
import org.apache.poi.sl.usermodel.ShapeType;
import org.junit.jupiter.api.Test;

import java.awt.geom.Rectangle2D;
import java.io.FileInputStream;
import java.io.FileOutputStream;
import java.nio.file.Files;
import java.nio.file.Path;

import static org.junit.jupiter.api.Assertions.*;

public class PoiClearTextPptxTest {

    @Test
    public void createShapeClearTextAndSave_pptxShouldBeReportedInvalidByPowerPoint() throws Exception {
        // create a new presentation
        try (XMLSlideShow ppt = new XMLSlideShow()) {
            // create a slide
            XSLFSlide slide = ppt.createSlide();

            // create a rectangle auto-shape and position it
            XSLFAutoShape rect = slide.createAutoShape();
            rect.setShapeType(ShapeType.RECT);
            rect.setAnchor(new Rectangle2D.Double(100, 100, 400, 120));

            // add text to the shape
            XSLFTextParagraph p = rect.addNewTextParagraph();
            XSLFTextRun run = p.addNewTextRun();
            run.setText("This is a test text that will be cleared.");

            // sanity-check text is present
            assertTrue(rect.getText().contains("This is a test text"));

            // THIS IS THE KEY CALL: remove all text content from the shape
            rect.clearText();

            // After clearText(), POI leaves the shape without text runs/paragraphs in its model.
            // That can produce a PPTX that PowerPoint flags as "needs repair" / invalid.

            // write to a temporary file
            Path out = Files.createTempFile("poi-cleartext-repro-", ".pptx");
            try (FileOutputStream fos = new FileOutputStream(out.toFile())) {
                ppt.write(fos);
            }

            // Basic checks
            assertTrue(Files.exists(out));
            assertTrue(Files.size(out) > 0);

            System.out.println("Wrote test file to: " + out.toAbsolutePath());

            // Optional: try to re-open with POI (may succeed even if PowerPoint complains)
            try (FileInputStream fis = new FileInputStream(out.toFile());
[...]

PR #994, opened 2026-01-13
Fix Bug 69855: Ensure XSLFTextShape.clearText() maintains at least one paragraph

joonda:
#919 

`clearText()` removed all paragraphs from the text body, which violates the OOXML schema.
because txBody must contain at least one paragraph.

This commit modifies clearText() to ensure a single empty paragraph remains,
maintaining schema compliance and preventing "needs repair" errors in PowerPoint.

Add a test code to verify the fix and prevent future regressions.


poi-ooxml/src/test/java/org/apache/poi/xslf/usermodel/TestXSLFTextShape.java
```diff
@@ -982,4 +984,36 @@ void metroBlob() throws IOException, ReflectiveOperationException {
...
+
+            XSLFTextParagraph p = shape.addNewTextParagraph();
+            p.addNewTextRun().setText("This text will be cleared");
+
+            shape.clearText();
+
+            assertEquals(1, shape.getTextParagraphs().size(),
+              "After clearText(), there should be exactly one empty paragraph for OOXML compliance.");
+
+            try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
```
pjfanning: use commons-io UnsynchronizedByteArrayOutputStream which has a toInputStream method - like we already use in poi

joonda: i've updated the test to use `UnsynchronizedByteArrayOutputStream`.
Thank you for the review!

joonda:
It seems that the Java CI with Gradle / build (pull_request) is failing.
Would it be okay if I fix this and update the PR?

PR #999, opened 2026-01-17
Fix Bug 69855 bug: Ensure at least one paragraph for OOXML compliance

joonda:
#919 

**Description**
Currently `XSLFTextShape.clearText()` removes all paragraphs, leaving the `txBody` empty.

according to the OOXML specification, a text body must contain at least one paragraph. 
This inconsistency causes PPT to report the file as corrupted when opening a PPTX file generated after calling `clearText()`.

**Changes**
Modified `clearText()` to ensure at least one empty paragraph is maintained using `addNewTextParagraph()`.

**Why update existing tests?**
Since a mandatory empty paragraph is now inserted at Index 0, subsequent content is appended at Index 1.
I have updated the assertions in the exisiting tests to reflect this required structural change for XML compliance.

**Additional Notes**
I am resubmitting this Pull Request because my previous submission was incomplete. This PR provides the correct fix and includes all necessary test adjustments. I apologize for the previous confusion.

### DIFF
--- a/poi-ooxml/src/test/java/org/apache/poi/xslf/TestXSLFBugs.java
+++ b/poi-ooxml/src/test/java/org/apache/poi/xslf/TestXSLFBugs.java
@@ -236,7 +236,7 @@ class TestXSLFBugs {
             try (XMLSlideShow ppt3 = writeOutAndReadBack(dest)) {
                 XSLFSlide slide3 = ppt3.getSlides().get(0);
                 XSLFTextBox shape3 = (XSLFTextBox) slide3.getShapes().get(0);
-                XSLFTextParagraph p3 = shape3.getTextParagraphs().get(1);
+                XSLFTextParagraph p3 = shape3.getTextParagraphs().get(2);
                 XSLFHyperlink h1 = p3.getTextRuns().get(0).getHyperlink();
                 assertNotNull(h1);
                 assertEquals("https://poi.apache.org", h1.getAddress());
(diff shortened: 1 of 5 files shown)


poi-ooxml/src/test/java/org/apache/poi/xslf/TestXSLFBugs.java
```diff
@@ -236,7 +236,7 @@ void bug61589() throws IOException {
             try (XMLSlideShow ppt3 = writeOutAndReadBack(dest)) {
                 XSLFSlide slide3 = ppt3.getSlides().get(0);
                 XSLFTextBox shape3 = (XSLFTextBox) slide3.getShapes().get(0);
-                XSLFTextParagraph p3 = shape3.getTextParagraphs().get(1);
+                XSLFTextParagraph p3 = shape3.getTextParagraphs().get(2);
```
pjfanning, line 239: Why is there is an extra paragraph here? the test doesn't seem to manipulate the text.
If we are going to modify the behaviour, we should check the 0 and 1 paragraphs to see that we don't duplicate the paragraphs by accident.

I'm not convinced that the other test changes are valid either. It seems like we are hacking the tests instead of improving the tests.

joonda: Thank you for your thoughtful feedback.
I completely agree that my previous approach of modifying the test indices was not the ideal way to solve this.

I'll close this PR for now to explore a cleaner solution that ensures OOXML compliance without affecting the existing logic.

I appreciate your patience and guidance.

PR #1006, opened 2026-02-09
Bug 69935: Add short-based column setters to ClientAnchor

jogakdal:
This PR addresses [Bug 69935](https://bz.apache.org/bugzilla/show_bug.cgi?id=69935),
where `ClientAnchor` exposes `short`-returning getters for the column indices (`getCol1()`, `getCol2()`) but only `int`-based setters (`setCol1(int)`, `setCol2(int)`).

This mismatch makes it difficult to use Kotlin property syntax, which expects matching getter/setter types for a property.

### What is changed

- Add default short-based setters on `ClientAnchor`:
  - `default void setCol1(short col1)` delegating to `setCol1(int)`.
  - `default void setCol2(short col2)` delegating to `setCol2(int)`.

  This keeps the existing `int`-based setters intact and binary compatible, while exposing a `short`-typed getter/setter pair that works better with Kotlin properties.

- Add `TestClientAnchorShortSetter` to cover:
  - The new short-based setters delegate correctly to the existing int-based setters.
  - The existing int-based setters still behave as before.

### Notes

- The `@since` tag on the new methods is left as `TBD` and can be adjusted by the committer to the appropriate release version.

### Testing

Locally ran:

```bash
./gradlew :poi:test --tests 'org.apache.poi.ss.usermodel.TestClientAnchorShortSetter'

### DIFF
--- a/poi/src/test/java/org/apache/poi/ss/usermodel/TestClientAnchorShortSetter.java
+++ b/poi/src/test/java/org/apache/poi/ss/usermodel/TestClientAnchorShortSetter.java
@@ -1,3 +1,19 @@
+/* ====================================================================
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
+   The ASF licenses this file to You under the Apache License, Version 2.0
+   (the "License"); you may not use this file except in compliance with
+   the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by pjfanning** [comment]
next release is 6.0.0 - could you replace the TBD?

File: poi/src/test/java/org/apache/poi/ss/usermodel/TestClientAnchorShortSetter.java [hunk 1]
```diff
@@ -0,0 +1,50 @@
+package org.apache.poi.ss.usermodel;
```
pjfanning, line 17: can you add a source header like those that exist in our other POI classes?

jogakdal:
Thanks for the review.

Added the standard ASF license header to `TestClientAnchorShortSetter` to match other POI source files.

pjfanning (approved):
lgtm - merged

PR #1020, opened 2026-03-03
Add fuzz targets for OSS-Fuzz integration (RLE + FormulaParser)

vishalcoc44:
Hi @centic9,

I'm contributing two fuzz targets for integration with Google's OSS-Fuzz continuous fuzzing platform. The OSS-Fuzz maintainers have requested that they be upstreamed.

1. POIRleFuzzer for RLEDecompressingInputStream
   - Fuzz Introspector showed 0% static reachability for this class.
   - It found an IllegalStateException ("Not enough bytes") caused by malformed RLE chunks.
   - Bug report: https://bz.apache.org/bugzilla/show_bug.cgi?id=69956
   - Related OSS-Fuzz PR: https://github.com/google/oss-fuzz/pull/14971

2. FormulaParserFuzzer for FormulaParser.parse()
   - Targets the recursive-descent formula parser, which had low fuzz coverage.
   - Found that malformed nested brackets (`[[`) cause a RuntimeException ("Parsed past the end of the formula") instead of a proper FormulaParseException (confirmed in POI 5.2.3).
   - Includes a dictionary file with Excel formula tokens to help guide the fuzzer.
   - Reproduction: `FormulaParser.parse("[[", workbook, FormulaType.CELL, -1)`
   - Related OSS-Fuzz PR: https://github.com/google/oss-fuzz/pull/14977

Thanks for taking the time to review.


pjfanning (changes requested):
These 2 files need Apache source headers like the ones in existing files in this repo. The Google copyright in one file. Is this needed? Provide full details if yes.

If you we do need a Google copyright then you need to also update the LICENSE file in this repo to mention this file with Google copyright code.

poi/src/test/java/org/apache/poi/fuzz/FormulaParserFuzzer.dict
```diff
@@ -0,0 +1,48 @@
+# Excel Formula Dictionary for FormulaParserFuzzer
```
pjfanning: Move this 2 files in their own package, maybe with fuzz in the package name.

pjfanning (dismissed):

vishalcoc44:
> These 2 files need Apache source headers like the ones in existing files in this repo. The Google copyright in one file. Is this needed? Provide full details if yes.
> 
> If you we do need a Google copyright then you need to also update the LICENSE file in this repo to mention this file with Google copyright code.

Thanks @pjfanning! I've addressed both points: 
1. Replaced Google copyright headers with standard Apache ASF headers.
2. Moved all fuzz files to their own package: org.apache.poi.fuzz. Please take another look!

vishalcoc44:
Hey @pjfanning  there were some build errors due to the jazzer api . I've made the required changes.
- Add com.code-intelligence:jazzer-api dependency to poi/build.gradle
- Update poi/src/test/java9/module-info.java for JPMS compatibility
- Configure Ant build.xml to download and include jazzer-api in test classpath
- Add **/*.dict to Apache RAT excludes in root build.gradle to fix license audit

poi/src/test/java9/module-info.java
```diff
@@ -111,6 +111,7 @@ Licensed to the Apache Software Foundation (ASF) under one or more
     requires org.junit.jupiter.api;
     requires org.junit.jupiter.params;
     requires org.mockito;
+    requires static com.code_intelligence.jazzer.api;
```
pjfanning: let's not do this - move the code out of here into its own module if this is needed

pjfanning (changes requested):

poi/src/test/java9/module-info.java
pjfanning: Can't you even host this change yourself? Why does this code need to be in POI repo. It can be in any arbitrary Git repo.

pjfanning:
If you want to proceed with this PR in this repo, don't update the Ant build. It is already a millstone around our necks and it doesn't need to support everything. Don't break the Ant build but if we add a new poi-fuzz module then we don't need the Ant build to support it. Concentrate on the Gradle build.

poi/src/test/java/org/apache/poi/fuzz/FormulaParserFuzzer.dict
pjfanning: Why can't this file have a license header? Much prefer this to have a license header than hacking the RAT excludes.

poi-fuzz/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+dependencies {
+    implementation project(':poi')
+
+    compileOnly 'com.code-intelligence:jazzer-api:0.22.0'
+}
+
+// Fuzz targets are not standard JUnit tests; disable the test task.
+test.enabled = false
+
+javadoc.enabled = false
```
pjfanning, line 31: Do you need a module-info.java for this module? Don't add one unless you need one but since you modified the module-info in poi module earlier, I thought I'd ask.

vishalcoc44: No we don't. Since we moved the fuzz files to its own separate submodule, the reason dissapeared, so we reverted the module-info.java

centic9:
Thanks for the PR, I have now merged this so I can start moving over the fuzz-targets from oss-fuzz. Only when we switch to this source-location in oss-fuzz, these 2 fuzz-targets will be actually included in fuzzing via oss-fuzz.

vishalcoc44:
> Thanks for the PR, I have now merged this so I can start moving over the fuzz-targets from oss-fuzz. Only when we switch to this source-location in oss-fuzz, these 2 fuzz-targets will be actually included in fuzzing via oss-fuzz.

Thanks, Let me know if i could do anything!

PR #1025, opened 2026-03-05
Add WorkbookEvaluator context-aware fuzz target

vishalcoc44:
This PR adds a context-aware fuzz target for the Apache POI `WorkbookEvaluator` engine to the `poi-fuzz` module. 

### Technical Approach
- **Context-Aware Strategy**: Pre-populates a **20×20 Excel data grid** with a diverse mix of numbers, negative values, strings, booleans, and error codes so that referential formulas (SUM, VLOOKUP, INDEX, MATCH, etc.) exercise real evaluation paths instead of fast-failing on empty references.
- **Guided Input Generation**: Uses a dictionary of **60 complex Excel function signatures** to steer the fuzzer toward deep calculation branches in the `WorkbookEvaluator`.
- **Coverage Impact**: In local testing, this increased code coverage within the `WorkbookEvaluator` from **19 lines** to over **2,200 lines** (~11,700% increase).
- **Security Alignment**: Expected exceptions (`FormulaParseException`, `NotImplementedException`, etc.) are suppressed while internal logic flaws (NPE, OOB) bubble up as findings.

2. PPTX2PNG (Slideshow Rendering Fuzzing)

- **Target:** Exercises the java.awt drawing subsystem by rendering slides into a BufferedImage.
- **Rationale:** The rendering pipeline is a high-complexity path prone to Resource Exhaustion (CPU/Memory) that was previously untested in OSS-Fuzz.
- **Resiliency:** Per feedback from Apache POI developers regarding the threat model ("Expect any type of Exception"), this fuzzer broadly catches and ignores Exception and Error types. This ensures the fuzzer remains focused strictly on identifying inputs that lead to memory or CPU exhaustion.

### Placement
`poi-fuzz/src/main/java/org/apache/poi/fuzz/WorkbookEvaluatorFuzzer.java`
`poi-fuzz/src/main/java/org/apache/poi/fuzz/POIPPTX2PNGFuzzer.java`


vishalcoc44:
@centic9

poi-fuzz/src/main/java/org/apache/poi/fuzz/WorkbookEvaluatorFuzzer.java
```diff
@@ -0,0 +1,122 @@
...
+ * setting fuzz-generated formulas on a pre-populated workbook and
+ * evaluating them through the {@link WorkbookEvaluator}. A 20x20 data
+ * grid with mixed cell types (numbers, strings, booleans, errors)
+ * ensures that referential functions (SUM, VLOOKUP, INDEX, etc.)
+ * reach deep evaluation branches.
+ * <p>
+ * Used by Google's OSS-Fuzz for continuous security testing.
+ */
+public class WorkbookEvaluatorFuzzer {
+    private static HSSFWorkbook workbook;
```
centic9: Keeping those static is not ideal as it will keep state across fuzz-calls and thus crashes/findings may not be reproducible.

It seems the same was already used in FormulaParserFuzzer, but we should avoid it to keep fuzzers independent from previous invocations.

vishalcoc44: @centic9 alright, i have refactord both the files to remove all static state

vishalcoc44:
@centic9

centic9-dt:
_FYI, I am working on Apache POI in my free time without payment and thus will usually not respond immediately on PR-updates. Please do not ping me repeatedly on PRs._ 

In general we should try to NOT add more and more fuzz-targets as oss-fuzz has quite some overhead per fuzz-target, e.g. fuzz-introspection runs for a very long time and each new target increases runtime considerably, leading to timeouts and failed runs.

We will need to actually look at ways to combine some fuzzers or somehow else speed up fuzz-introspection runs at oss-fuzz considerably.

Maybe we can get first rid of the POI(H|X)...Fuzzer classes as POIFileHandlerFuzzer already covers all formats in one go? Downside is that more specific fuzzing is still more effective... ?!?

vishalcoc44:
> _FYI, I am working on Apache POI in my free time without payment and thus will usually not respond immediately on PR-updates. Please do not ping me repeatedly on PRs._
> 
> In general we should try to NOT add more and more fuzz-targets as oss-fuzz has quite some overhead per fuzz-target, e.g. fuzz-introspection runs for a very long time and each new target increases runtime considerably, leading to timeouts and failed runs.
> 
> We will need to actually look at ways to combine some fuzzers or somehow else speed up fuzz-introspection runs at oss-fuzz considerably.
> 
> Maybe we can get first rid of the POI(H|X)...Fuzzer classes as POIFileHandlerFuzzer already covers all formats in one go? Downside is that more specific fuzzing is still more effective... ?!?

Reduced overhead by removing 15 format-specific fuzzers that are already covered by POIFileHandlerFuzzer. Keep only essential fuzzers:
- POIFileHandlerFuzzer (handles all formats)
- POIFuzzer
- WorkbookEvaluatorFuzzer
- FormulaParserFuzzer
- POIRleFuzzer
- EncryptDecryptFuzzer
- FuzzerRunner (new base class)

PR #1027, opened 2026-03-06
Use zip compressed size from entry when available.

jaragunde:
This provides more accurate results that prevent false positives when checking the compress ratio of every entry against MIN_INFLATE_RATIO.

Fixes: https://bz.apache.org/bugzilla/show_bug.cgi?id=69867


poi-ooxml/src/test/java/org/apache/poi/openxml4j/opc/TestPackage.java
```diff
@@ -798,6 +798,15 @@ private void openXmlBombFile(String file) throws IOException {
         }
     }
 
+    @Test
+    void testFakeZipBomb() throws IOException, InvalidFormatException {
+        String file = getSampleFileName("fake-zipbomb-reproducer.docx");
+        try (POITextExtractor extractor = ExtractorFactory.createExtractor(new File(file))) {
+            assertNotNull(extractor);
+            extractor.getText();
```
pjfanning, line 806: maybe assertNotNull here too

poi-ooxml/src/main/java/org/apache/poi/openxml4j/util/ZipArchiveThresholdInputStream.java
```diff
@@ -120,6 +121,17 @@ private void checkThreshold() throws IOException {
             rawSize = 0;
         }
 
+        // If available, read compressed size via ZipArchiveEntry.getCompressedSize() which takes
+        // data from the Zip central directory header. This is because the ZipArchiveInputStream
+        // "may return unknown sizes and CRC values for entries until the next entry has been
+        // reached if the archive uses the data descriptor feature."
+        if (entry != null) {
```
pjfanning, line 128: I don't like this - I prefer to get the real size from that stats object as opposed to trusting a possibly spoofed size value in the zip file.

I don't want a CVE and we will get one if this is merged.

pjfanning: I'm not against doing an additional check with this value but ignored the stats object is a no-no for me.

pjfanning:
stress.xls is no more and this PR is merge conflicted as a result

pjfanning:
If you think the commons-compress stats gathering is incorrect, then I would prefer if this is taken to the commons-compress team. If they can be convinced that there is a problem, then I might be convinced that POI should make changes.

jaragunde:
First of all, thank you for your thorough review.

I don't think there's anything incorrect in commons-compress, other than an inherent and documented limitation of stream access, that happens "if the archive uses the data descriptor feature", which we do (`allowStoredEntriesWithDataDescriptor==true`).

I found an example of a false positive due to this limitation, and I wanted to have the code use (what I thought was) a more accurate value provided by the zip central header, which would work for this false positive. But the situation I put ourselves is having to choose between stream size figures that we know are inaccurate, and header figures that may be spoofed. The lesser evil seems to be the former, spoofing a header is easier. 

All in all, I would rather take back my change. If we get back to this issue and manage to come up with a more robust zip bomb detection method, we would certainly contribute it!

Great timing for me to update stress.xls, it's been there for years :) 

Thanks again!

PR #1028, opened 2026-03-08
[fuzz] Integrate CIFuzz GitHub Action for continuous security testing

vishalcoc44:
I'm adding **CIFuzz** GitHub Action to automatically fuzz new contributions and prevent regressions.

- **Continuous Security**: Every PR will now be fuzzed for 10 minutes using the existing OSS-Fuzz infrastructure.
- **Ideal Integration**: This completes the "Ideal Integration" requirement for Google OSS-Fuzz, ensuring long-term maintenance of the fuzzer health.
- **Regression Testing**: Identified bugs (like the RLE decompression crash) will be caught automatically if re-introduced.

### Validation
- The workflow correctly references the `apache-poi` project on OSS-Fuzz.
- It triggers on Pull Requests targeting `trunk` and `5.5.x`.

@centic9
@pjfanning

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,29 @@
+name: CIFuzz
+on:
+  pull_request:
+    branches: [ trunk, 5.5.x ]
+permissions: {}
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'apache-poi'
+        dry-run: false
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'apache-poi'
+        fuzz-seconds: 600
+        dry-run: false
+    - name: Upload Crash
+      uses: actions/upload-artifact@v4
+      if: failure() && steps.build.outcome == 'success'
+      with:
+        name: artifacts
+        path: ./out/artifacts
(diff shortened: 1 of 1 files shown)


pjfanning:
I'm -1 on this.

I'm really concerned that GitHub will eventually limit ASF projects.

I don't see why we need to fuzz every commit or PR. A complete waste of electricity, water, human time analysing the results.

vishalcoc44:
> I'm -1 on this.
> 
> I'm really concerned that GitHub will eventually limit ASF projects.
> 
> I don't see why we need to fuzz every commit or PR. A complete waste of electricity, water, human time analysing the results.



> I'm -1 on this.
> 
> I'm really concerned that GitHub will eventually limit ASF projects.
> 
> I don't see why we need to fuzz every commit or PR. A complete waste of electricity, water, human time analysing the results.

Hi @pjfanning,

Thank you for your feedback, I do understand your concerns regarding CI resource usage and environmental impact.
Proposed Compromise: Scoped Automated Fuzzing Instead of running on every commit, the 
workflow has been updated to use strictly scoped path filtering. It will now only trigger if changes are made to core logic in:
poi/src/main/java/**
poi-ooxml/src/main/java/**
poi-scratchpad/src/main/java/**
poi-fuzz/src/main/java/**
so changes to documentation, tests, build scripts (Gradle/Ant), and examples will not trigger the fuzzer.

Additionally:
Strict 10-minute cap: The run is truncated to exactly 600 seconds. It is just  a smoke test to catch regressions early, not a long-running process, so clfuzz reports crashes only during that specific pr
Low Impact: This is a small fraction of the total CI time compared to the full build and test suite, but provides a critical security safety net.

we can still reduce the scope further if needed.

pjfanning:
Change the workflow to workflow_dispatch only and I'll merge it but running on every PR is crazy afaic.

vishalcoc44:
> Change the workflow to workflow_dispatch only and I'll merge it but running on every PR is crazy afaic.

done

pjfanning (changes requested):
Not what I asked for. No PR builds. No merge.

pjfanning (approved):
Ok. Diff was not displaying well. Actually seems ok now. Will merge.

Issue #1032, opened 2026-03-19
IndexOutOfBoundsException in XSSFTextParagraph.getBulletAutoNumberScheme() due to hardcoded Enum index access


amos-wnjsoft:
Issue Description:
In the current trunk version of XSSFTextParagraph.java, the method getBulletAutoNumberScheme() performs a direct array access using an integer value from the XML beans without boundary validation.

When an Excel file is created with a newer or unsupported numbering scheme (e.g., a type value that exceeds the pre-defined ListAutoNumber enum length), the application crashes with an ArrayIndexOutOfBoundsException.

Affected Code:
poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFTextParagraph.java
Line: (Approx. line 512 in current trunk)

```Java
public ListAutoNumber getBulletAutoNumberScheme() {
    ParagraphPropertyFetcher<ListAutoNumber> fetcher = new ParagraphPropertyFetcher<ListAutoNumber>(getLevel()){
        public boolean fetch(CTTextParagraphProperties props){
            if(props.isSetBuAutoNum() && props.getBuAutoNum().getType() != null) {
                // DANGEROUS LINE:
                setValue(ListAutoNumber.values()[props.getBuAutoNum().getType().intValue() - 1]);
                return true;
            }
            return false;
        }
    };
    // ...
}
```
Steps to Reproduce:
Create an Excel file containing a text box with a specific numbering/bullet format (e.g., newer Office 365 numbering schemes).

Load the file using XSSFWorkbook.

Iterate through XSSFTextParagraph and call getBulletAutoNumberScheme().

Observe java.lang.ArrayIndexOutOfBoundsException: Index 23 out of bounds for length 19.

Suggested Fix:
Add a boundary check or use a mapping method that defaults to ListAutoNumber.ARABIC_PLAIN if the index is out of range.

```Java
int typeIdx = props.getBuAutoNum().getType().intValue() - 1;
ListAutoNumber[] values = ListAutoNumber.values();
if (typeIdx >= 0 && typeIdx < values.length) {
    setValue(values[typeIdx]);
} else {
    setValue(ListAutoNumber.ARABIC_PLAIN);
}
```

pjfanning:
@amos-wnjsoft would you be interested in creating a PR with a fix and regression test?

amos-wnjsoft:
[...]

PR #1033, opened 2026-03-20
Fix ArrayIndexOutOfBoundsException in XSSFTextParagraph when auto number scheme index is out of range

amos-wnjsoft:
fix issue #1032

### DIFF
--- a/poi-ooxml/src/test/java/org/apache/poi/xssf/usermodel/TestXSSFTextParagraph.java
+++ b/poi-ooxml/src/test/java/org/apache/poi/xssf/usermodel/TestXSSFTextParagraph.java
@@ -18,6 +18,12 @@ package org.apache.poi.xssf.usermodel;
 
 import static org.junit.jupiter.api.Assertions.*;
 
+import org.openxmlformats.schemas.drawingml.x2006.main.STTextAutonumberScheme;
+import org.openxmlformats.schemas.drawingml.x2006.main.CTTextParagraphProperties;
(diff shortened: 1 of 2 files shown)


poi-ooxml/src/test/java/org/apache/poi/xssf/usermodel/TestXSSFTextParagraph.java
```diff
@@ -224,4 +230,37 @@ void testXSSFTextParagraph2() throws IOException {
...
+            bullet.setType(STTextAutonumberScheme.Enum.forInt(24));
+
+            // 4. Execute the test
+            // If you haven't applied the fix, this line will throw AIOOBE
+            // If the fix is applied, this should return ARABIC_PLAIN
+            ListAutoNumber result = para.getBulletAutoNumberScheme();
+
+            assertNotNull(result, "Result should not be null");
+            assertEquals(ListAutoNumber.ARABIC_PLAIN, result, "Should fallback to ARABIC_PLAIN for unknown index");
+        } catch (Exception e) {
```
pjfanning: please remove the catch - it is much better to have the full exception stack trace if an exception is thrown

amos-wnjsoft: Done. I've updated the test case to throw exceptions directly and removed the catch block. Thanks!

pjfanning:
thanks - merged

PR #1042, opened 2026-04-07
Open up xssfb to simplify client usage/modifications.

tballison:
This makes customization of behavior easier for advanced users.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBCellRange.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBCellRange.java
@@ -23,7 +23,7 @@ import org.apache.poi.util.LittleEndian;
 import org.apache.poi.util.LittleEndianConsts;
 
 /**
- * @since 3.16-beta3
+ * @since 6.0.0
  */
 @Internal
 public class XSSFBCellRange {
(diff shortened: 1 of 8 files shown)


poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBCellRange.java
```diff
@@ -26,17 +26,32 @@ Licensed to the Apache Software Foundation (ASF) under one or more
  * @since 3.16-beta3
```
pjfanning: Could you change this `@since 6.0.0`? Since it was not public, it made little sense to have a since version but making it public makes it important to document when it is available since.

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBRichStr.java
```diff
@@ -23,7 +23,7 @@ Licensed to the Apache Software Foundation (ASF) under one or more
  * @since 3.16-beta3
```
pjfanning: Could you change this `@since 6.0.0`? Since it was not public, it made little sense to have a since version but making it public makes it important to document when it is available since.

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBCommentsTable.java
```diff
@@ -63,9 +63,8 @@ public void handleRecord(int id, byte[] data) throws XSSFBParseException {
                 int offset = 0;
                 authorId = XSSFBUtils.castToInt(LittleEndian.getUInt(data)); offset += LittleEndianConsts.INT_SIZE;
                 cellRange = XSSFBCellRange.parse(data, offset, cellRange);
-                offset+= XSSFBCellRange.length;
```
pjfanning, line 66: removed because it is not used after this point?

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBUtils.java
```diff
@@ -85,14 +85,14 @@ public static int readXLWideString(byte[] data, int offset, StringBuilder sb) th
         return numBytes;
     }
 
-    static int castToInt(long val) {
+    public static int castToInt(long val) {
```
pjfanning: Do we really need to make this public? JDK has better methods to safely convert long to int. Math.toIntExact  - for instance.
Likewise castToShort.

tballison:
Thank you @pjfanning! Let me know how it looks now.

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBUtils.java
```diff
@@ -38,7 +37,7 @@ public class XSSFBUtils {
      * @return number of bytes read
      * @throws XSSFBParseException if there was an exception during reading
      */
-    static int readXLNullableWideString(byte[] data, int offset, StringBuilder sb) throws XSSFBParseException {
+    public static int readXLNullableWideString(byte[] data, int offset, StringBuilder sb) throws XSSFBParseException {
```
pjfanning: I can see why the other classes could be made more accessible. Is there a reason to open this util up?

tballison: On Tika, we now have a modified copy of XSSFBHyperlinksTable. For now, we've inlined this call. We don't _need_ this to be public. I'll revert.

pjfanning: I'm ok with it being public if Tika uses it. Maybe we could document in the code that that is the reason for it being public.

PR #1043, opened 2026-04-07
decouple xlsb from xmlbeans

tballison:
This attempts to decouple xmlbeans from xlsb.

The one area of concern I have is the public api change:

XSSFSheetXMLHandler.SheetContentsHandler.cell(String, String, XSSFComment) → cell(String, String, Comment)
   
This is:
* Source-compatible: Comment is a supertype of XSSFComment, so existing implementations that reference XSSFComment just need to change their parameter type and recompile
* Binary-incompatible: Any compiled code implementing SheetContentsHandler will need recompilation against the new version

If this is a bridge too far, we can try to refactor in another way.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
@@ -31,8 +31,10 @@ import org.apache.poi.util.Internal;
 import org.apache.poi.util.LittleEndian;
 import org.apache.poi.util.StringUtil;
 import org.apache.poi.ss.usermodel.Comment;
+import org.apache.poi.util.Removal;
 import org.apache.poi.xssf.eventusermodel.XSSFSheetXMLHandler;
 import org.apache.poi.xssf.model.SharedStrings;
+import org.apache.poi.xssf.usermodel.XSSFComment;
 
 /**
  * @since 3.16-beta3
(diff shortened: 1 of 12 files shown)


poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBSheetHandler.java
```diff
@@ -447,9 +447,9 @@ public interface XSSFBSheetContentsHandler {
          *     </code>. See the code in <code>
          * poi-examples/src/main/java/org/apache/poi/xssf/eventusermodel/XLSX2CSV.java</code> for an
          *     example of how to handle this scenario.
-         * @see #doubleCell(String, double, XSSFComment, ExcelNumberFormat)
+         * @see #doubleCell(String, double, Comment, ExcelNumberFormat)
          */
-        void stringCell(String cellReference, String value, XSSFComment comment);
+        void stringCell(String cellReference, String value, Comment comment);
```
pjfanning, line 454: these are breaking changes - we have a policy of deprecating APIs and later removing them

Can we keep the XSSFComment methods and deprecate them adding the Comment overloads?

Add `@Removal` annotation to the deprecated methods (v7.0.0 removal). `@since 6.0.0` to the new methods.

If methods are not public then it is ok to change them without this deprecate and overload approach.

pjfanning (approved):
lgtm

Issue #1041, opened 2026-04-07
Cannot parse DOC files created by WPS Office


lurongjiang:
## Description

DOC files created by WPS Office (Kingsoft Office) cannot be parsed by Apache POI. The files have valid OLE2 headers (`D0 CF 11 E0`) and are detected as `application/msword` by Tika, but POI throws `IndexOutOfBoundsException: Block XX not found` when attempting to parse them.

## Steps to Reproduce

1. Obtain a DOC file created by WPS Office (sample file attached)
2. Try to parse the file using POI HWPFDocument or Tika

Example code:
```java
try (FileInputStream fis = new FileInputStream("sample.doc")) {
    HWPFDocument document = new HWPFDocument(fis);
    WordExtractor extractor = new WordExtractor(document);
    String text = extractor.getText();
} catch (Exception e) {
    e.printStackTrace();
}
```

## Expected Behavior

The DOC file should be parsed successfully, returning the document text content.

## Actual Behavior

```
java.lang.IndexOutOfBoundsException: Block 52 not found
    at org.apache.poi.poifs.filesystem.POIFSFileSystem.getBlockAt(POIFSFileSystem.java:474)
    ...
Caused by: java.lang.IndexOutOfBoundsException: Unable to read 512 bytes from 27136 in stream of length 26750
```

### File Header
- File size: 26,750 bytes
- Magic bytes: `D0 CF 11 E0 A1 B1 1A E1` (valid OLE2 signature)
- Tika detects as: `application/msword`

### OLE2 Structure (from header)
```
Offset 0x00: D0 CF 11 E0 A1 B1 1A E1 00 00 00 00 00 00 00 00
Offset 0x10: 00 00 00 00 00 00 00 00 3E 00 03 00 FE FF 09 00
Offset 0x20: 06 00 00 00 00 00 00 00 00 00 00 00 01 00 00 00
Offset 0x30: 01 00 00 00 00 00 00 00 00 10 00 00 02 00 00 00
Offset 0x40: 01 00 00 00 FE FF FF FF 00 00 00 00 00 00 00 00
```

- FAT size: 1 byte
- FAT block count: 1
- First FAT sector ID: 0
- MiniFAT block count: -2 (0xFFFFFFFE)

### WPS Signature Found
The file contains WPS-specific metadata:
- `WPS Office_11.1.0.10009`
- `KSOTemplate`
- `KSOProductBuildVer`

### Comparison with WPS "Save As" Fixed File
[...]

PR #1044, opened 2026-04-08
Fix block reference handling in ByteArrayBackedDataSource

lurongjiang:
- Modify the ByteArrayBackedDataSource.read method to handle positions beyond the file size.
- Return a zero-padded buffer instead of throwing an exception when the position exceeds the file size.


lurongjiang:
side effect

lurongjiang:
### **Changes Made**  
1. **Removed zero-padding logic**  
   - During standard `.doc` file parsing, zero-padding introduced unexpected exceptions due to invalid buffer content.  
   - Added test cases for both **standard (Microsoft Office 97-2003 `.doc`)** and **non-standard (WPS-compatible)** formats to ensure robustness.  

2. **Enhanced out-of-bounds handling in `FileBackedDataSource`**  
   - Discovered that file-based operations rely on `FileBackedDataSource`, which previously threw exceptions when `position >= size()`.  
   - Now returns an empty `ByteBuffer.allocate(length)` instead of throwing an exception, ensuring consistent behavior with `ByteArrayBackedDataSource`.  
   - Added corresponding test cases to validate boundary conditions for file-backed reads.  

### **Impact Assessment**  
- Since the original implementation threw exceptions on out-of-bounds access, the new behavior (returning an empty buffer) maintains backward compatibility for valid cases.  
- For invalid cases (e.g., corrupted files or incomplete reads), the change prevents abrupt exceptions, allowing callers to handle incomplete data gracefully (e.g., failing silently or logging errors).  
- **No functional regression expected**, as the new logic aligns with the previous exception-throwing behavior in terms of outcome (failure to parse invalid data).  

### **Testing Validation**  
- Verified that:  
  - Standard `.doc` files parse correctly (no zero-padding artifacts).  
  - Non-standard WPS files remain supported.  
  - Out-of-bounds reads (both byte-array and file-backed) now return empty buffers instead of exceptions.

pjfanning:
target main branch - not 5.5.x

poi/src/main/java/org/apache/poi/poifs/nio/FileBackedDataSource.java
```diff
@@ -95,7 +95,7 @@ public FileChannel getChannel() {
     @Override
     public ByteBuffer read(int length, long position) throws IOException {
         if (position >= size()) {
-            throw new IndexOutOfBoundsException("Position " + position + " past the end of the file");
+           return ByteBuffer.allocate(length);
```
pjfanning: add a comment about why this is allowed - like in the other class above

poi-scratchpad/src/test/java/org/apache/poi/hwpf/TestHWPFParser.java
```diff
@@ -39,6 +44,47 @@ void testDoc() throws Exception {
...
+    @Test
+    void testOffice97_2003DocRead() throws Exception {
+        try (
+            InputStream stream = HWPFTestDataSamples.openSampleFileStream("issue_1041_2.doc");
+            HWPFDocument doc = HWPFParser.parse(stream)
+        ) {
+            WordExtractor extractor = new WordExtractor(doc);
+            String text = extractor.getText();
+            assertNotNull(doc);
+            assertNotNull(text);
```
pjfanning: must test the actual text - it being not null is not enough

pjfanning:
I must admit that I'm basically -1 on this.

I don't see why we should be returning empty ByteBuffers.

I think we should fail when we read corrupt files.

Feel free to fork POI and hack it how you please but I can't in good conscience agree to a change like this.

If it was an option where this behaviour only kicks in if you set a flag on the extractor to say ignore corrupt blocks - I would possibly agree.

pjfanning:
relates to #1041

lurongjiang:
### Key Changes:

1. **Added Configurable Tolerance Mode** (`org.apache.poi.poifs.allowCorruptBlocks`)
   - Modified `ByteArrayBackedDataSource.java` and `FileBackedDataSource.java`
   - Default behavior: **Strict mode** - throws `IndexOutOfBoundsException` when encountering blocks beyond EOF
   - Optional tolerant mode: Set system property to `true` to allow reading corrupt files with missing blocks
   - System property is checked dynamically at runtime, allowing flexible configuration

2. **Improved Test Coverage**
   - Updated `TestHWPFParser.java` with comprehensive tests:
     - `testDocRead()`: Verifies file can be read in tolerant mode with actual text content validation
     - `testDocReadStrictMode()`: Verifies strict mode properly rejects corrupt files
     - `testWpsDocByFs()`: Tests file system-based reading with content validation
     - `testOffice97_2003DocRead()`: Validates normal document reading
   - All tests now verify actual text content (not just non-null), checking:
     - Text is not null
     - Text is not empty
     - Text is not blank (after trimming)

3. **Clear Error Messages**
   - Exception messages guide users on how to enable tolerant mode if needed
   - Example: "Position X is beyond EOF (Y). Set system property 'org.apache.poi.poifs.allowCorruptBlocks' to true to allow reading corrupt files with missing blocks."

### Design Rationale:

- **Fail-fast by default**: Aligns with Apache POI's principle of strict validation
- **Opt-in tolerance**: Users who need to handle damaged documents can enable it via system property
- **No API changes**: Backward compatible, uses standard Java system properties
- **Follows existing patterns**: Similar to other POI configuration options like `org.apache.poi.ss.ignoreMissingFontSystem`

### Testing:

All tests pass successfully, verifying both strict and tolerant modes work as expected. 

@pjfanning

pjfanning:
This is a WPS bug so I don't think we should make hacks in fundamental parts of POI code:
https://github.com/apache/poi/issues/1041#issuecomment-4208001098

Report the issue to WPS

lurongjiang:
Thank you for the review and feedback. I understand your concerns.

My original intention was that since POI aims to support various Office document formats, I thought it should be able to handle WPS-generated files as well. When I encountered parsing failures with certain WPS documents, I attempted to add a tolerance mechanism as a workaround.

However, I now understand your point. This is not really a WPS bug, but rather a limitation of POI's current implementation - it cannot handle certain non-standard file formats generated by WPS. Adding workarounds in POI's core components (like `ByteArrayBackedDataSource` and `FileBackedDataSource`) would compromise the integrity of the codebase.

I'll close this PR as suggested. Users who need to handle such files can implement workarounds at the application level rather than modifying the library itself.

pjfanning:
> Thank you for the review and feedback. I understand your concerns.
> 
> My original intention was that since POI aims to support various Office document formats, I thought it should be able to handle WPS-generated files as well. When I encountered parsing failures with certain WPS documents, I attempted to add a tolerance mechanism as a workaround.
> 
> However, I now understand your point. This is not really a WPS bug, but rather a limitation of POI's current implementation - it cannot handle certain non-standard file formats generated by WPS. Adding workarounds in POI's core components (like `ByteArrayBackedDataSource` and `FileBackedDataSource`) would compromise the integrity of the codebase.
> 
> I'll close this PR as suggested. Users who need to handle such files can implement workarounds at the application level rather than modifying the library itself.

I can't open the file using Microsoft Word. So it is a WPS bug.

PR #1046, opened 2026-04-15
Use cached value for __xludf.DUMMYFUNCTION when recalculating cell

tucker-adkison:
## Summary
When converting from Google Sheets to Excel, Google Sheets puts the placeholder formula `__xludf.DUMMYFUNCTION` to maintain backwards compatibility with Excel. The value appears as `=IFERROR(@__xludf.DUMMYFUNCTION("..."), value)`, with value being the cached value from Google Sheets

This PR handles this issue by keeping the cached value instead of throwing a `NotImplementedException`

### DIFF
--- a/poi/src/main/java/org/apache/poi/ss/formula/atp/AnalysisToolPak.java
+++ b/poi/src/main/java/org/apache/poi/ss/formula/atp/AnalysisToolPak.java
@@ -68,6 +68,10 @@ public final class AnalysisToolPak implements UDFFinder {
         // case-sensitive
         if(name.startsWith(prefix)) name = name.substring(prefix.length());
 
+        // Google Sheets wraps non-Excel functions with the __xludf. prefix
+        final String xludfPrefix = "__xludf.";
+        if(name.startsWith(xludfPrefix)) name = name.substring(xludfPrefix.length());
+
         // FIXME: inconsistent case-sensitivity
         return _functionsByName.get(name.toUpperCase(Locale.ROOT));
     }
@@ -118,6 +122,7 @@ public final class AnalysisToolPak implements UDFFinder {
         r(m, "DEC2OCT", null);
         r(m, "DELTA", Delta.instance);
         r(m, "DISC", null);
+        r(m, "DUMMYFUNCTION", DummyFunction.instance);
         r(m, "DOLLARDE", DollarDe.instance);
         r(m, "DOLLARFR", DollarFr.instance);
         r(m, "DURATION", null);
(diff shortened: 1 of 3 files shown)


pjfanning:
Any docs to back up that this is a standard that we should support? And is there a reason to recalc? Poi will never have good support for edge case functions and the XML files in xlsx have cached eval results that poi uses.

pjfanning:
Poi let's you register your own functions. Google will find links for you. One is https://poi.apache.org/components/spreadsheet/user-defined-functions.html

tucker-adkison:
> Any docs to back up that this is a standard that we should support? And is there a reason to recalc? Poi will never have good support for edge case functions and the XML files in xlsx have cached eval results that poi uses.

To clarify, this change is not trying to evaluate __xludf.DUMMYFUNCTION in POI.

Google Sheets can emit formulas like __xludf.DUMMYFUNCTION(...) for functions Excel does not support, but it also stores the computed cached value in the XLSX. My change uses that cached value rather than failing on the unsupported formula itself.

So the benefit here is better interoperability with Google Sheets-generated XLSX files, without requiring POI to implement evaluation for those functions because those functions are not supported by Excel’s formula engine.

pjfanning:
-1 from me without proper docs.
And you can use POI's methods to register your own workaround without us adding anything to POI.

tucker-adkison:
> -1 from me without proper docs. And you can use POI's methods to register your own workaround without us adding anything to POI.

I could not find a formal spec that names __xludf.DUMMYFUNCTION itself. What I did find is formal OOXML documentation that formula cells may store a cached last-calculated result in <v>, and POI already documents using cached formula results when evaluation cannot be completed. The justification here is not to add this Google function, but more to keep the cache value when recalculation is unable to evaluate the cell

OOXML spec: https://learn.microsoft.com/en-us/office/open-xml/spreadsheet/working-with-formulas
POI Documentation: https://poi.apache.org/components/spreadsheet/eval.html
Microsoft Q&A: https://learn.microsoft.com/en-us/answers/questions/4986422/xludf-dummy-funtion

pjfanning:
and why won't you answer why you can't just add the function using POI's API to add UDFs?
or why you even need to evaluate the cell values? the evaluated values are already cached in the cell XML that POI reads and POI will happily use the cached values

tucker-adkison:
> and why won't you answer why you can't just add the function using POI's API to add UDFs?


I already implemented this as a UDF registration internally in our codebase, exactly as you're suggesting, and it's working as expected. The reason I put up this PR is so other POI users who hit this issue don't have to independently discover and implement the same workaround, when the fix is a single function that always returns an error. 

The cell shows up as an error natively in excel. So when parsing the excel, we need the formula cells to be evaluated to a value. When evaluating the cell, POI was throwing `NotImplementedException`, which is why I had to go with the UDF registration.

PR #1052, opened 2026-04-24
Fix: XSLFTableCell.setBorderColor Overwrites

micahsmith:
We've been attempting to generate a pptx using POI but kept getting prompted to repair when actually loading into Powerpoint. Through the debugging process, it was revealed that the violation is caused by the existence of multiple solidFill elements.

Current behavior: calling `setBorderColor` multiple times will result in appending multiple solidFill elements.

New behavior: `setBorderColor` checks for existing solidFill element and removes it before creating a new one.


pjfanning (approved):
lgtm - thanks

PR #1053, opened 2026-04-28
prevent overflow in blip buffer size calculation

jmestwa-coder:
### Summary
This change updates buffer size calculations in HSLF blip processing to avoid potential integer overflow before array allocation.

### What changed

\- Replaced inline size calculations using \`new byte\[...\]\` with:

\- \`long\`-based size computation

\- \`IOUtils.safelyAllocate(...)\` for allocation

\- Applied consistently to:

\- \`Bitmap\`

\- \`EMF\`

\- \`PICT\`

\- \`WMF\`

### Why

In the current implementation, buffer sizes are computed using integer arithmetic:

new byte\[a + b + c\]

If intermediate values exceed Integer.MAX\_VALUE, this can overflow before allocation, leading to incorrect buffer sizes and runtime failures.

Using IOUtils.safelyAllocate ensures:

*   size is evaluated using long
    
*   overflow is detected before allocation
    
*   behavior is consistent with existing POI safety patterns
    

### Scope

*   Limited to HSLF blip classes only
    
*   No API changes
    
*   No behavioral changes for valid inputs
    

### Notes

This follows the same allocation safety approach already used in other parts of POI.

### DIFF
--- a/poi-scratchpad/src/main/java/org/apache/poi/hslf/blip/WMF.java
+++ b/poi-scratchpad/src/main/java/org/apache/poi/hslf/blip/WMF.java
@@ -90,7 +90,8 @@ public final class WMF extends Metafile {
         header.setZipSize(compressed.length);
 
         byte[] checksum = getChecksum(data);
-        byte[] rawData = new byte[checksum.length * getUIDInstanceCount() + header.getSize() + compressed.length];
+        long size = Math.addExact(Math.addExact((long)checksum.length * getUIDInstanceCount(), header.getSize()), compressed.length);
+        byte[] rawData = IOUtils.safelyAllocate(size, org.apache.poi.hslf.record.RecordAtom.getMaxRecordLength());
         int offset = 0;
 
         System.arraycopy(checksum, 0, rawData, offset, checksum.length);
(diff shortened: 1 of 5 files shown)


poi-scratchpad/src/main/java/org/apache/poi/hslf/blip/WMF.java
```diff
@@ -90,7 +90,8 @@ protected byte[] formatImageForSlideshow(byte[] data) {
         header.setZipSize(compressed.length);
 
         byte[] checksum = getChecksum(data);
-        byte[] rawData = new byte[checksum.length * getUIDInstanceCount() + header.getSize() + compressed.length];
+        long size = (long)checksum.length * getUIDInstanceCount() + header.getSize() + compressed.length;
```
pjfanning: Move this calculation and the other new calculations in the PR to a shared method that can be tested in your tests. Your tests don't seem to test the code and don't seem to provide any regression protection for the future. I only have my phone with me so I may have missed something but my cursory check of the new code makes me suspicious that this new code is maintainable.

jmestwa-coder: I updated the calculations to use Math.addExact for clearer, overflow-safe logic and kept the change minimal without adding a new helper. I also updated the test to cover the actual calculation path and verify the overflow behaviour.

pjfanning:
I went with https://github.com/apache/poi/pull/1054 instead.

PR #1055, opened 2026-05-05
Enforce max entry size consistently for streaming ZIP entries

jmestwa-coder:
### Summary

Enforce `maxEntrySize` consistently for ZIP entries processed via streaming, including when the entry size is unknown.

---

### Changes

- Add fail-fast validation for known-size entries.
- Replace size-limited copy with explicit bounded streaming:
  - Track bytes read.
  - Throw `RecordFormatException` when the limit is exceeded.
- Apply the same enforcement to temp-file paths.

---

### Test

- Adds a deterministic test for an unknown-size entry exceeding the limit.
- Fails on previous behavior, passes with this change.

---

### Notes

- Tightens validation; oversized entries are now rejected earlier.
- Scope limited to `ZipArchiveFakeEntry`.


pjfanning:
LGTM.

I don't like the use of RecordFormatException here but it is consistent with some existing IOUtil code.

I might follow up and rework the exception type.

pjfanning:
I reverted this. I had a look again and I don't think this is needed. The whole point of ZipInputStreamZipEntrySource.getThresholdBytesForTempFiles() is that users can opt in to protecting memory by using temp files.
The MaxEntrySize is a way to protect memory usage and the code you modified is using temp files so we don't need the check.

PR #1062, opened 2026-05-11
Enforce ZipSecureFile limits for encrypted temp ZIP streams

jmestwa-coder:
## Summary

* restore `ZipSecureFile` threshold enforcement during encrypted temp ZIP processing
* avoid sharing mutable `Cipher` state across returned streams
* preserve the existing streaming-based processing flow
* add regression coverage for the affected paths

## Changes

* wrap encrypted temp ZIP streaming with `ZipArchiveThresholdInputStream`
* keep `ZipArchiveThresholdInputStream#getNextEntry()` package-local
* add a small internal streaming helper to preserve existing streaming semantics without widening API visibility
* create independent decrypt/encrypt `Cipher` instances for each stream/entry
* add regression tests for:

  * independent stream reads
  * max entry size enforcement
  * max file count enforcement
  * inflate ratio enforcement

## Before

* encrypted temp ZIP creation bypassed `ZipSecureFile` limits
* multiple returned streams shared mutable decrypt `Cipher` state
* partial reads could corrupt later stream reads

## After

* encrypted temp ZIP processing enforces existing `ZipSecureFile` protections
* returned streams use independent cipher state
* streaming behavior is preserved without entry materialization

## Validation

Verified with:

```bash
.\gradlew.bat :poi-ooxml:test --tests org.apache.poi.poifs.crypt.tests.TestSecureTempZip -PjdkVersion=17
```

```bash
.\gradlew.bat :poi-ooxml:test --tests org.apache.poi.poifs.crypt.tests.TestEncryptedTempZipThreshold -PjdkVersion=17
```


pjfanning (changes requested):
breaks dozens of tests - Please try to run the unit tests before submitting PRs or I will stop reviewing any other PRs.

jmestwa-coder:
I’ve fixed the failing tests. If there’s anything still remaining or any other failures, please let me know — I’ll be happy to fix them

pjfanning (approved):

pjfanning:
lgtm - merged, thanks

PR #1063, opened 2026-05-12
Harden temporary file permissions

metsw24-max:
This patch hardens POI temporary file and directory handling by restricting access permissions to the owning user where supported.

`DefaultTempFileCreationStrategy` previously relied on platform-default permissions when creating temporary artifacts. These temporary files and directories can contain sensitive intermediate document data during processing, including streamed workbook contents, extracted package data, and decrypted document material.

The change updates temp file handling to:

* Apply restrictive POSIX permissions atomically during temp file and temp directory creation (`rw-------` for files, `rwx------` for directories) on POSIX-capable platforms
* Fall back to best-effort owner-only restrictions on non-POSIX platforms
* Restrict the POI temp directory itself to owner-only access where supported
* Preserve existing behavior and compatibility while reducing unintended local filesystem exposure

The patch also adds regression tests validating that created temporary files do not expose group/other read or write permissions on POSIX-supported filesystems.

### DIFF
--- a/poi/src/main/java/org/apache/poi/util/DefaultTempFileCreationStrategy.java
+++ b/poi/src/main/java/org/apache/poi/util/DefaultTempFileCreationStrategy.java
@@ -24,6 +24,9 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.attribute.FileAttribute;
+import java.nio.file.attribute.PosixFilePermission;
+import java.nio.file.attribute.PosixFilePermissions;
+import java.util.Set;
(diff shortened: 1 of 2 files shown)


pjfanning:
Merged. If we see any CI build issues or anyone hits issues with this, it will be reverted.

The default temp file strategy is meant to be easy to use and to not have any risks that it does not work on every platform.

Users who want stricter policies can implement their own strategies.
https://poi.apache.org/security.html

pjfanning:
reworked and changed a bit in https://github.com/apache/poi/pull/1065

PR #1064, opened 2026-05-12
Harden HPSF allocation-size arithmetic against silent overflow

metsw24-max:
Harden HPSF parsing against integer overflow conditions in Unicode string and array size calculations. Malformed or extreme metadata values are now rejected explicitly instead of relying on wrapped arithmetic behavior.

## Changes Done
- Replaced unchecked multiplication with Math.multiplyExact(...) in:
  - UnicodeString.read
  - Array.ArrayHeader.getNumberOfScalarValues
- Added validation rejecting negative Unicode string lengths
- Converted overflow conditions into IllegalPropertySetDataException
- Ensured overflow is detected at parse time before allocation or downstream processing

## Tests
Added coverage for:

- Unicode string byte-length overflow
- Negative Unicode string lengths
- Array scalar-count overflow from oversized dimensions

### DIFF
--- a/poi/src/main/java/org/apache/poi/hpsf/Array.java
+++ b/poi/src/main/java/org/apache/poi/hpsf/Array.java
@@ -19,6 +19,7 @@ package org.apache.poi.hpsf;
 import org.apache.poi.util.IOUtils;
 import org.apache.poi.util.Internal;
 import org.apache.poi.util.LittleEndianByteArrayInputStream;
+import org.apache.poi.util.RecordFormatException;
 
 @Internal
 public class Array {
@@ -71,9 +72,14 @@ public class Array {
         long getNumberOfScalarValues() {
             long result = 1;
             for ( ArrayDimension dimension : _dimensions ) {
-                // Math.multiplyExact rejects compounding dimensions that would silently
-                // overflow long math and bypass the > Integer.MAX_VALUE guard in Array.read.
-                result = Math.multiplyExact(result, dimension._size);
+                try {
+                    // Math.multiplyExact rejects compounding dimensions that would silently
+                    // overflow long math and bypass the > Integer.MAX_VALUE guard in Array.read.
+                    result = Math.multiplyExact(result, dimension._size);
+                } catch (ArithmeticException e) {
+                    throw new RecordFormatException(
+                            "Overflow when calculating the number of scalar values", e);
+                }
             }
             return result;
         }
--- a/poi/src/main/java/org/apache/poi/hpsf/Array.java
+++ b/poi/src/main/java/org/apache/poi/hpsf/Array.java
@@ -71,7 +71,12 @@ public class Array {
         long getNumberOfScalarValues() {
             long result = 1;
             for ( ArrayDimension dimension : _dimensions ) {
-                result *= dimension._size;
+                try {
+                    result = Math.multiplyExact(result, dimension._size);
+                } catch (ArithmeticException e) {
+                    throw new IllegalPropertySetDataException(
+                            "Array dimensions overflow when computing scalar count", e);
+                }
             }
             return result;
         }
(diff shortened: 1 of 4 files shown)


poi/src/main/java/org/apache/poi/hpsf/Array.java
```diff
@@ -71,7 +71,12 @@ void read( LittleEndianByteArrayInputStream lei ) {
         long getNumberOfScalarValues() {
             long result = 1;
             for ( ArrayDimension dimension : _dimensions ) {
-                result *= dimension._size;
+                try {
+                    result = Math.multiplyExact(result, dimension._size);
+                } catch (ArithmeticException e) {
+                    throw new IllegalPropertySetDataException(
```
pjfanning: why this random unrelated exception type?

pjfanning: https://github.com/apache/poi/blob/7607d1bbeeef45f5bb2c7492e1c41d153dfa78ee/poi/src/main/java/org/apache/poi/hpsf/IllegalPropertySetDataException.java#L21-L26

Where does this mention anything about IllegalPropertySetDataException being used for this type of exception?

ArithmeticException would be a better choice. Or possibly a new dedicated exception.

metsw24-max: I have dropped the try/catch wrappers in both UnicodeString.read and `Array.ArrayHeader.getNumberOfScalarValues` and now let `Math.multiplyExact` native `ArithmeticException` propagate as per your suggestion. 

The redundant negative-length pre-check in `UnicodeString` is gone too 

`IOUtils.safelyAllocate` already rejects negative sizes downstream so there is no behavior gap.

 Tests updated to expect `ArithmeticException`

pjfanning:
Breaks a test.

```
        Test Additional - #2849 spreadsheet/poi-fuzz.xls HPSF FAILED

        org.opentest4j.AssertionFailedError: spreadsheet/poi-fuzz.xls - failed for handler HPSFFileHandler:  expected class org.apache.poi.util.RecordFormatException ==> Unexpected exception type thrown, expected: <org.apache.poi.util.RecordFormatException> but was: <java.lang.ArithmeticException>
            at app/org.apache.poi.stress@6.0.0-SNAPSHOT-tests/org.apache.poi.stress.TestAllFiles.verify(TestAllFiles.java:296)
        Caused by: java.lang.ArithmeticException: long overflow
            at org.apache.poi.stress@6.0.0-SNAPSHOT-tests/org.apache.poi.stress.TestAllFiles.lambda$handleAdditional$2(TestAllFiles.java:272)
```

test-data/poi-integration-exceptions.csv
```diff
@@ -100,7 +100,7 @@ spreadsheet/sample.strict.xlsx,extract,OPC,,org.apache.poi.ooxml.POIXMLException
 spreadsheet/57914.xlsx,"handle,extract",XSSF,,org.apache.poi.ooxml.POIXMLException,"Strict OOXML isn't currently supported, please see bug #57699",
 spreadsheet/57914.xlsx,extract,OPC,,org.apache.poi.ooxml.POIXMLException,"Strict OOXML isn't currently supported, please see bug #57699",
 spreadsheet/poi-fuzz.xls,additional,HSSF,,org.apache.poi.util.RecordFormatException,Not enough data (0) to read requested (4) bytes,
-spreadsheet/poi-fuzz.xls,additional,HPSF,,org.apache.poi.util.RecordFormatException,"Tried to allocate an array of length 1,468,570,764, but the maximum length for this record type is 100,000",
```
pjfanning, line 103: I'm working on a change and will revert this latest commit

metsw24-max: ok @pjfanning

pjfanning:
Thanks - merged

softboy99:
Hi,
i also encounter problem: org.apache.poi.util.RecordFormatException: Tried to allocate an array of length 133,657,726, but the maximum length for this record type is 100,000,000. 
when the new version will be released?

pjfanning:
@softboy99 this is nothing to do with changing the 100,000 limit. There are existing APIs to increase the limits in your JVM. Look at your stacktrace and try to work out where the exception is thrown. Read the code in the class that throws the exception and you might see where the limit is coming from and how to increase it.

PR #1066, opened 2026-05-12
Reject path traversal filenames during HMEF attachment extraction

jmestwa-coder:
## Summary

Reject attachment filenames that would escape the selected extraction directory during HMEF/TNEF attachment extraction.

Previously, attachment filenames from TNEF metadata were written directly using:

```java
new File(dir, filename)
```

This allowed traversal-style filenames such as `../escaped.txt` to resolve outside the intended extraction directory.

## Changes

* normalize and validate resolved attachment output paths before writing files
* reject attachment filenames that resolve outside the chosen extraction directory
* add regression coverage for traversal-style attachment names

## Test

Added a regression test that:

* creates an in-memory malicious TNEF attachment using a traversal filename
* verifies extraction throws `IOException`
* verifies no file is created outside the extraction directory

### DIFF
--- a/poi-scratchpad/src/main/java/org/apache/poi/hmef/extractor/HMEFContentsExtractor.java
+++ b/poi-scratchpad/src/main/java/org/apache/poi/hmef/extractor/HMEFContentsExtractor.java
@@ -22,6 +22,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.file.Files;
+import java.nio.file.Path;
 
 import org.apache.poi.hmef.Attachment;
 import org.apache.poi.hmef.HMEFMessage;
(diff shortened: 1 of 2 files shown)


poi-scratchpad/src/main/java/org/apache/poi/hmef/extractor/HMEFContentsExtractor.java
```diff
@@ -152,10 +153,20 @@ public void extractAttachments(File dir) throws IOException {
...
             try (OutputStream fout = Files.newOutputStream(file.toPath())) {
                 fout.write(att.getContents());
             }
         }
     }
+
+    private static File getOutputFile(File dir, String filename) throws IOException {
+        Path outputDir = dir.toPath().toAbsolutePath().normalize();
+        Path outputFile = outputDir.resolve(filename).normalize();
+        if (!outputFile.startsWith(outputDir)) {
```
pjfanning: Can you have a look at https://github.com/apache/poi/blob/145ca469d93f5cb3f9f476d5d3b9195ac3ada744/poi/src/main/java/org/apache/poi/util/IOUtils.java#L622 and adjust that to have a shared method to do this check? Code reuse is important.

jmestwa-coder: Thanks for the suggestion. Updated the patch to reuse `IOUtils.newFile(...)` for the extraction path validation.

pjfanning (approved):
lgtm

PR #1067, opened 2026-05-13
Reject oversized XSSF table counts during integer conversion

metsw24-max:
XSSFTable currently narrows workbook-controlled table metadata from `long` to `int` using direct casts.

Oversized values in header row counts, totals row counts, or table column counts can silently truncate into invalid integer values and propagate malformed internal table state.

Replace these narrowing casts with `Math.toIntExact(...)` so oversized metadata is rejected deterministically instead of wrapping.

Add regression coverage verifying that oversized table metadata values throw `ArithmeticException`.


pjfanning (approved):

pjfanning:
lgtm - thanks

PR #1070, opened 2026-05-13
Reject oversized OOXML ptCount values in XDDFDataSourcesFactory

metsw24-max:
XDDFDataSourcesFactory currently converts OOXML ptCount values from
CTUnsignedInt using direct long-to-int casts:

    return (int) category.getPtCount().getVal();

Because the OOXML schema type is xsd:unsignedInt, crafted chart metadata can supply values larger than Integer.MAX_VALUE. These values silently truncate during narrowing conversion and may become negative.

For example:

    <c:ptCount val="4294967295"/>

wraps to -1 when converted to int.

This patch replaces all direct narrowing casts in XDDFDataSourcesFactory#getPointCount() implementations with Math.toIntExact() so malformed OOXML metadata is rejected at the conversion boundary instead of silently propagating corrupted values.

The change covers all parser branches:
- CTAxDataSource:
  - NumRef
  - StrRef
  - NumLit
  - StrLit
- CTNumDataSource:
  - NumRef
  - NumLit

Also adds unit tests that construct oversized ptCount values (Integer.MAX_VALUE + 1) for both category and numerical data sources and assert ArithmeticException instead of the previous silent wraparound behavior.


pjfanning (approved):
Lgtm

PR #1071, opened 2026-05-13
boundary handling and decompression limits in CompressedRTF

jmestwa-coder:
## Summary
This patch strengthens `CompressedRTF` boundary handling and decompression validation for both compressed and uncompressed RTF streams.

## Changes
- Enforce maximum decompressed size validation using `IOUtils.safelyAllocateCheck`.
- Add bounded output enforcement to prevent expansion beyond the declared decompressed size plus documented protocol padding.
- Fix uncompressed RTF handling to:
  - copy only the declared payload bytes,
  - stop at the correct stream boundary,
  - and avoid falling through into LZW decompression.
- Bound compressed input reads to the declared compressed payload size using `BoundedInputStream`.
- Add documentation for the MS-OXRTFCP trailing padding behavior.

## Tests
Added tests covering:
- rejection of oversized declared decompressed sizes,
- rejection of malformed compressed streams expanding beyond allowed padding limits.


poi-scratchpad/src/main/java/org/apache/poi/hmef/CompressedRTF.java
```diff
@@ -52,6 +54,16 @@ public final class CompressedRTF extends LZWDecompresser {
       "\\fmodern \\fscript \\fdecor MS Sans SerifSymbolArialTimes New RomanCourier" +
       "{\\colortbl\\red0\\green0\\blue0\n\r\\par \\pard\\plain\\f0\\fs20\\b\\i\\u\\tab\\tx";
 
+   private static final int DEFAULT_MAX_RECORD_LENGTH = 50_000_000;
```
pjfanning, line 58: why this length as opposed to some other value?

jmestwa-coder: I mainly wanted a limit that was large enough to avoid rejecting legitimate larger RTF payloads, while still enforcing a reasonable upper bound for malformed or inconsistent size declarations.

The 50MB value was meant as a conservative default rather than a format-defined maximum. Happy to align it with an existing POI-wide limit pattern, or reduce/remove the configurable default if you’d prefer.

poi-scratchpad/src/main/java/org/apache/poi/hmef/CompressedRTF.java
```diff
@@ -79,19 +105,28 @@ public void decompress(InputStream src, OutputStream res) throws IOException {
...
       // TODO - Handle CRC checking on the output side
+      IOUtils.safelyAllocateCheck(decompressedSize, MAX_RECORD_LENGTH);
+      LimitedOutputStream limited = new LimitedOutputStream(res, decompressedSize + MAX_PADDING_LENGTH);
 
       // Do we need to do anything?
       if(compressionType == UNCOMPRESSED_SIGNATURE_INT) {
          // Nope, nothing fancy to do
-         IOUtils.copy(src, res);
+         copyCompressedPayload(src, limited);
+         return;
```
pjfanning: don't use return like this - also copyCompressedPayload has its own limit checks, quite untidy to do this

poi-scratchpad/src/main/java/org/apache/poi/hmef/CompressedRTF.java
```diff
@@ -127,4 +162,48 @@ protected int populateDictionary(byte[] dict) {
...
+         int read = src.read(buffer, 0, (int)Math.min(buffer.length, remaining));
+         if (read < 0) {
+            throw new IOException("Not enough data to read " + getCompressedSize() + " bytes of uncompressed RTF");
+         }
+         res.write(buffer, 0, read);
+         remaining -= read;
+      }
+   }
+
+   private static final class LimitedOutputStream extends OutputStream {
```
pjfanning: https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/output/CountingOutputStream.java should be usable - don't really like declaring our own class for this

poi-scratchpad/src/test/java/org/apache/poi/hmef/TestCompressedRTF.java
```diff
@@ -197,4 +200,43 @@ void testFull() throws Exception {
         String decompStr = rtfAttr.getDataString();
         assertEquals(expString, decompStr);
     }
+
+    @Test
+    void testRejectsDeclaredDecompressedSizeOverLimit() throws Exception {
+        int oldLimit = CompressedRTF.getMaxRecordLength();
+        try {
+            CompressedRTF.setMaxRecordLength(4);
```
pjfanning, line 210: any test class that manipulates static limits must be annotated with org.junit.jupiter.api.parallel.Isolated to avoid affecting tests running in parallel

pjfanning (changes requested):
changes were requested a few weeks ago

jmestwa-coder:
@pjfanning Thanks for the review. I've addressed the feedback and rerun the tests successfully.

Please let me know if there's anything else you'd like changed.

pjfanning (approved):
lgtm

PR #1075, opened 2026-05-14
Validate HDGF v6+ chunk Length before narrowing to int

metsw24-max:
The HDGF v6+ `ChunkHeader` parser previously narrowed the unsigned 32-bit `Length` field using a plain `(int)` cast:

```java
(int) LittleEndian.getUInt(...)
```

Values larger than `Integer.MAX_VALUE` silently wrapped to negative numbers and could propagate into chunk offset arithmetic in `ChunkFactory.createChunk(...)`:

```java
offset + getLength() + sizeInBytes
```

This patch replaces the v6+ `Length` cast with `Math.toIntExact(...)`, matching the existing v4/v5 handling already used in the same method.

Identifier-style fields (`type`, `id`, `unknown1`) intentionally retain their existing casts. Existing tests already verify that negative reinterpretation is part of the current behavior contract (for example `header.getId() == -1` for `0xFFFFFFFF`).

### Tests

Added regression coverage for:

* rejecting oversized v6+ `Length` values (`0x80000001`)
* continuing to accept `Integer.MAX_VALUE` as a valid representable length


pjfanning:
thanks - merged

PR #1076, opened 2026-05-15
Guard HDGF PointerFactory offset/length uint32 narrowing

metsw24-max:
Replace unchecked uint32-to-int narrowing in HDGF PointerFactory for offset and length fields with Math.toIntExact.

These values originate from LittleEndian.getUInt(...) and later flow into stream offset/length arithmetic and StreamStore construction. Malformed values above Integer.MAX_VALUE could previously wrap into negative or truncated integers.

This change aligns PointerFactory with the HDGF ChunkHeader hardening introduced in PR #1075 and with the broader parser hardening work using Math.toIntExact for structural size and offset fields.

Address parsing intentionally remains unchanged because the value is only used for diagnostic/debug output and does not participate in bounds calculations.

- oversized offset rejection
- oversized length rejection
- v5 and v6+ parsing paths
- acceptance of Integer.MAX_VALUE boundary values


poi-scratchpad/src/main/java/org/apache/poi/hdgf/pointers/PointerFactory.java
```diff
@@ -45,8 +45,13 @@ public Pointer createPointer(byte[] data, int offset) {
...
             p.setAddress((int)LittleEndian.getUInt(data, offset+4));
-            p.setOffset((int)LittleEndian.getUInt(data, offset+8));
-            p.setLength((int)LittleEndian.getUInt(data, offset+12));
+            // Offset and Length flow into Stream.createStream as the (offset, length)
+            // pair handed to StreamStore / CompressedStreamStore. Match the recent
+            // ChunkHeader v6+ Length fix (PR #1075) and reject uint32 values that
+            // would silently wrap to a negative int rather than letting the wrapped
+            // value reach the downstream IOUtils.safelyClone bounds check.
+            p.setOffset(Math.toIntExact(LittleEndian.getUInt(data, offset+8)));
+            p.setLength(Math.toIntExact(LittleEndian.getUInt(data, offset+12)));
```
pjfanning: any chance that you could get these values as longs and then call the check methods on IOUtils? At least for length. The exception messages thrown by IOUtils are more informative than 'int overflow'.

metsw24-max: Done both fields are now read as long up-front. 
Length goes through `IOUtils.safelyAllocateCheck(value, Integer.MAX_VALUE);` Offset has a small helper throwing RecordFormatException with the offending value
Tests and poi-integration-exceptions.csv updated to match the new exception.

pjfanning (approved):
lgtm - thanks

PR #1077, opened 2026-05-17
Handle oversized uint32 offsets in QuillContents descriptors

metsw24-max:
`QuillContents` reads bit descriptor `from`/`len` fields as uint32 values from Publisher `.pub` files, but previously narrowed them directly to `int` via a plain cast before validation.

A crafted descriptor with values larger than `Integer.MAX_VALUE` could therefore wrap to negative values and only fail later inside `IOUtils.safelyClone(...)` with a generic `"Invalid offset\\length specified"` error that lost the original unsigned value.

This change validates the uint32 values before narrowing:

* oversized offsets now throw `RecordFormatException` carrying the original uint32 value
* lengths are gated through `IOUtils.safelyAllocateCheck(...)` before casting

Adds regression tests that build minimal in-memory `.pub` containers with crafted descriptor values and verify that oversized uint32 fields are rejected during `QuillContents` parsing.

Tested:

* existing HPBF tests
* new `TestQuillContentsOverflow` regression tests


pjfanning (approved):
lgtm

PR #1078, opened 2026-05-19
HSLF:  reject oversized PersistPtr sheet offsets

metsw24-max:
Harden HSLF PersistPtr parsing against unsigned-to-signed integer truncation when reading sheet offsets from PersistPtrIncrementalBlock records.

Previously, PersistPtrHolder read 32-bit unsigned sheet offsets using `LittleEndian.getUInt(...)` and narrowed them with a plain `(int)` cast. Values larger than `Integer.MAX_VALUE` wrapped to negative offsets and were stored in `_slideLocations`.

These wrapped offsets later flowed into `Record.buildRecordAtOffset(docstream, offset)`, where malformed inputs could cause the parser to read records from unintended positions in the document stream instead of rejecting the corrupted input cleanly.

## Changes

* Replaced the narrowing cast with `Math.toIntExact(...)` when parsing PersistPtr sheet offsets
* Added an inline hardening comment documenting the threat model and parser impact
* Preserved support for all valid signed-int offsets, including `Integer.MAX_VALUE`

## Tests

Added regression coverage in `TestPersistPtrHolder`:

* `testRejectsOversizedSheetOffset`

  * Verifies oversized uint32 offsets now fail with `ArithmeticException`

* `testAcceptsMaxIntSheetOffset`

  * Verifies legitimate maximum signed-int offsets continue to parse correctly

### DIFF
--- a/poi-scratchpad/src/main/java/org/apache/poi/hslf/record/PersistPtrHolder.java
+++ b/poi-scratchpad/src/main/java/org/apache/poi/hslf/record/PersistPtrHolder.java
@@ -127,7 +127,14 @@ public final class PersistPtrHolder extends PositionDependentRecordAtom {
             // Grab the offsets for each of the sheets
             for(int i=0; i<offset_count; i++) {
                 int sheet_no = offset_no + i;
-                int sheet_offset = (int)LittleEndian.getUInt(_ptrData,pos);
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #1079, opened 2026-05-25
guard XSSFBUtils wide-string byte length against int overflow

jmestwa-coder:
readXLWideString and readXLNullableWideString size the string as 2*(int)numChars, which overflows for char counts at or above 0x40000000 read from an .xlsb stream. The wrapped negative length slips past the offset+numBytes > data.length check, so an oversized record throws a raw StringIndexOutOfBoundsException instead of XSSFBParseException, and high-bit counts read the wrong length and desync the record stream. Compute the byte count in long so the bounds check applies before narrowing.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBUtils.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBUtils.java
@@ -53,7 +53,7 @@ public class XSSFBUtils {
             throw new XSSFBParseException("trying to read beyond data length: " +
              "offset="+offset+", numBytes="+numBytes+", data.length="+data.length);
         }
-        sb.append(new String(data, offset, (int)numBytes, StandardCharsets.UTF_16LE));
+        sb.append(new String(data, offset, Math.toIntExact(numBytes), StandardCharsets.UTF_16LE));
         return Math.toIntExact(numBytes+4);
     }
 
@@ -78,7 +78,7 @@ public class XSSFBUtils {
         if (offset+numBytes > data.length) {
             throw new XSSFBParseException("trying to read beyond data length");
         }
-        sb.append(new String(data, offset, (int)numBytes, StandardCharsets.UTF_16LE));
+        sb.append(new String(data, offset, Math.toIntExact(numBytes), StandardCharsets.UTF_16LE));
         return Math.toIntExact(numBytes+4);
     }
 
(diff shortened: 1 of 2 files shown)


poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBUtils.java
```diff
@@ -47,15 +47,14 @@ static int readXLNullableWideString(byte[] data, int offset, StringBuilder sb) t
...
+        long numBytes = 2L*numChars;
         offset += 4;
         if (offset+numBytes > data.length) {
             throw new XSSFBParseException("trying to read beyond data length: " +
              "offset="+offset+", numBytes="+numBytes+", data.length="+data.length);
         }
-        sb.append(new String(data, offset, numBytes, StandardCharsets.UTF_16LE));
-        numBytes+=4;
-        return numBytes;
+        sb.append(new String(data, offset, (int)numBytes, StandardCharsets.UTF_16LE));
```
pjfanning: you should validate before the cast in the sb.append params

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBUtils.java
```diff
@@ -74,14 +73,13 @@ public static int readXLWideString(byte[] data, int offset, StringBuilder sb) th
...
-        int numBytes = 2*(int)numChars;
+        long numBytes = 2L*numChars;
         offset += 4;
         if (offset+numBytes > data.length) {
             throw new XSSFBParseException("trying to read beyond data length");
         }
-        sb.append(new String(data, offset, numBytes, StandardCharsets.UTF_16LE));
-        numBytes+=4;
-        return numBytes;
+        sb.append(new String(data, offset, (int)numBytes, StandardCharsets.UTF_16LE));
```
pjfanning: you should validate before the cast in the sb.append params

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBUtils.java
jmestwa-coder: Good point. Switched it to Math.toIntExact(numBytes) so the narrowing is checked, same as the return line right below, instead of a silent (int) cast.

poi-ooxml/src/main/java/org/apache/poi/xssf/binary/XSSFBUtils.java
jmestwa-coder: Same change here, using Math.toIntExact(numBytes) for the append param.

pjfanning (approved):
lgtm

PR #1080, opened 2026-05-25
Guard Ole10Native readUtf16 byte length against int overflow

jmestwa-coder:
Found while fuzzing Ole10Native record parsing. readUtf16 reads a 32-bit char count and hands size * 2 to IOUtils.toByteArray; a count near 0x40000000 makes that product overflow to a negative int, which slips past the MAX_STRING_LENGTH cap so an oversized record surfaces a raw IllegalArgumentException from getFromUnicodeLE rather than the RecordFormatException the cap is meant to raise. Compute the byte count as size * 2L so the long overload applies the cap before narrowing.

### DIFF
--- a/poi/src/test/java/org/apache/poi/poifs/filesystem/TestOle10Native.java
+++ b/poi/src/test/java/org/apache/poi/poifs/filesystem/TestOle10Native.java
@@ -17,6 +17,7 @@
 
 package org.apache.poi.poifs.filesystem;
 
+import static org.apache.poi.POITestCase.assertContains;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.core.IsEqual.equalTo;
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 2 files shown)


poi/src/test/java/org/apache/poi/poifs/filesystem/TestOle10Native.java
```diff
@@ -107,4 +108,23 @@ void testOleNativeOOM() throws IOException {
...
+        data[8] = 'B';                                // fileName (AsciiZ)
+        // flags2, unknown1, ascii command length and data length stay zero
+        LittleEndian.putInt(data, 22, 0x40000001);    // command2 char count
+        LittleEndian.putInt(data, 0, data.length - 4); // totalSize
+
+        RecordFormatException ex = assertThrows(
+            RecordFormatException.class,
+            () -> new Ole10Native(data, 0)
+        );
+        assertTrue(ex.getMessage().contains("Tried to allocate"));
```
pjfanning: Use POITestCase assertContains

jmestwa-coder: Done, switched it to assertContains.

pjfanning (approved):
lgtm

PR #1082, opened 2026-05-25
Fix out-of-bounds fill in EmfPlusPath RLE point-type expansion

jmestwa-coder:
An EmfPlusPath with the RLE_COMPRESSED flag set expands its per-point type table from (runCount, type) pairs, but the fill starts at pointTypes.length instead of the running offset, so any RLE-compressed path overruns the buffer. Start the run at the current index and clamp its end to the array length so a crafted runCount can't write past pointTypes.


pjfanning (approved):
lgtm - thanks

PR #1089, opened 2026-05-26
Prevent crashes when calculating cells on absolute anchors.

jaragunde:
Absolute anchors in OOXML aren't anchored to cells, they have absolute positions in the document. The XSSFClientAnchor has code to calculate the cell equivalent, using the absolute position and the cell width and height. The same code is used to calculate the second anchor cell in one-cell anchors.

For absolute anchors, it takes an empty cell (EMPTY_MARKER) and calculates the first anchor cell by compounding it with the anchor position, the problem is the empty cell has some null fields that the code is not expecting, and it crashes.

We add those null checks and add a test that exercises this use case. The same test would crash if run on trunk, with the following exception:
```
java.lang.NullPointerException: Cannot invoke "org.apache.xmlbeans.XmlAnySimpleType.getStringValue()" because "coordUnion" is null
	at org.apache.poi.ooxml/org.apache.poi.ooxml.util.POIXMLUnits.parseLengthInner(POIXMLUnits.java:236)
	at org.apache.poi.ooxml/org.apache.poi.ooxml.util.POIXMLUnits.parseLength(POIXMLUnits.java:167)
	at org.apache.poi.ooxml/org.apache.poi.xssf.usermodel.XSSFClientAnchor.calcCell(XSSFClientAnchor.java:166)
	at org.apache.poi.ooxml/org.apache.poi.xssf.usermodel.XSSFClientAnchor.getCell1(XSSFClientAnchor.java:202)
	at org.apache.poi.ooxml/org.apache.poi.xssf.usermodel.XSSFClientAnchor.getCol1(XSSFClientAnchor.java:211)
	at org.apache.poi.ooxml/org.apache.poi.xssf.usermodel.TestXSSFDrawing.testAbsoluteAnchor(TestXSSFDrawing.java:938)


pjfanning (approved):
lgtm - merged

PR #1090, opened 2026-05-27
Implement ImageUtils.getDimensionFromAnchor variation.

jaragunde:
Add a variation of ImageUtils.getDimensionFromAnchor() that accepts ClientAnchor objects. A Sheet needs to be passed to obtain the cell width and height needed for the calculation. It uses the same code underneath as getDimensionFromAnchor(Picture).


pjfanning (approved):
lgtm

PR #1091, opened 2026-05-27
Support mc:AlternateContent for XWPFRun.getPictureText().

jaragunde:
Modern Word versions create shapes inside blocks like these:
```
    <mc:AlternateContent>
      <mc:Choice Requires="wps">
        <w:drawing>...</w:drawing>
      </mc:Choice>
      <mc:Fallback>
        <w:pict>...</w:pict>
      </mc:Fallback>
    </mc:AlternateContent>
```
The existing implementation of `XWPFRun.getPictureText()` only searched through drawing or pict items that are direct children of the run, ignoring whatever was inside an `<mc:AlternateContent>` block. We add code to include any `<w:drawing>` items inside these blocks.

We also modify `XWPFParagraph.getPictureText()` to insert line jumps between different picture texts, following the mold from `XWPFRun.getPictureText()`.


pjfanning (approved):
lgtm

PR #1092, opened 2026-05-27
add MathUtil

pjfanning:
can make more of use of it in later commits and also need to extend test coverage

### DIFF
--- a/poi/src/main/java/org/apache/poi/sl/draw/DrawPaint.java
+++ b/poi/src/main/java/org/apache/poi/sl/draw/DrawPaint.java
@@ -94,7 +94,7 @@ public class DrawPaint {
                         return new Color(color.getRed(), color.getGreen(), color.getBlue());
                     }
                     @Override
-                    public int getAlpha() { return (int)Math.round(color.getAlpha()*100000./255.); }
+                    public int getAlpha() { return Math.toIntExact(Math.round(color.getAlpha()*100000./255.)); }
                     @Override
                     public int getHueOff() { return -1; }
                     @Override
--- a/poi/src/main/java/org/apache/poi/ss/util/ImageUtils.java
+++ b/poi/src/main/java/org/apache/poi/ss/util/ImageUtils.java
@@ -40,6 +40,7 @@ import org.apache.poi.ss.usermodel.PictureData;
 import org.apache.poi.ss.usermodel.Row;
 import org.apache.poi.ss.usermodel.Sheet;
 import org.apache.poi.ss.usermodel.Workbook;
+import org.apache.poi.util.MathUtil;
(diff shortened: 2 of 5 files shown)


poi/src/main/java/org/apache/poi/util/MathUtil.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public static int safeFloatToInt(float f) {
+        if (f > Integer.MAX_VALUE || f < Integer.MIN_VALUE) {
+            throw new IllegalArgumentException("Value out of range: " + f);
+        }
+        return (int) f;
+    }
+
+    public static int safeDoubleToInt(double d) {
+        if (Double.isNaN(d)) {
```
xzel23, line 36: I don't know how much of a difference this makes (it could when used in a tight loop), but only one check is necessary. If we use `Double.isFinite()` instead, it will also catch NaN. `Double.isFinite(Double.NaN)` returns `false` just as `Double.isInfinite(Double.NaN)`, the functions are not symmetric in this aspect. So this should catch both:

```java
        if (!Double.isFinite(d)) { // DO NOT REPLACE WITH Double.isInfinite(d) !!!
            throw new IllegalArgumentException("Cannot convert number to int: " + d);
        }
 ```

File: poi/src/main/java/org/apache/poi/sl/draw/DrawPaint.java [hunk 2]
```diff
@@ -94,7 +94,7 @@ public Color getColor() {
                         return new Color(color.getRed(), color.getGreen(), color.getBlue());
                     }
                     @Override
-                    public int getAlpha() { return (int)Math.round(color.getAlpha()*100000./255.); }
+                    public int getAlpha() { return Math.toIntExact(Math.round(color.getAlpha()*100000./255.)); }
```
xzel23, line 97: I think this would be a good candidate for the new MathUtil class:

```java
public static int roundToInt(double d) {
    return safeDoubleToInt(Math.rint(d));
}
```

I used your rint() version I discovered below because it will handle infinity and NaN correcty.

poi/src/main/java/org/apache/poi/ss/util/ImageUtils.java
```diff
@@ -317,6 +298,6 @@ private static int getDimFromCell(double imgSize, int startCell, int startD, int
             }
         }
 
-        return (int)Math.rint(targetSize);
+        return MathUtil.safeDoubleToInt(Math.rint(targetSize));
```
xzel23, line 301: I think should use a method in MathUtil (as described above). Actually using MathUtil.safeDoubleToInt(Math.rint(...)) seems like a good idea as infinity and NaN should be handled correctly.

PR #1098, opened 2026-06-01
use long for EmfPlusDrawDriverString glyph byte length

metsw24-max:
EmfPlusDrawDriverString.init reads a 32-bit glyph count from untrusted EMF+ bytes and allocates glyphCount*2 bytes. The product is computed in int, so a count of 0x40000000 wraps negative and slips past the MAX_OBJECT_SIZE cap in IOUtils.toByteArray, surfacing a raw IllegalStateException instead of a RecordFormatException. Using glyphCount*2L routes through the long toByteArray overload so the cap fires first.


pjfanning (approved):
lgtm

PR #1104, opened 2026-06-03
Apply allocation bounds checks to EMF+ path and brush element counts

jmestwa-coder:
## Summary

EMF+ path and brush records allocate arrays directly from count fields read from the input stream.

This change applies the existing POI allocation safety checks before the following count-driven allocations:

- Path point arrays in `HemfPlusPath`
- Surrounding color arrays in `HemfPlusBrush`
- Boundary point arrays in `HemfPlusBrush`
- Position arrays in `HemfPlusBrush`

Invalid or excessively large counts are now rejected with `RecordFormatException` instead of attempting oversized allocations.

## Tests

- `Integer.MAX_VALUE` counts
- Negative counts (`0xFFFFFFFF`)

These tests verify that malformed inputs are rejected through the standard allocation validation path.


pjfanning (approved):
lgtm

PR #1105, opened 2026-06-03
EscherBSERecord fillFields() to return consumed bytes

jmestwa-coder:
## Summary

Restore the `fillFields()` contract in `EscherBSERecord` by returning the number of bytes actually consumed from the input stream rather than a reconstructed serialized size.

## Problem

`EscherBSERecord.fillFields()` returned the embedded blip's `getRecordSize()`, which can differ from the number of bytes actually read when malformed records contain inconsistent size information.

This can cause parent containers to advance by an incorrect offset and mis-locate subsequent sibling records during parsing.

## Fix

- Return the actual number of bytes consumed (`bytesRead`) when calculating the value returned by `fillFields()`.
- Preserve the expected parser invariant that `fillFields()` reports consumed input bytes.

## Test

- Add a regression test covering a malformed `EscherMetafileBlip` whose `cbSave` value exceeds the data declared by the record length.
- Verify that `EscherBSERecord.fillFields()` reports the record's actual on-disk length rather than an inflated reconstructed size.
- Demonstrate that the previous behavior would over-report consumed bytes and cause parser offset drift.

### DIFF
--- a/poi/src/main/java/org/apache/poi/ddf/EscherBSERecord.java
+++ b/poi/src/main/java/org/apache/poi/ddf/EscherBSERecord.java
@@ -125,7 +125,10 @@ public final class EscherBSERecord extends EscherRecord {
 
         _remainingData = IOUtils.safelyClone(data, pos, bytesRemaining, MAX_RECORD_LENGTH);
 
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #1107, opened 2026-06-03
Apply allocation bounds checks to EmfPolyDraw and WmfCreateRegion

sahvx655-wq:
EmfPolyDraw and WmfCreateRegion allocate arrays directly from count fields read from the input stream.

This change applies POI's existing allocation safety checks before the following count-driven allocations:

Point arrays in EmfPolyDraw
Scan object arrays in WmfCreateRegion

Invalid, excessively large, or negative counts are now rejected through POI's standard allocation validation path instead of attempting unchecked array allocations.

Tests

Integer.MAX_VALUE count in EmfPolyDraw
Negative count in WmfCreateRegion

These tests verify that malformed inputs are rejected before array allocation occurs.

Executed the automated unit tests:

./gradlew :poi-scratchpad:test -PjdkVersion=17


pjfanning (approved):
lgtm

PR #1108, opened 2026-06-03
Modernize test assertions using JUnit 5 assertInstanceOf

sahvx655-wq:
Replace legacy instanceof assertions with JUnit 5's assertInstanceOf across selected tests in the poi, poi-ooxml, and poi-scratchpad modules.
This modernizes the test suite, improves assertion readability, and eliminates several redundant casts while preserving existing test behavior.

Changes
Replaced assertTrue(obj instanceof Type) patterns with assertInstanceOf(Type.class, obj).
Removed redundant casts made unnecessary by assertInstanceOf.

Updated tests in:
poi
poi-ooxml
poi-scratchpad

Verification
Executed the affected test suites locally using Gradle and verified that all updated tests pass.


pjfanning (approved):
lgtm

PR #1116, opened 2026-06-04
Use unsigned element count when parsing EscherArrayProperty

jmestwa-coder:
## Summary

Use an unsigned element count when parsing `EscherArrayProperty`.

## Changes

* Replace `LittleEndian.getShort()` with `LittleEndian.getUShort()` when reading the array element count (`nElems`) in `EscherArrayProperty#setArrayData`.
* Align parsing behavior with the existing unsigned accessors in the same class.
* Ensure values in the full unsigned range (`0-65535`) are interpreted correctly.

## Why

* The element count field is treated as unsigned elsewhere in `EscherArrayProperty`.
* Reading the field as a signed value can misinterpret counts greater than `32767`.
* This creates inconsistent behavior between parsing and subsequent access to the property data.

## Tests

* Added a regression test covering an element count larger than `32767`.
* Verified that the parsed complex data size is calculated correctly using the unsigned value.

### DIFF
--- a/poi/src/main/java/org/apache/poi/ddf/EscherArrayProperty.java
+++ b/poi/src/main/java/org/apache/poi/ddf/EscherArrayProperty.java
@@ -175,7 +175,7 @@ public final class EscherArrayProperty extends EscherComplexProperty implements 
         if (emptyComplexPart) {
             resizeComplexData(0);
         } else {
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm

PR #1122, opened 2026-06-08
read OldFfn font description length as unsigned byte

metsw24-max:
**Signed font-description length in OldFfn**

The length byte for a Word 6/95 font record is read as a signed `byte`, so a description of 128 bytes or more turns negative and slips past the table-end guard. A record that large then loses its alternate font name and reports the wrong record length, so the remaining font table is parsed from the wrong offset. The newer `Ffn` reader already takes the equivalent `cbFfnM1` field through `getUByte`, so this keeps `OldFfn` consistent with it.


pjfanning (approved):
lgtm

PR #1123, opened 2026-06-08
Fix resource leak in ZipSecureFile constructors and correct path logging in ZipPackage

kali834x:
This PR fixes two issues in the `poi-ooxml` module:

1. Prevents a file descriptor leak in `ZipSecureFile` when `validateEntryNames()` fails during construction by ensuring the opened ZIP file is properly closed on initialization failure.
2. Corrects the warning log in `ZipPackage.closeImpl()` to report the actual temporary file path instead of the target document path when temporary file deletion fails.

 Verification

* Added a regression test covering invalid ZIP archives with duplicate entry names and verifying proper resource cleanup.
* All unit tests pass successfully.


pjfanning (approved):

pjfanning:
lgtm - thanks - merged

PR #1124, opened 2026-06-09
Harden duplicate ZIP entry validation against path separator bypasses

kali834x:
Normalizes backslash separators ('\') to forward slashes ('/') in ZipSecureFile and ZipInputStreamZipEntrySource during duplicate entry validation.

When validating ZIP archives against duplicate entry names (introduced for CVE-2025-31672), entry names were matched without normalizing path separators. Because some Windows zip utilities output backslashes (`\`) instead of forward slashes (`/`), an attacker could craft an archive containing both `xl/workbook.xml` and `xl\workbook.xml`. This would bypass the duplicate check but could lead to path lookup shadowing during file parsing.

 Normalizes path separators in `ZipSecureFile.validateEntryNames()`.
 Normalizes path separators in the `ZipInputStreamZipEntrySource` constructor map index.
 Added `testValidateMixedSeparatorDuplicateEntryNames()` in `TestZipSecureFile` to verify correct detection and rejection of mixed-separator duplicates.


pjfanning (approved):
lgtm - thanks

PR #1128, opened 2026-06-10
Fix temporary file leak in ZipInputStreamZipEntrySource on constructor failure

kali834x:
This PR fixes a temporary file leak in `ZipInputStreamZipEntrySource` when constructor initialization fails after temporary-backed entries have already been created.

Changes

* Added exception-safe cleanup to `ZipInputStreamZipEntrySource`.
* If initialization fails due to a duplicate ZIP entry, I/O error, or other unexpected exception, all previously created `ZipArchiveFakeEntry` instances are closed to ensure their temporary files are deleted.
* Cleans up internal state before rethrowing the original exception.
* Added `testZipInputStreamZipEntrySourceExceptionReleasesResources` to verify that temporary files are properly removed when constructor initialization fails.

Verification

* Executed the new unit test and confirmed that temporary files are cleaned up on constructor failure.
* Verified that the original exception is still propagated after resources are released.

PR #1135, opened 2026-06-12
Fix temporary file leak in ZipArchiveFakeEntry constructor on copy failure

kali834x:
When constructing a `ZipArchiveFakeEntry`, temporary resources (`TempFile` or `EncryptedTempData`) are allocated before the entry is fully initialized. If `IOUtils.copy(...)` throws an exception (e.g., Zip Bomb detection or an I/O failure), the constructor exits before returning an object reference.

In this failure path, the partially constructed `ZipArchiveFakeEntry` is never returned to `ZipInputStreamZipEntrySource` and is never added to the `zipEntries` map. Therefore, the cleanup logic introduced in PR #1128—which only closes successfully added entries—cannot reach these resources.

As a result, the temporary file or encrypted temporary data allocated by the failed constructor can remain on disk, causing a resource leak.

This fix adds constructor-level cleanup using a `try-finally` block to ensure that any allocated resources are released if initialization does not complete successfully.


pjfanning:
#1136 seems to have included these changes too

PR #1144, opened 2026-06-15
Optimize getEntry performance in ZipInputStreamZipEntrySource using O(1) case-insensitive lookups

kali834x:
Entry names are normalized and lowercased before being stored in the zipEntries map:

name = IOUtils.normalizePath(name).toLowerCase(Locale.ROOT);

However, getEntry() performs its initial lookup using a non-lowercased key:

final ZipArchiveEntry ze = zipEntries.get(normalizedPath);

As a result, lookups for mixed-case paths may miss the direct map lookup and fall back to a linear case-insensitive scan of all entries, degrading lookup performance from O(1) to O(N).

Solution

Apply the same normalization strategy during lookup:

return zipEntries.get(normalizedPath.toLowerCase(Locale.ROOT));

This makes lookup behavior consistent with key storage, restores O(1) retrieval, and removes the need for the fallback linear scan.


pjfanning (approved):
lgtm

PR #1148, opened 2026-06-24
Fix metadata loss in ZipArchiveFakeEntry when reading ZIP files from input streams

kali834x:
This pull request fixes a metadata preservation issue in ZipArchiveFakeEntry when ZIP files are read from an InputStream using ZipInputStreamZipEntrySource.

Previously, the constructor used super(entry.getName()), which created a new entry containing only the name and discarded metadata such as size, modification time, compression method, and extra fields. This could result in missing or invalid metadata when reading OOXML packages from streams.

The constructor now uses super(entry) and preserves the entry comment, ensuring that all available metadata is retained.

The tests were updated to separately verify metadata preservation for stream-based and file-based ZIP entry sources, accounting for the fact that entry comments are only available when reading through ZipFile.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/openxml4j/util/ZipArchiveFakeEntry.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/openxml4j/util/ZipArchiveFakeEntry.java
@@ -66,7 +66,8 @@ public final class ZipArchiveFakeEntry extends ZipArchiveEntry implements Closea
     private EncryptedTempData encryptedTempData;
 
     ZipArchiveFakeEntry(ZipArchiveEntry entry, InputStream inp) throws IOException {
-        super(entry.getName());
+        super(entry);
+        setComment(entry.getComment());
 
         final long entrySize = entry.getSize();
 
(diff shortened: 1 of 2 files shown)


pjfanning:
So when does POI need this metadata?
This change opens up an attack vector because info that POI doesn't need will now be kept in memory (by default) so an attacker could craft sneaky zip data that has sizeable metadata.
I'd prefer this to be driven by a real xlsx file (or docx, pptx, etc) where POI fails to read the contents of the file due to this issue.

kali834x:
Thanks for the feedback.
One observable issue is that when a package is loaded from an InputStream, PackagePart.getSize() returns -1 because the size metadata is not preserved in ZipArchiveFakeEntry. Loading the same package via ZipFileZipEntrySource returns the correct size.
I understand the concern about retaining unnecessary metadata. Instead of copying the entire entry, I can revise the PR to preserve only the metadata that POI actually requires. I'll also investigate whether I can produce a concrete OOXML example demonstrating a user-visible impact from the missing metadata.

pjfanning:
I'm not convinced that we need the size. The data is in the fake entry and we can use the size of the data. The data is held in a byte array so getting the size from that is easy.

POI is not a zip reading tool. It is a tool for parsing and writing Microsoft data formats.

pjfanning:
I added #1150 to try to improve the size value.

kali834x:
#1150 handles the size case, which was the only concrete issue I could actually point to here, and it does it without holding the extra metadata in memory. Agree POI doesn't need the rest of the entry fields to parse OOXML, and your point about not keeping attacker-controlled metadata around is fair. Closing this in favour of #1150, thanks.

PR #1149, opened 2026-06-24
Fix equals() and hashCode() contract violation in XSSFColor

kali834x:
This pull request fixes an equals()/hashCode() contract violation in XSSFColor.

equals() compares the logical color properties (ARGB, theme, index, tint, and auto), while hashCode() was based on the underlying CTColor XML string representation. As a result, two logically equal XSSFColor instances could produce different hash codes when created from different XML representations, causing incorrect behavior in hash-based collections and caches.

The fix replaces the XML-based hash code calculation with a field-based implementation consistent with the properties used by equals(). A unit test was added to verify that logically equivalent XSSFColor instances created from different XML contexts remain equal and produce identical hash codes.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFColor.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFColor.java
@@ -382,7 +382,13 @@ public class XSSFColor extends ExtendedColor {
 
     @Override
     public int hashCode() {
-        return ctColor.toString().hashCode();
+        int result = 17;
+        result = 31 * result + (isAuto() ? 1 : 0);
+        result = 31 * result + (isIndexed() ? Short.hashCode(getIndex()) : 0);
+        result = 31 * result + (isRGB() ? java.util.Arrays.hashCode(getARGB()) : 0);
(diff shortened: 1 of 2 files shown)


poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFColor.java
```diff
@@ -382,7 +382,13 @@ public static XSSFColor toXSSFColor(Color color) {
 
     @Override
     public int hashCode() {
-        return ctColor.toString().hashCode();
+        int result = 17;
+        result = 31 * result + (isAuto() ? 1 : 0);
```
pjfanning: how about using Objects.hash method?
https://www.baeldung.com/java-objects-hash-vs-objects-hashcode

kali834x:
Thanks for the suggestion. I'll update the implementation to use Objects.hash(...) and push an updated commit

poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFColor.java
kali834x: Done, switched it to Objects.hash. Tests still pass.

pjfanning (approved):
merged - thanks

PR #1185, opened 2026-07-13
Prevent NPE on XWPFTableCell.getWidth() when tcW is not set.

jaragunde:
Returns -1 mirroring the behavior of XWPFTable.getWidth(), although getWidthDecimal() returns 0 in that case, for both XWPFTable and XWPFTableCell. The standard implies, in a roundabout way, that the value should default to 0 but -1 can be, and may have already been, used as a marker of "presence" by clients of the POI API.


jaragunde:
@pjfanning I would like to know your opinion about returning -1 or 0 for `XWPFTableCell.getWidth()` when there's no explicit width set in the document. `XWPFTable.getWidth()` returns -1, but ``XWPFTable|XWPFTableCell.getWidthDecimal()` both return 0. I'm for using -1, as I explain in the message.

Heads-up: I will still work a bit on this topic; while I was testing this I realized it's possible to crash with NPE if the tcW element of the cell is set but empty.

pjfanning:
@jaragunde I wouldn't favour changing the existing method to return -1 when it never did before. Breaks backward compatibility. Introduce a new method or overload this method but include a new param, maybe a boolean, to control whether -1 or an exception is thrown in the not set case.
To be honest, it's not nice but compatibility guarantees matter. My preferred solution would be to add an isSet method that returns a boolean if tcW is set and to update the getWidth javadoc to mention that that is how you get work out the is not set case.

PR #1187, opened 2026-07-20
Adjusting XSLFDrawing's slide shape tree scan in constructor to prevent shape ID warnings.

### LINKED ISSUE #924: Bug: shape id <shapeId> has been already used warning when using XSLFGroupShape on XSLFSlide
Using latest Apache POI 5.4.1, I get warnings when working with XSLFGroupShape to contain shapes with other shapes created on XSLFSlide.  As far as I can tell, I'm using the API as intended.

The following is a small reproduction of the issue:
```java
		XMLSlideShow ppt = new XMLSlideShow();
		ppt.setPageSize(new java.awt.Dimension(pageWidth, pageHeight));
		
		XSLFSlide slide = ppt.createSlide();
		XSLFTextBox slideHeaderLabel = slide.createTextBox();
		slideHeaderLabel.setAnchor(new Rectangle(mainContentContainerMargin, 82, pageWidthSansMargins, 43));
		slideHeaderLabel.setInsets(new Insets2D(0, slideHeaderLabel.getInsets().left, 0, slideHeaderLabel.getInsets().right));
		XSLFTextParagraph slideHeaderParagraph = slideHeaderLabel.addNewTextParagraph();
		XSLFTextRun slideHeaderRun = slideHeaderParagraph.addNewTextRun();
		slideHeaderRun.setText("Title");

		XSLFTextBox slideSubHeaderLabel = slide.createTextBox();
		slideSubHeaderLabel.setAnchor(new Rectangle(mainContentContainerMargin, 125, pageWidthSansMargins, 32));
		slideSubHeaderLabel.setInsets(new Insets2D(0, slideSubHeaderLabel.getInsets().left, 0, slideSubHeaderLabel.getInsets().right));
		XSLFTextParagraph slideSubHeaderParagraph = slideSubHeaderLabel.addNewTextParagraph();
		XSLFTextRun slideSubHeaderRun = slideSubHeaderParagraph.addNewTextRun();
		slideSubHeaderRun.setText("Subtitle");

		XSLFGroupShape shapeGroup = slide.createGroup();
		XSLFAutoShape roundedRect = shapeGroup.createAutoShape();
		roundedRect.setShapeType(ShapeType.ROUND_RECT);
		roundedRect.setAnchor(new Rectangle(10, 0, 100, 86));
		roundedRect.setLineColor(Color.LIGHT_GRAY);
		roundedRect.setFillColor(Color.WHITE);
		roundedRect.setLineWidth(1.5);

		// Save the presentation
		try (FileOutputStream out = new FileOutputStream("example.pptx")) {
			ppt.write(out);
		}
		ppt.close();
``` 

Running the above example produces the following warning log messages:
```
2025-10-25 13:47:48,076  [Thread:main] WARN [org.apache.poi.xslf.usermodel.XSLFSheet] - shape id 1 has been already used.
2025-10-25 13:47:48,078  [Thread:main] WARN [org.apache.poi.xslf.usermodel.XSLFSheet] - shape id 2 has been already used.
2025-10-25 13:47:48,078  [Thread:main] WARN [org.apache.poi.xslf.usermodel.XSLFSheet] - shape id 3 has been already used.
2025-10-25 13:47:48,078  [Thread:main] WARN [org.apache.poi.xslf.usermodel.XSLFSheet] - shape id 4 has been already used.
```

dbunner2:
Creating a shape directly on an XSLFGroupShape produces a misleading "shape id * has been already used" warning for every shape already on the slide, even though those IDs are valid. This happens because the XSLFDrawing constructor always scans the entire slide and re-registers every shape ID, regardless of whether it's creating the top-level drawing for the slide or a drawing for a nested group.

That extra scan isn't necessary for group drawings. By the time a group drawing is created (through createGroup() or getShapes()), the top-level drawing has already recursively scanned the entire slide, including all nested groups. Scanning again just re-registers the same IDs and causes the false warnings.

This change limits the scan to the top-level drawing only. It also adds a regression test that captures the logged output and verifies no warnings are produced for the reported scenario. Reverting the fix causes the test to fail with the original warnings. I also ran the full org.apache.poi.xslf.* test suite (231 tests) with no regressions.

Fixes #924.

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFDrawing.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFDrawing.java
@@ -49,8 +49,7 @@ public class XSLFDrawing {
             for(XmlObject o : cNvPr) {
                 // powerpoint generates AlternateContent elements which cNvPr elements aren't recognized
                 // ignore them for now
-                if (o instanceof CTNonVisualDrawingProps) {
-                    CTNonVisualDrawingProps p = (CTNonVisualDrawingProps)o;
+                if (o instanceof CTNonVisualDrawingProps p) {
                     try {
                         sheet.registerShapeId(Math.toIntExact(p.getId()));
                     } catch (ArithmeticException e) {
--- a/poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFDrawing.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFDrawing.java
@@ -38,17 +38,24 @@ public class XSLFDrawing {
     /*package*/ XSLFDrawing(XSLFSheet sheet, CTGroupShape spTree){
         _sheet = sheet;
         _spTree = spTree;
-        XmlObject[] cNvPr = sheet.getSpTree().selectPath(
-                "declare namespace p='" + NS_PRESENTATIONML + "' .//*/p:cNvPr");
-        for(XmlObject o : cNvPr) {
-            // powerpoint generates AlternateContent elements which cNvPr elements aren't recognized
-            // ignore them for now
-            if (o instanceof CTNonVisualDrawingProps) {
-                CTNonVisualDrawingProps p = (CTNonVisualDrawingProps)o;
-                try {
-                    sheet.registerShapeId(Math.toIntExact(p.getId()));
-                } catch (ArithmeticException e) {
-                    throw new IllegalStateException("Int overflow with drawing id " + p.getId());
+
+        // Only performs the shape id scan when this is the sheet's own top-level
+        // drawing, since that scan is already recursive and covers every shape in
+        // the sheet. This avoids unnecessary re-scanning and the annoying
+        // "shape id _ has been already used" warning. See https://github.com/apache/poi/issues/924.
+        if (spTree == sheet.getSpTree()) {
+            XmlObject[] cNvPr = _spTree.selectPath(
+                    "declare namespace p='" + NS_PRESENTATIONML + "' .//*/p:cNvPr");
+            for(XmlObject o : cNvPr) {
+                // powerpoint generates AlternateContent elements which cNvPr elements aren't recognized
+                // ignore them for now
+                if (o instanceof CTNonVisualDrawingProps) {
+                    CTNonVisualDrawingProps p = (CTNonVisualDrawingProps)o;
+                    try {
+                        sheet.registerShapeId(Math.toIntExact(p.getId()));
+                    } catch (ArithmeticException e) {
+                        throw new IllegalStateException("Int overflow with drawing id " + p.getId());
+                    }
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFDrawing.java
```diff
@@ -38,17 +38,24 @@ public class XSLFDrawing {
...
+        // drawing, since that scan is already recursive and covers every shape in
+        // the sheet. This avoids unnecessary re-scanning and the annoying
+        // "shape id _ has been already used" warning. See https://github.com/apache/poi/issues/924.
+        if (spTree == sheet.getSpTree()) {
+            XmlObject[] cNvPr = _spTree.selectPath(
+                    "declare namespace p='" + NS_PRESENTATIONML + "' .//*/p:cNvPr");
+            for(XmlObject o : cNvPr) {
+                // powerpoint generates AlternateContent elements which cNvPr elements aren't recognized
+                // ignore them for now
+                if (o instanceof CTNonVisualDrawingProps) {
```
pjfanning: we support Java 17 min - so use `if (o instanceof CTNonVisualDrawingProps p)` - avoid the need for additional cast

poi-ooxml/src/test/java/org/apache/poi/xslf/usermodel/TestXSLFGroupShape.java
```diff
@@ -86,6 +95,38 @@ void testCreateShapes() throws Exception {
...
+        Appender captureAppender = new AbstractAppender("capture-for-test", null, null, false, null) {
+            @Override
+            public void append(LogEvent event) {
+                capturedWarnings.add(event.getMessage().getFormattedMessage());
+            }
+        };
+        captureAppender.start();
+        config.addAppender(captureAppender);
+        config.getRootLogger().addAppender(captureAppender, Level.WARN, null);
+        loggerContext.updateLoggers();
```
pjfanning: I must admit to disliking tests that read log messages. Can't we pass the info via the call stack?

poi-ooxml/src/test/java9/module-info.java
```diff
@@ -124,6 +124,7 @@ Licensed to the Apache Software Foundation (ASF) under one or more
     requires org.junit.jupiter.params;
     requires com.google.common;
     requires org.mockito;
+    requires org.apache.logging.log4j.core;
```
pjfanning: hopefully this can be removed

poi-ooxml/src/test/java/org/apache/poi/xslf/usermodel/TestXSLFGroupShape.java
dbunner2: That is fair.  I switched to a Mockito spy approach that verifies "registerShapeId" never gets called in this test case.  I also removed the log4j requires from module-info.java as it's no longer needed.

poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFDrawing.java
dbunner2: Done.

poi-ooxml/src/test/java9/module-info.java
dbunner2: This has been removed.

dbunner2:
Hi @pjfanning - Anything else I can add to this while I have some time?  Just let me know.

pjfanning:
Merged. I still don't like using mockito here but I don't have time to come up with a different approach. In future, I'm likely to just block PRs with mockito testing. This is a lib where everything should be testable already and we can always expose state via package private getters to expose more state that can be verified.

PR #1189, opened 2026-07-20
Keep cell number formats when plotting chart data from a sheet range

### LINKED ISSUE #1007: Bug: the number format is not updated after updating the XSLFChart in the presentation
I have a presentation template with a chart created and populated with data from Excel. It's a combo chart with three series, all of which are clustered column, with one series on the secondary axis. The numbers on the chart are formatted with an option "Linked to source" in Format Data Labels.

First, I load the presentation into memory, get a slide, find the chart, and process it.

From the chart, I get an Excel workbook, create a data format, create two new cell styles and assign each.
```
XSSFWorkbook workbook = chart.getWorkbook();

XSSFDataFormat dataFormat = workbook.createDataFormat();

XSSFCellStyle decimalCellStyle = workbook.createCellStyle();
decimalCellStyle.setDataFormat(dataFormat.getFormat("#,##0.###"));

XSSFCellStyle intCellStyle = workbook.createCellStyle();
intCellStyle.setDataFormat(dataFormat.getFormat("#,##0"));
```
 Next, I get a page, update the cell values and style.

```
XSSFSheet sheet = workbook.getSheetAt(0);

updateCategoryCells(periods, sheet);
updateValueCells(vals1, sheet, 1, decimalCellStyle, intCellStyle);
updateValueCells(vals2, sheet, 2, decimalCellStyle, intCellStyle);
updateValueCells(vals3, sheet, 3, decimalCellStyle, intCellStyle);
```

```
    private void updateCategoryCells(String[] periods, XSSFSheet sheet) {
        for (int i = 0; i < periods.length; i++) {
            XSSFRow row = sheet.getRow(i + 1);
            if (row == null) row = sheet.createRow(i + 1);

            XSSFCell categoryCell = row.getCell(0, Row.MissingCellPolicy.CREATE_NULL_AS_BLANK);
            categoryCell.setCellValue(periods[i]);
        }
    }

    private void updateValueCells(
            BigDecimal[] values,
            XSSFSheet sheet,
            int cellNum,
            XSSFCellStyle decimalCellStyle,
            XSSFCellStyle intCellStyle
    ) {
        for (int i = 0; i < values.length; i++) {
            XSSFRow row = sheet.getRow(i + 1);
            if (row == null) row = sheet.createRow(i + 1);

            XSSFCell cell = row.getCell(cellNum, Row.MissingCellPolicy.CREATE_NULL_AS_BLANK);

            if (values[i] == null) {
                cell.setBlank();
            } else {
                cell.setCellValue(values[i].doubleValue());
                setCellStyle(values[i], cell, decimalCellStyle, intCellStyle);
            }
        }
    }

    private void setCellStyle(
            BigDecimal value,
            XSSFCell cell,
            XSSFCellStyle decimalCellStyle,
            XSSFCellStyle intCellStyle
    ) {
        if (value.stripTrailingZeros().scale() > 0) {
            cell.setCellStyle(decimalCellStyle);
        } else {
            cell.setCellStyle(intCellStyle);
        }
    }
```

With data source factory I create data sources, get data lists from the chart, then get the series and replace the data. For each data list, I call the plot method.

```
            XDDFCategoryDataSource categoryDataSource = XDDFDataSourcesFactory.fromStringCellRange(
                    sheet,
                    new CellRangeAddress(1, Math.max(1, periods.length), 0, 0)
            );
            XDDFNumericalDataSource<Double> vals1DataSource = XDDFDataSourcesFactory.fromNumericCellRange(
                    sheet,
                    new CellRangeAddress(1, Math.max(1, periods.length), 1, 1)
            );
            XDDFNumericalDataSource<Double> vals2DataSource = XDDFDataSourcesFactory.fromNumericCellRange(
                    sheet,
                    new CellRangeAddress(1, Math.max(1, periods.length), 2, 2)
            );
            XDDFNumericalDataSource<Double> vals3DataSource = XDDFDataSourcesFactory.fromNumericCellRange(
                    sheet,
                    new CellRangeAddress(1, Math.max(1, periods.length), 3, 3)
            );

            List<XDDFChartData> chartDataList = chart.getChartSeries();
            XDDFChartData firstAndSecondValsChartData = chartDataList.getFirst();
            XDDFChartData thirdValsChartData = chartDataList.get(1);

            firstAndSecondValsChartData.getSeries(0).replaceData(categoryDataSource, vals1DataSource);
            firstAndSecondValsChartData.getSeries(1).replaceData(categoryDataSource, vals2DataSource);

            thirdValsChartData.getSeries(0).replaceData(categoryDataSource, vals3DataSource);

            chartDataList.forEach(chart::plot);
```

After that I import data to a new slide show, write into new file, save and open it. The ranges are displayed correctly, but the numbers are not formatted. However, after opening Excel, the chart updates and the numbers are displayed with formatting.

I suspect the problem lies in the plot method.

kamilkrzywanski:
When chart data is built with `fromNumericCellRange` and then `plot()` is called, the series format code in the chart `numCache` was left empty (and any existing format was cleared). PowerPoint then shows unformatted data labels if they are linked to source, until the embedded workbook is opened in Excel and the cache is refreshed.

This change reads number formats from the cells, uses the first non-blank numeric cell as the series format, and writes a per-point format when it differs. `setFormatCode(...)` still overrides the series format. Includes a regression test.

Fixes #1007

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSource.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSource.java
@@ -57,6 +57,14 @@ public interface XDDFDataSource<T> {
     String getFormatCode();
 
     /**
+     * @return number format for the point, or {@code null} to use {@link #getFormatCode()}
+     * @since 6.0.0
+     */
+    default String getPointFormatCode(int index) {
+        return null;
+    }
+
+    /**
      * @since 5.0.0
      */
     @Internal
@@ -78,6 +86,10 @@ public interface XDDFDataSource<T> {
                 CTNumVal ctNumVal = cache.addNewPt();
                 ctNumVal.setIdx(i);
                 ctNumVal.setV(value.toString());
+                String pointFormat = getPointFormatCode(i);
+                if (pointFormat != null && !pointFormat.equals(formatCode)) {
+                    ctNumVal.setFormatCode(pointFormat);
+                }
                 effectiveNumOfPoints++;
             }
         }
--- a/poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSource.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSource.java
@@ -65,7 +65,7 @@ public interface XDDFDataSource<T> {
      * </p>
      *
      * @param index zero-based point index
-     * @return the point's format code, or {@code null} to fall back to {@link #getFormatCode()}
+     * @return the data point's format code, or {@code null} if the underlying class does not support this method or if there is no format code set for this data point - NumericalCellRangeDataSource is one class that implements this method.
      * @since 6.0.0
      */
     default String getPointFormatCode(int index) {
--- a/poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSource.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSource.java
@@ -57,7 +57,15 @@ public interface XDDFDataSource<T> {
     String getFormatCode();
 
     /**
-     * @return number format for the point, or {@code null} to use {@link #getFormatCode()}
+     * Optional number format for a single data point in the numerical cache.
+     * <p>
+     * Return {@code null} when this point has no point-specific format and should
+     * inherit the series format from {@link #getFormatCode()}. Implementations that
+     * never use per-point formats can keep the default.
+     * </p>
+     *
+     * @param index zero-based point index
+     * @return the point's format code, or {@code null} to fall back to {@link #getFormatCode()}
      * @since 6.0.0
      */
     default String getPointFormatCode(int index) {
--- a/poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSource.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSource.java
@@ -61,11 +61,13 @@ public interface XDDFDataSource<T> {
      * <p>
      * Return {@code null} when this point has no point-specific format and should
      * inherit the series format from {@link #getFormatCode()}. Implementations that
-     * never use per-point formats can keep the default.
+     * never use per-point formats can keep the default ({@code null}).
+     * NumericalCellRangeDataSource is one class that implements this method.
      * </p>
      *
      * @param index zero-based point index
-     * @return the data point's format code, or {@code null} if the underlying class does not support this method or if there is no format code set for this data point - NumericalCellRangeDataSource is one class that implements this method.
+     * @return the data point's format code, or {@code null} if the underlying class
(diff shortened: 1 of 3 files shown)


poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSource.java
```diff
@@ -56,6 +56,14 @@ default String getFormula() {
 
     String getFormatCode();
 
+    /**
+     * @return number format for the point, or {@code null} to use {@link #getFormatCode()}
```
pjfanning: I don't understand this comment. Does `null` indicate that you should use `getFormatCode()` instead?

pjfanning:
@kamilkrzywanski I still don't understand what you intend null to mean. The updated javadoc doesn't give me any hints. Can you provide a comment describing what you want and I can update the Javadoc?

poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSource.java
```diff
@@ -56,6 +56,22 @@ default String getFormula() {
...
+    /**
+     * Optional number format for a single data point in the numerical cache.
+     * <p>
+     * Return {@code null} when this point has no point-specific format and should
+     * inherit the series format from {@link #getFormatCode()}. Implementations that
+     * never use per-point formats can keep the default.
+     * </p>
+     *
+     * @param index zero-based point index
+     * @return the point's format code, or {@code null} to fall back to {@link #getFormatCode()}
```
pjfanning:
```suggestion
     * @return the data point's format code, or {@code null} if the underlying class does not support this method or if there is no format code set for this data point - NumericalCellRangeDataSource is one class that implements this method.
```

pjfanning: @kamilkrzywanski would this be closer to what we need here?

pjfanning: Is there anywhere other than test code that overrides this default that returns null? And if not, why not?

pjfanning (changes requested):

poi-ooxml/src/main/java/org/apache/poi/xddf/usermodel/chart/XDDFDataSource.java
pjfanning: I don't see the point in this if someone has to subclass this class to provide an impl. Why can't we provide a class with an impl?

pjfanning: Ok, I get it now. I see NumericalCellRangeDataSource overrides the default impl to have a useful impl. I was looking at a specific commit as opposed to the PR change in its totality.

kamilkrzywanski:
@pjfanning glad that clicked. Yes, `NumericalCellRangeDataSource` is the real override; default null keeps other sources series-only.

On the Javadoc: null means use the series format from `getFormatCode()`, not always Excel's global default. Happy for you to reword it, or I can push a small Javadoc tweak.

pjfanning (approved):
lgtm - merged

PR #1190, opened 2026-07-21
Adjusting clearText() to avoid leaving shapes with no paragraphs

### LINKED ISSUE #919: BUG: Shape with empty paragraph & textrun generates invalid XML

dbunner2:
XSLFTextShape.clearText() currently removes every paragraph from a text body, leaving zero <a:p> elements. Since the OOXML schema requires at least one paragraph, PowerPoint opens presentations saved after a clearText() call in repair mode.

This change brings clearText() in line with how setText() already behaves: instead of removing every paragraph, it keeps the first paragraph and clears its contents. That preserves a valid text body while still removing all text.

Making this change exposed two existing bugs that clearText() had never hit before:

1. XSLFTextParagraph.clearButKeepProperties() removed field elements (<a:fld>, such as slide number or date placeholders) before reading the last run's formatting. If the last run was a field, this resulted in an XmlValueDisconnectedException. It also used _runs.size() to determine how many <a:r> elements to remove, even though _runs includes field runs, which could lead to an IndexOutOfBoundsException. The fix captures the last run's formatting before removing anything and removes only the runs that actually exist in the XML.

2. XSLFTextShape.copy() called clearText() and then appended the source paragraphs, leaving an extra empty paragraph at the beginning and shifting all paragraph indices by one. The fix removes that placeholder paragraph before copying whenever the source contains paragraphs. If the source is empty, the single empty paragraph is left intact, which is already the correct state.

I also added a regression test that verifies both the in-memory state and a save/reopen round trip.  

Fixes [#919.](https://github.com/apache/poi/issues/919)

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFTextParagraph.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFTextParagraph.java
@@ -1026,9 +1026,6 @@ public class XSLFTextParagraph implements TextParagraph<XSLFShape,XSLFTextParagr
         for (int i=thisP.sizeOfFldArray(); i>0; i--) {
             thisP.removeFld(i-1);
         }
-        // _runs.size() counts fields too, but fields and runs are removed from
-        // separate XML arrays - remove exactly what's actually there instead of
-        // relying on _runs.size() as the bound.
         for (int i=thisP.sizeOfRArray(); i>0; i--) {
             thisP.removeR(i-1);
         }
--- a/poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFTextParagraph.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xslf/usermodel/XSLFTextParagraph.java
@@ -1002,15 +1002,14 @@ public class XSLFTextParagraph implements TextParagraph<XSLFShape,XSLFTextParagr
      */
     /* package */ void clearButKeepProperties() {
         CTTextParagraph thisP = getXmlObject();
-        for (int i=thisP.sizeOfBrArray(); i>0; i--) {
-            thisP.removeBr(i-1);
-        }
-        for (int i=thisP.sizeOfFldArray(); i>0; i--) {
-            thisP.removeFld(i-1);
-        }
+
+        // Capture the last run's properties before removing anything - _runs can
+        // include field-backed runs (<a:fld>, e.g. slide number/date placeholders),
+        // and reading from one after its underlying XML has already been removed
+        // throws XmlValueDisconnectedException. See
+        // https://github.com/apache/poi/issues/919
         if (!_runs.isEmpty()) {
-            int size = _runs.size();
-            XSLFTextRun lastRun = _runs.get(size-1);
+            XSLFTextRun lastRun = _runs.get(_runs.size() - 1);
             CTTextCharacterProperties cpOther = lastRun.getRPr(false);
             if (cpOther != null) {
                 if (thisP.isSetEndParaRPr()) {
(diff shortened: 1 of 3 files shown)


dbunner2:
@pjfanning Anything I can help out with here to get this one merged?

pjfanning:
could you look at #1192 to see if there is anything useful there? - it covers the same issue

dbunner2:
@pjfanning  Of course.  It looks like #1192 took a slightly different approach.  It will insert a new blank paragraph after the clearText() call whereas mine (#1190) clears the paragraph but keeps the properties by utilizing clearButKeepProperties(), matching how setText() already behaves elsewhere in this class.  It is probable that #1192 would leave a stray empty paragraph at the front, since the source paragraphs get appended without removing the newly added placeholder first.

With that said it is a smaller change and might be preferable if we don't care about maintaining the original paragraph's properties.  Happy to close this one in favor of #1192 if you'd rather go with the smaller change, or fold any of this in.  Just let me know which direction you'd like.

pjfanning (commented):
lgtm - thanks

PR #1211, opened 2026-08-13
Fix getDimFromCell trailing space for XSSF in-cell picture anchors

### LINKED ISSUE #1201: Picture Resize fails due to error in getDimFromCell()
Trying to resize a picture in a XSSF Spreadsheet but this fails consistently because the target size is computed as negative. The computation of the `picture.resize(double, double)` is done by the help of `getDimFromCell(...)` method in ImageUtils.class where the bug is located.
I havent'checked whether this logic works for pictures spreading over several cells, but it surely fails when the picture should be within one single cell. Both the width and height computation return 0 in the best case (when no Dx/Dy is defined), negative values otherwise (which leads to subsequent IllegalArgumentException)

    for (int cellIdx = startCell; cellIdx<=endCell; cellIdx++) {
                final double dim = nextSize.apply(cellIdx).doubleValue() * EMU_PER_PIXEL;
                double leadSpace = 0;
                if (cellIdx == startCell) {
                    //space in the leftmost cell
                    leadSpace = (hssfUnits > 0)
                        ? dim * startD/(double)hssfUnits
                        : startD;
                }

                double trailSpace = 0;
                if (cellIdx == endCell) {
                    // space after the rightmost cell
                    trailSpace = (hssfUnits > 0)
                        ? dim * (hssfUnits-endD)/(double)hssfUnits
                        : dim - endD;
                }
                targetSize += dim - leadSpace - trailSpace; // --> = dim - startD - dim - endD = -startD - endD
            }

The computation of trail space should not be `trailSpace = dim - endD` but rather `trailSpace = endD`

arimu1:
## Summary
- Fix `ImageUtils.getDimFromCell` when `hssfUnits == 0` (XSSF/OOXML anchors): use `endD` for trailing space instead of `dim - endD`, matching the semantics of the HSSF branch above it.
- Add regression test for a single-cell XSSF anchor where zero offsets previously produced 0-sized dimensions and broke `picture.resize()`.

Fixes #1201

## Test plan
- [x] `./gradlew :poi-ooxml:test --tests org.apache.poi.xssf.usermodel.TestXSSFPicture.testGetDimensionFromAnchorSingleCell`
- [x] `./gradlew :poi:test --tests org.apache.poi.ss.util.TestImageUtils`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/poi-ooxml/src/test/java/org/apache/poi/xssf/usermodel/TestXSSFPicture.java
+++ b/poi-ooxml/src/test/java/org/apache/poi/xssf/usermodel/TestXSSFPicture.java
@@ -17,10 +17,13 @@
 
 package org.apache.poi.xssf.usermodel;
 
+import static org.apache.poi.util.Units.EMU_PER_PIXEL;
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.awt.Dimension;
 import java.io.IOException;
@@ -31,6 +34,7 @@ import org.apache.poi.ss.usermodel.*;
 import org.apache.poi.ss.usermodel.ClientAnchor.AnchorType;
 import org.apache.poi.ss.usermodel.Shape;
 import org.apache.poi.ss.util.ImageUtils;
+import org.apache.poi.hssf.HSSFITestDataProvider;
 import org.apache.poi.util.LocaleUtil;
 import org.apache.poi.xssf.XSSFITestDataProvider;
 import org.apache.poi.xssf.XSSFTestDataSamples;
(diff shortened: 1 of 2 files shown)


arimu1:
Thanks for the review — you're right that the global `trailSpace = endD` change was incorrect.

**What was wrong:** For XSSF, `endD` is the EMU offset from the left of the *end* cell, so trailing space must stay `dim - endD` (same fraction semantics as HSSF). Setting `trailSpace = endD` globally would:
- add a full extra cell for multi-cell anchors with `dx2/dy2 == 0`
- under-report normal in-cell anchors where `startD < endD` (`dim - startD - endD` instead of `endD - startD`)

**Actual fix (#1201):** The bug is only the degenerate *same-cell* case (`col1==col2`, `dx1==dy1==dx2==dy2==0`) where `endD - startD == 0`, causing `picture.resize()` to fail. The updated commit reverts the global flip and handles only that case: when `startCell == endCell` and computed span `<= 0`, use full cell size if both offsets are 0, otherwise remaining cell space (`dim - leadSpace`).

**Tests added:**
1. Degenerate in-cell `(0,0,0,0)` → full cell width/height (issue repro + resize)
2. Normal in-cell `startD < endD` → size == `endD - startD`
3. Multi-cell `dx2==0` → width == first column only (no extra cell)

Tip: `05307f004acc5427f5c1164a37b6e1c299413757`

poi-ooxml/src/test/java/org/apache/poi/xssf/usermodel/TestXSSFPicture.java
```diff
@@ -31,6 +34,7 @@ Licensed to the Apache Software Foundation (ASF) under one or more
 import org.apache.poi.ss.usermodel.ClientAnchor.AnchorType;
 import org.apache.poi.ss.usermodel.Shape;
 import org.apache.poi.ss.util.ImageUtils;
+import org.apache.poi.hssf.HSSFITestDataProvider;
```
pjfanning: use XSSFITestDataProvider

arimu1:
Thanks @pjfanning — switched `testGetDimensionFromAnchorDegenerateInCell` to `XSSFITestDataProvider.instance.getTestDataFileContent` and dropped the unused HSSF import. `TestXSSFPicture` (13 tests) and `TestImageUtils` pass locally (JDK 17). Tip: `2b9a8ab760c07e2b19edeedbe5bf4effe9a5868a`.

pjfanning:
thanks - merged

PR #1234, opened 2026-09-01
Fix `findColumnIndex()` to handle newline characters in column headers

bdw429s:
When a table column header contains a newline character (`\n`), `updateHeaders()` encodes it as `_x000a_` (the OOXML `_xXXXX_` escape convention) before storing in the column definition. However, `findColumnIndex()` did not perform the reverse encoding when looking up column names, causing lookups against multi-line column headers to always return `-1`.

This fix adds `\n` → `_x000a_` and `\r` → `_x000d_` replacements to the `findColumnIndex()` string normalization chain, matching the encoding already applied by `updateHeaders()`.

**Changes:**
- `XSSFTable.java` — `findColumnIndex()` now replaces `\n` with `_x000a_` and `\r` with `_x000d_` before looking up the column name in the map
- `TestXSSFTable.java` — Added `findColumnIndex` assertion to `testNamesWithNewLines()` and new `testFormulaWithNewlineColumnName()` test that verifies `SUM(Table1[Value\nA])` evaluates correctly with newline column names

### DIFF
--- a/poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFTable.java
+++ b/poi-ooxml/src/main/java/org/apache/poi/xssf/usermodel/XSSFTable.java
@@ -877,10 +877,15 @@ public class XSSFTable extends POIXMLDocumentPart implements Table {
             }
         }
         // Table column names with special characters need a single quote escape
-        // but the escape is not present in the column definition
+        // but the escape is not present in the column definition.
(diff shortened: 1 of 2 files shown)


bdw429s:
@pjfanning Let me know if you have any questions.

pjfanning (approved):
lgtm - thanks

PR #1265, opened 2026-09-13
Fix getRelationshipPartName for percent-encoded part names

### LINKED ISSUE #1264: PackagingURIHelper.getRelationshipPartName() returns null for part names with percent-encoded characters
Hi, ran into a bug in `PackagingURIHelper.getRelationshipPartName()` that silently returns `null` when the input part name contains percent-encoded characters like `%20` for spaces.

## What's happening

The method calls `partName.getURI().getPath()` which decodes percent-encoding — so `/aasx/test%20document.txt` becomes `/aasx/test document.txt`. That decoded string then gets passed to `createPartName()`, which internally calls `new URI(fullPath)`. Since literal spaces are illegal in URIs, it throws `URISyntaxException`, which gets caught and silently swallowed with a `return null`.

The comment even says "Should never happen in production" — but it absolutely does happen whenever filenames contain spaces or other characters requiring percent-encoding.

## Affected code

`org.apache.poi.openxml4j.opc.PackagingURIHelper.getRelationshipPartName()`

```java
String fullPath = partName.getURI().getPath();  // BUG: decodes %20 to space
// ... path manipulation ...
retPartName = createPartName(fullPath);  // FAILS: space is illegal in URI
```

## Steps to reproduce

1. Create an OPC package (`.aasx`, `.docx`, `.xlsx`, `.pptx`) with a part that has spaces in its filename, e.g. stored internally as `/aasx/test%20text%20document.txt`
2. Open it with POI and try to resolve relationships for that part
3. `getRelationshipPartName()` returns `null`

## Suggested fix

Replace `getPath()` with `getRawPath()` to preserve percent-encoding:

```java
// Before (buggy):
String fullPath = partName.getURI().getPath();

// After (fixed):
String fullPath = partName.getURI().getRawPath();
```

The comparison at the top of the method also uses `getPath()` on both sides — it works by coincidence since both sides are decoded consistently, but should also be updated to `getRawPath()` for correctness.

## Context

Hit this while working on [eclipse-aas4j](https://github.com/eclipse-aas4j/aas4j/issues/517) where AASX packages with embedded files containing spaces in their names caused relationship resolution to fail silently. The fix is minimal — just swapping `getPath()` for `getRawPath()`. Happy to submit a patch if this looks good to you.

kalayciburak:
## What this PR does

`PackagingURIHelper.getRelationshipPartName()` used `URI.getPath()`, which decodes percent-encoding. A part name such as `/aasx/test%20document.txt` then became `/aasx/test document.txt`, `createPartName` threw `InvalidFormatException`, and the method returned `null`.

Use `URI.getRawPath()` (and take the filename from that raw path) so the relationship part name stays a valid URI: `/aasx/_rels/test%20document.txt.rels`.

Fixes #1264

## Tests

Executed:

- `./gradlew :poi-ooxml:test --tests org.apache.poi.openxml4j.opc.TestPackagePartName` — 3/0
- `./gradlew :poi-ooxml:test --tests org.apache.poi.openxml4j.opc.TestPackagePartName --tests org.apache.poi.openxml4j.opc.TestRelationships --tests org.apache.poi.openxml4j.opc.compliance.TestOPCCompliancePartName --tests org.apache.poi.openxml4j.opc.TestPackage` — 66 completed, 1 skipped, 0 failed


pjfanning:
Thanks @kalayciburak - could you have a look at this?

Suggested fix in the same PR: make getSourcePartUriFromRelationshipPartUri operate on relationshipPartUri.getRawPath() (and derive the filename from the raw path the same way this PR does), and add a round-trip test getSourcePartUriFromRelationshipPartUri(getRelationshipPartName(p).getURI()) == p.getURI() for /aasx/test%20document.txt. Without that, the PR's claim of fixing #1264 only holds for the narrow read-side lookup.

No other correctness issues found in the diff.

Findings:
- poi-ooxml/src/main/java/org/apache/poi/openxml4j/opc/PackagingURIHelper.java:647 — inverse getSourcePartUriFromRelationshipPartUri (line 457) still decodes the path, so ZipPartMarshaller.marshall/OPCPackage.save and OPCPackage.removePart still throw (IllegalArgumentException: path) for percent-encoded part names; the #1264 scenario remains broken on the write side.

kalayciburak:
switched the inverse to getRawPath too, same as getRelationshipPartName. round-trip for /aasx/test%20document.txt is in the test.

pjfanning:
Thanks - merged

PR #1271, opened 2026-09-17
Standalone formula API: evaluate Excel formulas outside sheets

rmannibucau:
# Standalone formula API: evaluate Excel formulas outside sheets

## Motivation

Evaluating an Excel formula currently requires a file-backed workbook (XSSF/HSSF): you must build a `Workbook`, create a sheet and cells, then run a `FormulaEvaluator` against them. This is heavyweight for the common case of "just evaluate this formula with these inputs" - e.g. expression validation, calculators, batch re-computation, or server-side engines - and it is measurably slower (file/record parsing, sheet and cell allocation) than what the formula engine actually needs.

This PR adds an opt-in **standalone formula engine** that evaluates real Excel formulas (same syntax, same evaluator machinery, no re-implementation) without any file or sheet:

```java
try (StandaloneFormulaEngine engine = StandaloneFormulaEngine.create()) {
    CompiledFormula compiled = engine.compile("SUM(A1:A10)*2");
    try (StandaloneFormulaEvaluator evaluator = compiled.newEvaluator()) {
        evaluator.setCellValue("A1", 1);
        evaluator.setCellValue("A2", 2);
        // ...
        LightCellValue result = evaluator.evaluate();
    }
}
```

## What is added

All new public types are in `org.apache.poi.ss.usermodel` and annotated `@since 6.0.0`:

- `StandaloneFormulaEngine` - entry point: parses/compiles formulas and validates defined names (mirrors `XSSFName` validation: max length, invalid characters, R/C shorthand, `TRUE`/`FALSE` literals, and names that collide with cell references such as `A1` are rejected; legal Excel defined names like bare column letters `X` are accepted, matching Excel semantics).
- `CompiledFormula` - a parsed, reusable formula; compiling once and evaluating many times with different inputs avoids re-parsing.
- `StandaloneFormulaEvaluator` - evaluates a compiled formula against a virtual sheet; inputs are set with `setCellValue(name, value)` and results read as `LightCellValue`.
- `LightCellValue` - a sealed, memory-leaner value abstraction; `CellValue` now implements it. Existing `CellValue` users are unaffected.

Implementation lives in 6 package-private classes under `org.apache.poi.ss.formula` (`CompiledFormulaImpl`, `StandaloneFormulaEngineImpl`, `StandaloneFormulaEvaluatorImpl`, `VirtualEvaluationCell`, `VirtualEvaluationSheet`, `VirtualFormulaWorkbook`), reusing the existing `ss.formula` evaluation engine - no fork of the formula model.

## Impact on existing users

None, by design: the feature is strictly additive and opt-in. No existing class changes behavior (the only existing-file touch is a javadoc reference on `CellValue` to `LightCellValue`, plus `implements LightCellValue`).

Note: the `OperationEvaluationContext` reference-evaluator caching that was part of the first iteration of this patch was extracted and merged separately as #1272, so it is no longer part of this diff.

## Performance

Because no file, workbook, sheet or cell records are materialized, repeated evaluations run **2x-10x faster** than the equivalent file-backed `FormulaEvaluator` flow (formula parsed once via `compile`, then re-evaluated with new inputs). Parity with the file-backed evaluator is covered by `TestStandaloneFormulaEvaluatorParity`.


pjfanning:
I see the link to the Apache Hop discussion but can you provide a more detailed description (in PR description) which can document this for release notes and all new public APIs and classes should have `@since 6.0.0`.

I haven't yet reviewed this in detail but if is opt-in and that existing users are unaffected by this changes (unless they opt-in by changing their code), then it probably ok to have this merged.

poi/src/main/java/org/apache/poi/ss/usermodel/CellValue.java
```diff
@@ -23,8 +23,12 @@ Licensed to the Apache Software Foundation (ASF) under one or more
  * Mimics the 'data view' of a cell. This allows formula evaluator
  * to return a CellValue instead of precasting the value to String
  * or Number or boolean type.
+ *
+ * Consider {@link LightCellValue} for new code: it is a sealed, memory-leaner
+ * counterpart of this class that CellValue itself implements.
  */
-public final class CellValue {
+// @Deprecated
```
pjfanning: remove this comment - this is not deprecated

rmannibucau: this is 100% superseeded and intended to be deprecated, the only reason it is not is that it would break downstream project in next minor but intent is `@Deprecated(forRemoval="7.0.0")`

pjfanning:
Input names that look like cell references silently resolve to the wrong thing. requireValidName only rejects null/blank. FormulaParser tries cell-reference parsing before name lookup, so inputs("Q1", "FY2024", "H2") — realistic column headers — would evaluate Q1*2 against virtual cell Q1 (blank, or a different input if you have ≥17 inputs) with no error. The engine should reject names that CellReference.classifyCellReference(name, version) classifies as CELL/COLUMN/ROW, and probably names containing characters Excel forbids in defined names (spaces, operators), which otherwise fail only at compile() with an unhelpful parse error.

poi/src/main/java/org/apache/poi/ss/usermodel/CellValue.java
pjfanning: I disagree and actually this change is not even needed in this PR. I'm in no hurry to merge this. This whole evaluator looks like it can live in Apache Hop anyway.

pjfanning (changes requested):
I'm not convinced that this needs to be in POI at all.

rmannibucau:
@pjfanning how do you propose to solve that a workbook is super slow using HSSF and insanely slow using XSSF cause it is backed by the direct file format and not in memory computation/model?
Also, and more on an user perspective, I think it belongs to Apache POI to expose the Excel formula model as an API outside workbooks, not sure another project would make sense there. It is a great feature to propose for projects enabling not technical users to map/combine data - already hit that need multiple times and everybody knows excel format.

pjfanning:
POI is for Microsoft formats. Just because you like some of its abstractions for your own work that doesn't mean that the POI PMC which very few active members needs to take on providing extensible Formula Evaluators.
I'm happy to open up the POI APIs to make it easier to extend in 3rd party code.

poi/src/main/java/org/apache/poi/ss/usermodel/CellValue.java
rmannibucau: will not, current Hop proposal is to keep apache poi and flag it as such and too slow for more than a few line and create a new formula component for the rest totally unrelated....also means we loose excel syntax which is not great.

can you explain why you do disagree since it is not redundant and consumes x2 memory?

rmannibucau:
@pjfanning I understand that more than you think but also see that POI is not just MS file format, it is also Excel formula support by design (ack it can be seen as a side effect but it is there). This feature alone is worth exposing without the overhead of the file format and this is what this PR does, most of the code is just having file-less impl of the format api to be able to get the formula and access `WorkbookEvaluator`. Rest is perf optim all impl will benefit at some point.

Doesn't need to urge on my side too, only trying to make everyone happy (both projects + end users) without having to fork anything.

poi/src/main/java/org/apache/poi/ss/usermodel/CellValue.java
pjfanning: POI is a 20 year old project and we turn people off upgrading by removing APIs.

rmannibucau: well if for goods it is always doable and major (6.0.0 is coming) are the only moment you can do it so think it is worth there, note that I dont plan to remove in 6, just hope to communicate to not use anymore

rmannibucau:
Ack, this is now handled by aa3c6a12 + 474aac4b: input names go through `requireValidName` (rules based on `XSSFName.validateName`: max 255 chars, first char letter/underscore/backslash, no digits or invalid characters; plus rejection of the `R`/`C` R1C1 shorthand and the `TRUE`/`FALSE` literals which the formula parser would otherwise swallow as booleans) and then `requireReferenceFreeName`, which rejects any name that `CellReference.classifyCellReference(name, version)` classifies as a cell reference for the active `SpreadsheetVersion`. So `setCellValue("Q1", ...)` now fails fast with an `IllegalArgumentException` instead of silently resolving to the virtual cell.

Bare letters like `x` stay accepted on purpose: `NameType.COLUMN` names are legal Excel defined names - only `R` and `C` are reserved as R1C1 shorthand, and those are already rejected by `requireValidName` - and rejecting them would exclude names Excel itself accepts (consistent with the `XSSFName.validateName` policy). This rationale is documented in the javadoc of `requireReferenceFreeName`. If you prefer a stricter deny-list there, happy to discuss it.

pjfanning:
My opinion has not changed on this, that this code does not need to be in POI. You actually gain by it not being in POI. Moving it to POI means:
* a PMC that doesn't worry about this feature
* needing approval for every change
* waiting on POI releases which are infrequent

You can keep the poi jar dependency so that you don't need to duplicate everything. You can replace the POI code that you don't like.
When POI 6.0.0 is released and it is not yet planned, you can adjust your standalone code to pick up on some recent changes that I added which means you would need to duplicate a little less code.
* #1273
* #1274 
* #1275
* #1305

rmannibucau:
> a PMC that doesn't worry about this feature

It is the same for xss/hss/etc so not a bit deal

> needing approval for every change

yep, this is apache and all good for me

> waiting on POI releases which are infrequent

this is very solvable and the way apache move forward (this is why we do reports to the board)

The gains are obvious on the other side:

* Formula component becomes a shared backbone provided by a central project and we don't compete between projects at apache
* It is easy for end users to consume it and use it accross the board
* Communication is clear (apache to external)
* Project potentially gets more activity which is sane if I understand your feedback right

So for me pros >> cons.

pjfanning:
There are zero consumers for this. As far as I can see Apache Hop don't want it.
I'm unsubscribing from this issue and will vote against merging it.

rmannibucau:
> There are zero consumers for this. As far as I can see Apache Hop don't want it.

Both statements are 100% wrong, not sure which part you misread - and I can guarantee you there are even several private consumers for this as well.
Please do not conclude things when the opposite is stated even if everything can't be made public there.

Also apache hop even said the opposite, the disagreement is more on the merge in a single transform or the creation of another one (this part was validated and useless if apache poi host this code).

rmannibucau:
Here is the results I get with this PR:

| Formula | trunk classic XSSF | branch classic HSSF | standalone | standalone/trunk |
|---|---|---|---|---|
| `A1+B1*C1-D1/E1` | 12.660 ±2.363 | 6.122 ±1.412 | 0.374 ±0.072 | x33.9 |
| `IF(A1>800,"High",IF(A1>500,"Medium",IF(A1>200,"Low","Very Low")))` | 17.495 ±2.871 | 8.139 ±1.726 | 1.080 ±0.192 | x16.2 |
| `A1&"_"&B1` | 8.612 ±1.628 | 3.593 ±1.167 | 0.951 ±0.159 | x9.1 |
| `IF(A1>500,"High-"&B1,"Low-"&B1)` | 12.708 ±2.538 | 5.407 ±1.064 | 0.959 ±0.165 | x13.3 |
| `SQRT(POWER(A1,2)+POWER(B1,2))` | 12.071 ±2.205 | 5.249 ±1.090 | 0.328 ±0.056 | x36.8 |
| `SUM(A1:C1)*2` | 7.354 ±1.997 | 3.731 ±1.046 | 0.219 ±0.031 | x33.6 |

pjfanning:
Apache POI™ - the Java API for Microsoft Documents.

The PR is not useful for Microsoft Documents.

The new evaluator is not for formula evaluation in xls and xlsx files.

The benchmarks are not comparing POI use cases for Microsoft Documents. Any code added to the POI project that is not for parsing or writing Microsoft Documents is scope creep.

rmannibucau:
yes and no, before entering into more details I'd like another PMC optinion since formula are generally a big criteria to import POI in a project (the other one being to read xls(x) files but really not the only one.

now the benchmark is ultra relevant for documents: there is no valid reason it is not as fast for HSS and XSS for most cases (edge ones where it can be valid is bloated sheets - not common, cross workbook docs - acceptable), so at some point I think this PR could be the base for the HSS and XSS impl which would be a 2 ways sync instead of using storage as in memory computation model (and please don't change it on trunk right now as the changes you did yesterday which overlapped with this PR after you reviewed it, i'd prefer to converge even with multiple iteration/commits/PR).

PR #1273, opened 2026-09-17
Make WorkbookEvaluator.evaluateFormula(OperationEvaluationContext, Ptg[]) public

pjfanning:
`WorkbookEvaluator.evaluateFormula(OperationEvaluationContext, Ptg[])` is the low-level interpreter entry point: it evaluates parsed formula tokens without a formula cell, result cache or dependency tracking. It was package-private ("visibility raised for testing").

The rest of the evaluation SPI is already public — `EvaluationWorkbook`, `EvaluationSheet`, `EvaluationCell`, `FormulaParsingWorkbook`, `FormulaParser.parse(...)`, the `WorkbookEvaluator` and `OperationEvaluationContext` constructors — so this was the one missing hook for evaluating formulas against a custom in-memory `EvaluationWorkbook` without building a `Workbook`. That lets a "standalone formula engine" such as the one proposed in #1271 be built outside POI.

Kept `@Internal` to signal that it is a low-level hook, and added javadoc.

Extracted from #1271 (see discussion there).


pjfanning:
I reverted this because it seems quite incomplete but I then added it back because it might be enough.

PR #1276, opened 2026-09-18
Prevent negative remaining size in EscherBSERecord

billowgao:
### Problem
`EscherBSERecord.fillFields()` subtracts the fixed 36-byte BSE
record header from `bytesRemaining` without checking whether the
declared record size is smaller than 36 bytes.
 
For malformed files with a truncated BSE record, this can make
`bytesRemaining` negative and later cause a negative-size operation.

### Change
Clamp `bytesRemaining` to zero when subtracting the fixed record
size and embedded blip size.
 
This does not catch or suppress parser exceptions; it only prevents
the remaining-byte count from becoming negative.
 
### Test
Added a regression XLS containing a BSE record whose declared
payload is 35 bytes.
 
The test reproduces the failure before the change and passes after
the change.


poi/src/main/java/org/apache/poi/ddf/EscherBSERecord.java
```diff
@@ -107,7 +107,7 @@ public int fillFields(byte[] data, int offset, EscherRecordFactory recordFactory
         field_9_name = data[pos + 33];
         field_10_unused2 = data[pos + 34];
         field_11_unused3 = data[pos + 35];
-        bytesRemaining -= 36;
+        bytesRemaining = Math.max(0, bytesRemaining - 36);
```
pjfanning: A BSE record with fewer than 36 bytes of payload is malformed (the fixed fields have already been read past its declared end at this point), so fail with a clear message rather than clamping and reporting more bytes consumed than the record declared:

```suggestion
        if (bytesRemaining < 36) {
            throw new RecordFormatException("EscherBSERecord declares " + bytesRemaining
                    + " bytes but needs at least 36 for its fixed fields");
        }
        bytesRemaining -= 36;
```

poi/src/main/java/org/apache/poi/ddf/EscherBSERecord.java
```diff
@@ -121,7 +121,7 @@ public int fillFields(byte[] data, int offset, EscherRecordFactory recordFactory
             bytesRead = field_12_blipRecord.fillFields( data, pos + 36, recordFactory );
         }
         pos += 36 + bytesRead;
-        bytesRemaining -= bytesRead;
+        bytesRemaining = Math.max(0, bytesRemaining - bytesRead);
```
pjfanning: Same here: if the embedded blip consumed more than the BSE declared, the file is inconsistent. Before this PR that surfaced as the misleading `safelyClone` message; clamping hides it and the returned size again overshoots the declared record. Better to say what went wrong:

```suggestion
        if (bytesRead > bytesRemaining) {
            throw new RecordFormatException("Embedded blip of EscherBSERecord consumed " + bytesRead
                    + " bytes but only " + bytesRemaining + " were declared");
        }
        bytesRemaining -= bytesRead;
```

poi/src/test/java/org/apache/poi/ddf/TestEscherBSERecordMalformedInput.java
```diff
@@ -0,0 +1,42 @@
...
+
+        LittleEndian.putShort(data, 2, EscherBSERecord.RECORD_ID);
+        // A BSE record has 36 bytes of fixed fields after its 8-byte header.
+        // This malformed record declares only 35 bytes of payload.
+        LittleEndian.putInt(data, 4, 35);
+
+        EscherBSERecord record = new EscherBSERecord();
+        int bytesRead = record.fillFields(data, 0, new DefaultEscherRecordFactory());
+
+        assertEquals(44, bytesRead);
```
pjfanning: With the explicit check the test would assert the clear failure instead of the over-reported size (note `44` here is one more than the 8 + 35 bytes the record declares):

```suggestion
        EscherBSERecord record = new EscherBSERecord();
        DefaultEscherRecordFactory factory = new DefaultEscherRecordFactory();
        RecordFormatException e = assertThrows(RecordFormatException.class,
                () -> record.fillFields(data, 0, factory));
        assertTrue(e.getMessage().contains("needs at least 36"), e.getMessage());
```

(imports: `assertThrows`, `assertTrue`, `org.apache.poi.util.RecordFormatException`.)

pjfanning (commented):
Thanks for the PR. The old behaviour here was already an exception rather than a crash: a negative `bytesRemaining` reaches `IOUtils.safelyClone`, which throws `RecordFormatException` — but with a misleading message ("You can set a higher override value with EscherBSERecord.setMaxRecordLength()"). Clamping to 0 turns that into silent acceptance, and the value returned from `fillFields()` then overshoots the declared record length (44 for a record that declared 43 bytes), so `EscherContainerRecord.fillFields` starts the next sibling one byte late.

I'd suggest failing explicitly with a clear message instead (which is what `EscherDggRecord` does for malformed remaining data) — suggestions inline. Both need `import org.apache.poi.util.RecordFormatException;`.

A couple of smaller things:
- Ideally the `< 36` check sits before the fixed-field reads (`data[pos] .. data[pos + 35]`) so a truncated buffer fails with the clear message rather than an `ArrayIndexOutOfBoundsException`; the suggestion below can only touch the lines in the diff.
- `TestEscherBSERecord` already exists in this package — the new test would fit better there than in a new class.

pjfanning:
Merged - thanks

PR #1279, opened 2026-09-18
EVEN and ODD: compute in doubles instead of overflowing a long cast

pjfanning:
Follow-up to #1275.

`EVEN` and `ODD` cast the magnitude of their argument to `long` and masked the low bit, which saturates at 2^63: `EVEN(1E19)` gave `-9.22E18` and `EVEN(-1E19)` gave `+9.22E18` (the sign came out wrong as well as the magnitude). Both are now one small helper that works in doubles for any finite value: round the magnitude up to an integer, add one if it has the wrong parity, restore the sign.

Like `INT`, `FLOOR` and `CEILING` since #1275, the ceiling acts on the 15 significant digits Excel exposes (`ExcelArithmetic.approxValue`), so `EVEN(2.0000000000000004)` is `2` rather than `4` and `ODD(1.0000000000000002)` is `1` rather than `3`. Values that differ from an integer within 15 digits (`EVEN(2.00000000000001)` → `4`) round up as before.

There were no `EVEN`/`ODD` tests at all; `TestNumericFunction.testEVEN`/`testODD` now cover the documented examples from the Microsoft function pages, signs, coercion, the 2^63 overflow, 2^53 boundaries and the 15-digit view. Ran `TestNumericFunction`, `TestExternalFunctionFormulas` and `TestFormulasFromSpreadsheet` locally: 396 tests, 0 failures.


pjfanning:
will merge as part of #1280

PR #1328, opened 2026-09-20
[bug-64369] regression test: COUNTIF(range,"<>*") counts the non-text cells

pjfanning:
Regression test for https://bz.apache.org/bugzilla/show_bug.cgi?id=64369 — already fixed by #1327 (bug 69853).

`COUNTIF(range,"<>*")` is the "count non-text cells" idiom: `*` matches any text, so `<>*` is every cell that is not text. It always gave 0 because a non-text cell never matched a string criterion, whatever the operator — the same root cause as 69853, where `"<>"` ignored numbers.

The test is the reporter's attached matrix (numbers, texts including ones that look like criteria, a date) checked cell by cell for `"*"` and `"<>*"`, plus the whole-range totals for the five criteria from their test class and a blank cell (not text either). It fails on the commit before #1327 (`expected: <1.0> but was: <0.0>`) and passes on trunk.


utafrali (approved):
Clean, well-structured regression test that mirrors the reporter's exact scenario. It correctly covers individual cell-type classification, aggregate counts, text-that-looks-like-criteria edge cases, and blank-cell behavior, all following established TestXSSFBugs conventions.

